PR #4, opened 2023-11-28
Add Trivy scan (NDISC-61)

hannes-mk:
I did not use the template which scanned an image.
Instead I followed the docs: on [Git repository scanning](https://github.com/aquasecurity/trivy-action#using-trivy-to-scan-your-git-repo)

Here's how a scan looks like: https://github.com/digitalservicebund/ris-norms/actions/runs/7019008273/job/19095553247

It *does* add data to the "GitHub Security Tab"

### DIFF
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -144,17 +144,31 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
-
       - name: Run Trivy vulnerability scanner in repo mode
         uses: aquasecurity/trivy-action@2b6a709cf9c4025c5438138008beaddbb02086f0
         with:
           scan-type: 'fs'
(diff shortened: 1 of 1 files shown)


VictorDelCampo:
- Use SHA of trivy-action, see https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions
- severity only "CRITICAL"? we used to have in ris-backend-service "CRITICAL/HIGH"
- what about failing the pipeline if agreed severity is found? together with caselaw we decided to fail the pipeline so that we react as soon as possible. Otherwise the dependabot shifter (also responsible for open security issues) may not have time / forget it
- please be aware of this thread where Kai reported that there was an update on the trivy-action: https://digitalservicebund.slack.com/archives/C01JUUDHG5Q/p1699463163583089
- what is this  ignore-unfixed: true

PR #97, opened 2024-02-16
Add sample data via migration

hannes-mk:


hannes-mk:
We should rewrite the tests so that they become independent from the state of the database.

hannes-mk:
We decided to not use flyway for adding sample data.

PR #113, opened 2024-02-22
Update backend libraries

hannes-mk:
- Fix format and sort using the gradle task "versionCatalogUpdate"

RISDEV-3431


hannes-mk:
@VictorDelCampo This is what I get when running the gradle task "versionCatalogUpdate". I'd be interested if you think the generated TOML is fine.

hannes-mk:
@VictorDelCampo And maybe we can handle those together? 
image

VictorDelCampo:
@hannes-mk the generated TOML looks good! I see it is ordered alphabetically right? Why not... But i see the flyway-core is missing and the sl4fj? Please be aware the flyway-core is not going to be easy

hannes-mk:
Done in https://github.com/digitalservicebund/ris-norms/commit/ebc0aac04800e8123a7346ec6a21c17ab73e71d0

PR #119, opened 2024-02-23
Use C4 diagrams for documentation

hannes-mk:
Document the existing state with C4 diagrams, cf. https://c4model.com/
Support of C4 in mermaid is marked "experimental", but seems to work so far, cf.  http://mermaid.js.org/syntax/c4.html


hannes-mk (commented):
The current version is not the context diagram, but too detailed.

There should be the BmJ with the Documentalists and DS with the Web App. Nothing more.

doc/architecture/c4-system-context.md
```diff
@@ -0,0 +1,35 @@
...
+
+    Person("documentalist", "Documentalist", "A staff member of the<br> 'Normendokumentation (BfJ)'")
+
+    System("ris-norms", "RIS Norms", "Allows users to<br> consolidate and annotate<br> amending laws<br> (later also: new norms)")
+    Rel("documentalist", "ris-norms", "Adjusts new amending laws,<br> adjusts metadata <br> and publishes the new norms<br> (or versions of norms)")
+
+    System("portal", "Portal", "Gives the public access to norms")
+    Rel("ris-norms", "portal", "Makes new norms accessible to the portal")
+
+    Person("public", "General public, lawyer, law maker,<br> public servant, private enterprise...", "Searches, reads or processes norms for<br> private or professional reasons")
```
VictorDelCampo: Why this "Person"? it is not relevant for the architecture of norms. As soon as we send data (LDML.de) to portal our system context is done. What users/stakeholders are interacting with the portal is not relevant for the norms architecture

hannes-mk: I agree. That's the scope of the portal team and we should not make assumptions about it in our docs.
Will update.

PR #139, opened 2024-02-28
Add time machine core functionality to the API (RISDEV-3467 as part of RISDEV-3446)

hannes-mk:
Business logics only. For the actual interfacing (API), there exist separate tickets.


VictorDelCampo:
@hannes-mk not sure if the the domain package is the right place for the time machine. For me the domain package contains only the domain model. For the time machine there is no new entity in the domain model. I would go for:
- Creating 1 input port something like `applyTimeMachine` that would maybe take two elis, the one of the amending law and the one of the target law and output the new version of the target law with the changes applied.
- Creating 1 output port that would communicate with the db for getting the XML of both elis? Not sure if you were thinking of sending the XMLs from the frontend, could also be an option, but in my opinion not needed. Although if somebody wants to preview changes made on the amending law without saving this one then we would need to send at least the amending law xml from the fontend.

Then we would have the service in the application layer implementing the input port and the db service in the adapter layer implementing the output port

hannes-mk:
Thanks , @VictorDelCampo !
I wondered where to put the code and will bring this up for discussion.
My hunch is that it's pure business logics, hence part of the "core". And at the same time, I agree that it feels wrong to make it part of the "domain".

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/TimeMachineFunctions.java
```diff
@@ -0,0 +1,118 @@
+package de.bund.digitalservice.ris.norms.application.service;
+
+import java.util.Optional;
+import org.w3c.dom.Document;
+import org.w3c.dom.Node;
+
+/**
+ * Namespace for business Logics related to "time machine" functionality, i.e. to applying LDML.de
+ * "modifications" to LDML.de files. For details on LDML.de modifications, cf.
+ * https://gitlab.opencode.de/bmi/e-gesetzgebung/ldml_de/-/tree/main/Spezifikation?ref_type=heads
```
VictorDelCampo: Link specified as plain text. Maybe replace URL with HTML link?

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/TimeMachineFunctions.java
```diff
@@ -0,0 +1,118 @@
...
+ */
+public class TimeMachineFunctions {
+
+  // prevent default constructor
+  private TimeMachineFunctions() {}
```
VictorDelCampo: I suppose you want to prevent a default constructor because all functions underneath are static? Why not just using a @Service or @Component annotation from spring?

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/TimeMachineFunctions.java
```diff
@@ -0,0 +1,118 @@
...
+
+  /**
+   * Applies the modifications of the amending law onto the target law.
+   *
+   * @param amendingLaw An Document that contains LDML.de modifications to be applied on the target
```
VictorDelCampo: A Document instead of An Document

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/TimeMachineFunctions.java
```diff
@@ -0,0 +1,118 @@
...
+   *     law
+   * @param targetLaw The Document that the modifications will be applied to
```
VictorDelCampo: I guess you use Document with a capital letter because you are referring to the java class? What about using @link?

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/XmlFunctions.java
```diff
@@ -0,0 +1,91 @@
...
+      // XXE vulnerability hardening, cf. https://www.sonarsource.com/blog/secure-xml-processor/
+      factory.setFeature("http://apache.org/xml/features/disallow-doctype-decl", true);
+      factory.setFeature(XMLConstants.FEATURE_SECURE_PROCESSING, true);
+      factory.setExpandEntityReferences(false);
+
+      final DocumentBuilder builder = factory.newDocumentBuilder();
+      final InputSource is = new InputSource(new StringReader(xmlText));
+      final Document document = builder.parse(is);
+      return Optional.of(document);
+    } catch (Exception e) {
```
VictorDelCampo: Why don't you just throw the exception in the method? This functionality will be called through a controller, for now, and we have a spring @ControllerAdvice catching all exceptions being thrown in all use cases. Also  catching the general "Exception" is not recommended as best practice, since it may also catch unexpected runtime exceptions including those you didn't anticipate.
Also catching an exception and not even logging anything would make it more difficult to investigate in case of errors

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/XmlFunctions.java
```diff
@@ -0,0 +1,91 @@
...
+   * @return the Node identified by the <code>xPathExpression</code>
+   */
+  public static Optional<Node> getNodeByXPath(String xPathExpression, Node sourceNode) {
+    try {
+      XPathFactory xpathfactory = XPathFactory.newInstance();
+      XPath xpath = xpathfactory.newXPath();
+      Node nodeByXPath = (Node) xpath.evaluate(xPathExpression, sourceNode, XPathConstants.NODE);
+      return Optional.of(nodeByXPath);
+    } catch (Exception e) {
+      // we don't analyze, right now
```
VictorDelCampo: Same topic with catching Exception

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/XmlFunctions.java
```diff
@@ -0,0 +1,91 @@
...
+      Node originalRootNode = originalDocument.getDocumentElement();
+      DocumentBuilderFactory dbf = DocumentBuilderFactory.newInstance();
+      DocumentBuilder db = dbf.newDocumentBuilder();
+      Document clonedDocument = db.newDocument();
+      Node clonedRootNode = clonedDocument.importNode(originalRootNode, true);
+      clonedDocument.appendChild(clonedRootNode);
+      return Optional.of(clonedDocument);
+      // I don't know how to test that DocumentBuilderFactory.newInstance() can throw
+    } catch (Exception e) {
+      // we don't analyze, right now
```
VictorDelCampo: Same topic with catching Exception

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/TimeMachineFunctions.java
```diff
@@ -0,0 +1,118 @@
...
+import java.util.Optional;
+import org.w3c.dom.Document;
+import org.w3c.dom.Node;
+
+/**
+ * Namespace for business Logics related to "time machine" functionality, i.e. to applying LDML.de
+ * "modifications" to LDML.de files. For details on LDML.de modifications, cf.
+ * https://gitlab.opencode.de/bmi/e-gesetzgebung/ldml_de/-/tree/main/Spezifikation?ref_type=heads
+ */
+public class TimeMachineFunctions {
```
VictorDelCampo: I see this class as a service or component and would call it therefore "TimeMachine"

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/TimeMachineFunctions.java
```diff
@@ -0,0 +1,118 @@
...
+   * @param amendingLaw An Document that contains LDML.de modifications to be applied on the target
+   *     law
+   * @param targetLaw The Document that the modifications will be applied to
+   * @return the Document that results in applying the amending law's modifications to the target
+   *     law
+   */
+  public static Optional<Document> applyTimeMachine(
+      final Document amendingLaw, final Document targetLaw) {
+
+    final Optional<Document> targetLawClone = XmlFunctions.cloneDocument(targetLaw);
```
VictorDelCampo: Not sure about the usage of the Optional all the time, specially thinking about the throwing of errors if they happen at the method level. This way you could program the expected use case without always checking for isEmpty()

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/TimeMachineFunctions.java
```diff
@@ -0,0 +1,118 @@
...
+    Optional<Node> optionalNodeHrefAttribute =
+        XmlFunctions.getNodeByXPath("//*[local-name()='ref']/@href", modificationNode);
+
+    if (optionalNodeHrefAttribute.isPresent())
+      return Optional.of(optionalNodeHrefAttribute.get().getNodeValue());
+
+    return Optional.empty();
+  }
+
+  static Optional<String> getEIdfromModificationHref(String modificationHref) {
```
VictorDelCampo: Why package-private access level for some methods? Why not just private? I would also remove all static modifiers and inject the TimeMachine service as a whole. Like we do with the other services (AmendingLawService and TargetLawService)

VictorDelCampo (changes requested):
@hannes-mk I have just checked quickly your PR (not the tests sorry) since I am not sure I will have the time later. Could you please raise the need for technical review in the tech-daily?

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/TimeMachineFunctions.java
hannes-mk: nice

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/TimeMachine.java
hannes-mk, line 1: Should the file name be `TimeMachineService.java`?

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/TimeMachine.java
```diff
@@ -0,0 +1,119 @@
...
+   *
+   * @param modificationNode the node containing the substitution
+   * @param targetLaw the document version of the target law
+   * @return the <code>targetLaw</code> with the applied modification
+   */
+  private Document applyOneSubstitutionModification(Node modificationNode, Document targetLaw) {
+    final Document targetLawClone = xmlDocumentExtractor.cloneDocument(targetLaw);
+
+    XmlDocumentExtractor.ReplacementPair replacementText =
+        xmlDocumentExtractor.getReplacementText(modificationNode);
```
hannes-mk: Maybe have the function name similar to the type we're getting.
`getReplacementPair` or `getReplacementTextPair` (if the latter, then maybe also the type `XmlDocumentExtractor.ReplacementTextPair`

backend/src/main/java/de/bund/digitalservice/ris/norms/application/port/input/TimeMachineUseCase.java
```diff
@@ -0,0 +1,31 @@
...
+  /**
+   * A record representing the query for loading the xml representation of a target law and the
+   * string representation of the amending law xml. The query includes the ELI (European Legislation
+   * Identifier) to identify the target law.
+   *
+   * @param eliTargetLaw The ELI (European Legislation Identifier) used to identify the amending law
+   *     in the query.
+   * @param amendingLawXml The String representation of the amending law xml
+   */
+  record Query(String eliTargetLaw, String amendingLawXml) {}
```
hannes-mk: maybe `targetLawEli` ? (not sure we follow a pattern in the backend)

backend/src/main/java/de/bund/digitalservice/ris/norms/application/port/input/TimeMachineUseCase.java
```diff
@@ -0,0 +1,31 @@
...
+   * @return An {@link String} containing the updated {@link TargetLaw}.
+   */
+  String applyTimeMachine(Query query);
+
+  /**
+   * A record representing the query for loading the xml representation of a target law and the
+   * string representation of the amending law xml. The query includes the ELI (European Legislation
+   * Identifier) to identify the target law.
+   *
+   * @param eliTargetLaw The ELI (European Legislation Identifier) used to identify the amending law
```
hannes-mk: "... identify the target law ..."

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/XmlDocumentExtractor.java
```diff
@@ -0,0 +1,167 @@
...
+import org.w3c.dom.Document;
+import org.w3c.dom.Node;
+
+/**
+ * Service class within the application core part of the backend. It is responsible for working with
+ * w3c documents. It is annotated with {@link Service} to indicate that it's a service component in
+ * the Spring context.
+ */
+@Service
+public class XmlDocumentExtractor {
```
hannes-mk: `...Extractor` looks very specific to me. It does much more than extracting.

As for a better name, I struggle to find one that's better than `...Helper` (as we can't just have a `...Functions` namespace).

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/XmlDocumentExtractor.java
```diff
@@ -0,0 +1,167 @@
...
+    }
+  }
+
+  /**
+   * Get the first modification in the amending law that can be found
+   *
+   * @param amendingLaw the law to be searched
+   * @return the Node containing the modification
+   */
+  Node getFirstModification(Document amendingLaw) {
```
hannes-mk: I wonder about the place of these functions. They are too specific for a general `XmlHelper` as they rely on the semantics of the TimeMachine.

backend/src/test/java/de/bund/digitalservice/ris/norms/application/service/TargetLawServiceTest.java
```diff
@@ -62,4 +65,28 @@ void canLoadTargetLawXmlByEli() {
     verify(loadTargetLawXmlAdapter, times(1))
         .loadTargetLawXmlByEli(argThat(command -> command.eli().equals(eli)));
   }
+
+  @Test
+  void appliesTimeMachineWhenCalled() {
```
hannes-mk: I'm a little confused as to what's tested, here.

Is it that `timeMachine.apply()` has been called? If yes, then maybe rather `callsTimeMachineApply()`.

backend/src/test/java/de/bund/digitalservice/ris/norms/application/service/TargetLawServiceTest.java
```diff
@@ -62,4 +65,28 @@ void canLoadTargetLawXmlByEli() {
...
+    when(loadTargetLawXmlAdapter.loadTargetLawXmlByEli(any()))
+        .thenReturn(Optional.of(targetLawXmlAsString));
+    when(timeMachine.apply(any(), any())).thenReturn("Success");
+
+    // When
+    final String appliedXml = service.applyTimeMachine(query);
+
+    // Then
+    assertThat(appliedXml).contains("Success");
+    verify(loadTargetLawXmlAdapter, times(1))
```
hannes-mk, line 88: I notice three asserts/verifies. Maybe the name of the test can be more specific (as to what this test makes sure of).

backend/src/test/java/de/bund/digitalservice/ris/norms/application/service/TimeMachineTest.java
```diff
@@ -0,0 +1,227 @@
...
+import de.bund.digitalservice.ris.norms.application.service.exceptions.XmlProcessingException;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+import org.w3c.dom.Node;
+
+class TimeMachineTest {
+
+  @Test
+  void allRelevantMethodsAreCalled() {
```
hannes-mk: If it's not too hard, we should list the "relevantMethods". Otherwise the description is equivalent to "works correctly" and nobody's the wiser.

backend/src/test/java/de/bund/digitalservice/ris/norms/application/service/TimeMachineTest.java
```diff
@@ -0,0 +1,227 @@
...
+
+    // when
+    Throwable thrown = catchThrowable(() -> timeMachine.apply(amendingLawString, targetLawString));
+
+    // then
+    assertThat(thrown).isInstanceOf(XmlProcessingException.class);
+  }
+
+  @ParameterizedTest
+  @ValueSource(
```
hannes-mk: Btw. this `@ParameterizedTest` was recommended by Sonar. I see the point, but are not too happy about how it looks.

backend/src/test/java/de/bund/digitalservice/ris/norms/conventions/ArchitectureFitnessTest.java
```diff
@@ -36,7 +36,7 @@ class ArchitectureFitnessTest {
 
   static final String INPUT_PORT_LAYER_PACKAGES = BASE_PACKAGE + ".application.port.input";
   static final String OUTPUT_PORT_LAYER_PACKAGES = BASE_PACKAGE + ".application.port.output";
-  static final String SERVICE_LAYER_PACKAGES = BASE_PACKAGE + ".application.service";
+  static final String SERVICE_LAYER_PACKAGES = BASE_PACKAGE + ".application.service..";
```
hannes-mk, line 39: Just curious: What does this do?

PR #150, opened 2024-03-04
Adjust navigation for edit view

malte-laukoetter:


frontend/src/components/amendingLaws/RisAmendingLawInfoHeader.story.vue
```diff
@@ -0,0 +1,38 @@
+<script setup lang="ts">
+import RisAmendingLawInfoHeader from "@/components/amendingLaws/RisAmendingLawInfoHeader.vue"
```
andreasphil, line 2: Nitpick: I think here it's fine to use the relative path because components and stories should always be colocated. That way moving is easier because it doesn't require refactoring. At least that's how I do it 🤷‍♂️

```suggestion
import RisAmendingLawInfoHeader from "./RisAmendingLawInfoHeader.vue"
```

frontend/src/components/controls/RisNavbarSide.vue
```diff
@@ -140,7 +140,7 @@ export interface LevelTwoMenuItem {
   <aside aria-labelledby="sidebarNavigation" class="w-[16rem]">
     <span id="sidebarNavigation" hidden>SideBar Navigation</span>
     <router-link
-      aria-labelledby="toProceduresListButton"
+      aria-labelledby="toOverviewButton"
```
andreasphil: Todo: IMHO this should be:

- Remove the `aria-labelledby`
- Remove the ID on the span that is supposed to be the label
- Leave the span there and add a [`sr-only`](https://tailwindcss.com/docs/screen-readers)

That should accomplish the same (button is labeled) without unnecessary use of aria or IDs, right?

I'm also wondering about the `aria-labelledby` of the aside, I don't think an aside needs a label in the first place and if it does it should probably be a hidden heading or something like that.

(I know you didn't implement this and only adjusted, but would still be grateful if you fixed it. I think this is an old-ish component and in those we sometimes have "unorthodox" accessibility stuff)

frontend/src/composables/useAmendingLaw.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+        defineStore("loaded-amending-law", () => {
+          return {
+            loadedAmendingLaw: ref(),
+            loadAmendingLawByEli,
+          }
+        })(),
+      ),
+    }))
+
+    const { useAmendingLaw } = await import("./useAmendingLaw")
```
andreasphil, line 73: Question: why the async import?

frontend/src/views/AmendingLawArticleEditor.vue
```diff
@@ -38,60 +42,77 @@ function handleGeneratePreview() {
...
-      </button>
-      <button
-        class="ds-button ds-button-small ds-button-tertiary h-fit self-end"
-        @click="handleGeneratePreview"
-      >
-        Vorschau generieren
-      </button>
-    </div>
+    <router-link
+      aria-labelledby="toOverviewButton"
```
andreasphil: Todo: Same as above with the aria labeling

frontend/src/composables/useAmendingLaw.ts
```diff
@@ -0,0 +1,33 @@
...
+  DeepReadonly,
+} from "vue"
+import { AmendingLaw } from "@/services/amendingLawsService"
+
+/**
+ * Get the data of an amending law.
+ * @param eli a reference to the eli for which the law data will be returned. Changing the value of the reference will load the data for the new eli.
+ * @returns A reference to the amending law data or undefined if it is not available (or still loading).
+ */
+export function useAmendingLaw(
```
andreasphil, line 18: Question: I'm not sure how this relates to the Amending Law store, maybe ping me if you have time to quickly discuss?

andreasphil (changes requested):

frontend/src/composables/useAmendingLaw.spec.ts
malte-laukoetter: it allows each test case to mock the store independently. Otherwise the mocks are hoisted out of the individual test cases and are all applied to all test cases.

frontend/src/components/controls/RisNavbarSide.vue
malte-laukoetter: I don't think that accomplishes the same thing as the text is currently visible for everyone. I'm not that knowledgeable about the aria stuff but i thought that this would ensure that the icon is not read out by the screenreader? Do you have another idea on how to solve this? Or is this already not a problem to begin with?

andreasphil: You're right, in this case I think the label is unnecessary altogether for the link.

The aside should IMO actually be a `nav`, and that should be enough to communicate what it is.

malte-laukoetter: i didn't adjust the aside to nav yet as i would first want to figure out if that has any other unexpected changes :)

PR #164, opened 2024-03-07
Add saving of amending laws - Extend frontend page to have button and call that API

malte-laukoetter:
RISDEV-3472

### DIFF
--- a/frontend/src/components/editor/RisCodeEditor.vue
+++ b/frontend/src/components/editor/RisCodeEditor.vue
@@ -1,7 +1,7 @@
 <script setup lang="ts">
-import { EditorView, basicSetup } from "codemirror"
+import { basicSetup, EditorView } from "codemirror"
 import { xml } from "@codemirror/lang-xml"
-import { ref, shallowRef, toRef, watch } from "vue"
+import { computed, ref, shallowRef, toRef, watch } from "vue"
 import { useCodemirrorVueReadonlyExtension } from "@/components/editor/composables/useCodemirrorVueReadonlyExtension"
 import { useCodemirrorVueEditableExtension } from "@/components/editor/composables/useCodemirrorVueEditableExtension"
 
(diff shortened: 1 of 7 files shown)


frontend/src/components/editor/RisCodeEditor.vue
```diff
@@ -64,17 +64,27 @@ const codemirrorVueEditableExtension = useCodemirrorVueEditableExtension(
   toRef(props.editable),
 )
 
+const editorContent = computed(() => editorView.value?.state.doc.toString())
```
andreasphil: Todo: As discussed, this will most likely not work

frontend/src/composables/useArticleXml.ts
```diff
@@ -16,27 +16,49 @@ import {
...
-      // TODO: Switch this to article XML once we have that. For now we'll display
-      // the entire amending law.
-      articleXml.value = await getAmendingLawXmlByEli(is.eli)
+      await load(is)
     },
     { immediate: true },
   )
 
-  return readonly(articleXml)
+  async function refresh() {
```
andreasphil: Todo: As discussed

- [x] Move the update function from the component to the composable
- [x] The API should return the updated data if the call is successful, then use that to update `articleXml`
- [x] Feel free to remove `refresh` after that if you want

frontend/src/services/amendingLawsService.ts
```diff
@@ -29,3 +29,19 @@ export async function getAmendingLawXmlByEli(eli: string): Promise<string> {
...
 }
+
+/**
+ * Save the xml version of an amending law to the API.
+ *
+ * @param eli Eli of the amending law
+ * @param xml New xml of the amending law
+ */
+export async function putAmendingLawXml(eli: string, xml: string) {
+  return await apiFetch.raw<undefined>(`/amending-laws/${eli}`, {
```
andreasphil: Todo: As discussed, I would recommend to use the regular call without `raw` and do any error handling in the `catch` block, see https://github.com/unjs/ofetch?tab=readme-ov-file#%EF%B8%8F-handling-errors

andreasphil (changes requested):

PR #178, opened 2024-03-12
RISDEV-3448 publishing view

malte-laukoetter:


andreasphil:
@malte-laukoetter

Teleport in general is exactly the right strategy for this type of requirement.

The fact that the component that wants to show alerts implements the teleport itself somehow seems wrong to me, as a minimum we should abstract this (e.g. a `RisAlerts` component that then handles this under the hood). But for now I think it's ok, we'll have to rework it anyway.

We still have to think about global vs. contextual alerts at some point, they should also be handled differently. For global ones, I actually prefer an imperative API.

But you can still merge like this for the time being

andreasphil (approved):

malte-laukoetter:
@andreasphil 

> The fact that the component that wants to show alerts implements the teleport itself somehow seems wrong to me, as a minimum we should abstract this (e.g. a `RisAlerts` component that then handles this under the hood). But for now I think it's ok, we'll have to rework it anyway.

I see the `RisAlert` component created in this PR as the visual component for an alert the handling of showing it in a specific place as a second one, but yes the logic created for showing the alerts can be moved into its own component. I'll probably do this.

> We still have to think about global vs. contextual alerts at some point, they should also be handled differently. For global ones, I actually prefer an imperative API.

I agree that an imperatice API for such global alerts would be nice, I just haven't really found a good idea on how to do this. But I have the feeling there is some possibility of doing it using vue's `provide` and `inject` that could be quite convenient 🤔 

Thanks for your thoughts :)

andreasphil:
@malte-laukoetter 

> I see the `RisAlert` component created in this PR as the visual component for an alert the handling of showing it in a specific place as a second one, but yes the logic created for showing the alerts can be moved into its own component. I'll probably do this.

Sounds good :)

> I agree that an imperatice API for such global alerts would be nice, I just haven't really found a good idea on how to do this. But I have the feeling there is some possibility of doing it using vue's `provide` and `inject` that could be quite convenient 🤔

I've implemented this before with provide and inject, basically like this:

- An alerts provider component that wraps the rest of the application, and:
    - Takes care of rendering the alerts
    - Manages showing / hiding alerts etc.
    - Provides a method for pushing alerts
- A simple wrapper composable that does the inject for you just with a nicer name (`inject(xy, fallback)` is a bit annoying compared to `useAlerts`)
- And then whoever wants to show notifications can `const pushAlert = useAlerts();` and start showing alerts 

If you wanted to leave open the option for teleporting to also use alerts declaratively, that could also be done in the same pattern with little effort.

frontend/src/components/controls/RisAlert.vue
```diff
@@ -0,0 +1,53 @@
...
+  close: []
+}>()
+</script>
+
+<template>
+  <div
+    role="alert"
+    class="flex gap-20 border-l-2 p-16"
+    :class="{
+      'border-l-green-700': variant === 'success',
```
andreasphil: Suggestion: This could be simplified as `'border-l-green-700 bg-white': variant === 'success'` (same for the error)

frontend/src/components/controls/RisAlert.vue
```diff
@@ -0,0 +1,53 @@
...
+    }"
+  >
+    <IconCheck
+      v-if="variant === 'success'"
+      width="24"
+      height="24"
+      class="text-green-700"
+    />
+    <IconErrorOutline
+      v-if="variant === 'error'"
```
andreasphil:
```suggestion
      v-else-if="variant === 'error'"
```

frontend/src/composables/useAlerts.ts
```diff
@@ -0,0 +1,79 @@
+import { inject, InjectionKey, provide, readonly, Ref, ref } from "vue"
+
+/**
+ * An alert message
+ */
+export type Alert = {
+  /**
+   * The variant of the alert, determines the visual appear
```
andreasphil:
```suggestion
   * The variant of the alert, determines the visual appearance
```

frontend/src/views/Publishing.vue
```diff
@@ -0,0 +1,142 @@
...
+      />
+    </section>
+
+    <section class="mt-[60px] flex flex-col gap-40">
+      <h1 class="ds-heading-02-reg">Letzte Abgabe</h1>
+
+      <span v-if="publishedAt"
+        >Dieses Änderungsgesetz wurde zuletzt abgegeben am
+        <time :datetime="publishedAt.toISOString()">
+          {{
```
andreasphil: Todo: Function calls should not be inside templates for readability and performance. Maybe a `computed` could be used here instead?

andreasphil (changes requested):

PR #225, opened 2024-04-03
Renderings Endpoint

malte-laukoetter:
We currently create the HTML-render for different laws by using different endpoints and the `Accept: text/html` header. This does not yet support the use case of rendering a not yet saved law for eg. the metadata preview.
We also have two different rendering modes: one including the metadata table and one without this table. Currently the decision which mode to use is based on the endpoint (eg. target-laws get rendered with metadata and amending-laws without metadata). For different parts of our application we need a more fine grained way to decide when to use which render mode (eg. render a target-law without the metadata table).

Therefor a new endpoint is proposed that can handle both of these use cases by:
1. rendering any law whose XML representation is send to the endpoint
2. having a Query parameter that decides which rendering mode to use

As our api is resource based I propose the URL `/renderings` for this endpoint (think "create a new rendering-resource for the given XML")

### DIFF
--- a/doc/backend/norms-api-v1.yaml
+++ b/doc/backend/norms-api-v1.yaml
@@ -204,6 +204,30 @@ paths:
               schema:
                 type: string
 
+  /renderings:
+    post:
+      summary: Create and return an HTML rendering of the provided LDML.de XML
+      parameters:
+        - in: query
+          name: show_metadata
+          schema:
+            type: boolean
+            default: false
+          description: Should the metadata table be included?
+      requestBody:
+        description: The LDML.de XML
+        required: true
+        content:
+          application/xml:
+            schema:
+              $ref: '#/components/schemas/LegalDocML.de'
+      responses:
+        '200':
+          description: HTML rendering the provided LDML.de XML
+          content:
+            text/html:
+              schema:
+                type: string
 
 components:
   schemas:


andreasphil:
Thanks @malte-laukoetter, I get it now 😄 LGTM

I somehow suspect that how we approach the entire rendering/preview stuff will change quite a bit anyhow with the new designs in this EA ...

andreasphil (approved):

doc/backend/norms-api-v1.yaml
```diff
@@ -204,6 +204,30 @@ paths:
               schema:
                 type: string
 
+  /renders:
+    post:
+      summary: Create and return a HTML render of a provided law
```
hannes-mk: Just some minor feedback on the wording:
1. Typo:  "an HTML" instead of "a HTML"
2. "rendering" instead of "render" (I looked this up and "rendering" means "Darstellung / Wiedergabe"; could not find "render")
3. How about "...of the provided LDML.de XML" instead of "a provided law" - the functionality holds for all LDML, not just for laws, I suppose.

doc/backend/norms-api-v1.yaml
```diff
@@ -204,6 +204,30 @@ paths:
...
+      requestBody:
+        description: The XML of the law
+        required: true
+        content:
+          application/xml:
+            schema:
+              $ref: '#/components/schemas/LegalDocML.de'
+      responses:
+        '200':
+          description: HTML render of the given law
```
hannes-mk: If we change the terminology above ("rendering", "LDML.de XML"), then also here.

hannes-mk (approved):

doc/backend/norms-api-v1.yaml
malte-laukoetter: oh true "rendering" is the noun and "render" the verb. I adjusted that (and the other wordings) :)

malte-laukoetter:
@hamo225 could you check if the documentation is still correct (as in the same as your implementation of the endpoint) and then also merge this when you merge https://github.com/digitalservicebund/ris-norms/pull/229 Thanks :)

PR #236, opened 2024-04-08
Added note to readme, that nodejs/npm is needed for spotless to run

reckseba:


backend/README.md
```diff
@@ -11,6 +11,8 @@ brew install openjdk@21
 brew install --cask docker # or just `brew install docker` if you don't want the Desktop app
 ```
 
+Install node for Spotless with Prettier being able to run.
```
hannes-mk: Works for me.

Two minor ideas:

- The expression "... for Spotless with Prettier being able to run" could be simplified by using active voice: " ... for running Spotless with Prettier".
- We generally aim for the latest LTS version of node; we could mention that, here: "Install the latest LTS version of node for ...".

hannes-mk (approved):
Welcome to the code base! :wave:

backend/README.md
reckseba: Thanks for the advise! I committed the change.

reckseba:
Close because commit were not signed.

PR #245, opened 2024-04-10
Remove build script

reckseba:
By simply moving the Docker build command to the Compose file, we could tidy up the project a little more. Do you agree?

### DIFF
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -54,3 +54,21 @@ lefthook install
 
 Opt in to CI posting notifications for failing jobs to a particular Slack channel by setting a repository secret
 with the name `SLACK_WEBHOOK_URL`, containing a url for [Incoming Webhooks](https://api.slack.com/messaging/webhooks).
+
+# How to run locally
+
+With Docker being installed (Compose Plugin needed) run following to start all containers:
+```bash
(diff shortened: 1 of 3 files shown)


VictorDelCampo:
@reckseba it is a good idea although we would then be building the ris-norms image every time we run the docker compose right? And maybe you don't need to build the image again but rather just restart the services? 

Also documenting the docker-compose start-up is great and I would also do it in DEVELOPING.md but I would also add a sentence about the other way we have to start the application for local development (redis+postgres as containers but then using npm run dev and the gradle wrapper)

reckseba:
@VictorDelCampo no that's not what happens. It will take the image from the local registry. Only on `docker compose build` you can force it to build. Though it will re-use cached layers which can be forced to re-create by adding `--no-cache` argument. The exact docu part can be found [here](https://docs.docker.com/compose/compose-file/build/#using-build-and-image).

For the second part: I can add that, too.

VictorDelCampo:
Got it. Thanks for the explanation. Then yes!

VictorDelCampo (approved):

hannes-mk:
:+1:  for getting rid of a "magic" shell script

PR #251, opened 2024-04-12
Delete ds store and ignore it

reckseba:


hannes-mk:
@reckseba  I had no idea what that file would be relevant for before googling.

On the web I found https://en.wikipedia.org/wiki/.DS_Store, so I'd rather **refer to a dev with a Mac** for judging on whether it may or may not be in the repo by mistake (as I am assuming).

Also, I did not check what commit it came with which may indicate whether it was on purpose or not.

reckseba:
@hannes-mk you are right. I could have checked the history to address the right person. Did that now. @derjan4g would it be okay for you not to check-in .DS_Store?

derjan4g:
@reckseba Sorry, it seems as VSC automatically added it. Will find out how to configure VSC to not commit it again. Should I delete it, or are you taking care of that?

reckseba:
@derjan4g already taken care of. I added it to the .gitignore file. :)

PR #262, opened 2024-04-16
Adjust endpoints for new domain model

malte-laukoetter:
RISDEV-3645

### DIFF
--- a/backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/AmendingLawService.java
+++ b/backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/AmendingLawService.java
@@ -28,6 +28,7 @@ public class AmendingLawService
   private final LoadArticlesPort loadArticlesPort;
   private final LoadArticlePort loadArticlePort;
   private final UpdateAmendingLawXmlPort updateAmendingLawXmlPort;
+  private final UpdateNormXmlUseCase updateNormXmlUseCase;
 
   public AmendingLawService(
       LoadAmendingLawPort loadAmendingLawPort,
@@ -35,13 +36,15 @@ public class AmendingLawService
       LoadAllAmendingLawsPort loadAllAmendingLawsPort,
       LoadArticlesPort loadArticlesPort,
       LoadArticlePort loadArticlePort,
-      UpdateAmendingLawXmlPort updateAmendingLawXmlPort) {
+      UpdateAmendingLawXmlPort updateAmendingLawXmlPort,
+      UpdateNormXmlUseCase updateNormXmlUseCase) {
     this.loadAmendingLawPort = loadAmendingLawPort;
     this.loadAmendingLawXmlPort = loadAmendingLawXmlPort;
     this.loadAllAmendingLawsPort = loadAllAmendingLawsPort;
     this.loadArticlesPort = loadArticlesPort;
     this.loadArticlePort = loadArticlePort;
     this.updateAmendingLawXmlPort = updateAmendingLawXmlPort;
+    this.updateNormXmlUseCase = updateNormXmlUseCase;
   }
 
   @Override
@@ -73,6 +76,12 @@ public class AmendingLawService
 
   @Override
   public Optional<String> updateAmendingLawXml(UpdateAmendingLawXmlUseCase.Query query) {
+    try {
+      updateNormXmlUseCase.updateNormXml(new UpdateNormXmlUseCase.Query(query.eli(), query.xml()));
+    } catch (UpdateNormXmlUseCase.InvalidUpdateException e) {
+      throw new RuntimeException(e);
+    }
+
     return updateAmendingLawXmlPort.updateAmendingLawXmlByEli(
         new UpdateAmendingLawXmlPort.Command(query.eli(), query.xml()));
   }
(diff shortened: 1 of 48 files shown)


backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/AmendingLawService.java
```diff
@@ -73,6 +78,13 @@ public Optional<String> loadAmendingLawXml(LoadAmendingLawXmlUseCase.Query query
 
   @Override
   public Optional<String> updateAmendingLawXml(UpdateAmendingLawXmlUseCase.Query query) {
+    var updatedNorm = Norm.builder().document(XmlMapper.toDocument(query.xml())).build();
```
hannes-mk: Curious: do we have a (silent) agreement on `var`?

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/AmendingLawService.java
```diff
@@ -73,6 +78,13 @@ public Optional<String> loadAmendingLawXml(LoadAmendingLawXmlUseCase.Query query
...
+    if (updatedNorm.getEli().isPresent() && updatedNorm.getEli().get().equals(query.eli())) {
```
hannes-mk: Is this conditional being tested?
(It makes good sense)

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/AmendingLawService.java
```diff
@@ -73,6 +78,13 @@ public Optional<String> loadAmendingLawXml(LoadAmendingLawXmlUseCase.Query query
...
+      updateNormPort.updateNorm(
+          new UpdateNormPort.Command(
+              Norm.builder().document(XmlMapper.toDocument(query.xml())).build()));
```
hannes-mk: This looks just like `updatedNorm` above. Can we reuse it?

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/ReleaseService.java
```diff
@@ -16,21 +18,40 @@
...
     this.loadAmendingLawPort = loadAmendingLawPort;
+    this.loadAnnouncementPort = loadAnnouncementPort;
     this.updateAmendingLawPort = updateAmendingLawPort;
+    this.updateAnnouncementPort = updateAnnouncementPort;
   }
 
   @Override
   public Optional<AmendingLaw> releaseAmendingLaw(Query query) {
-    final Optional<AmendingLaw> amendingLawOptional =
+    var amendingLawOptional =
```
hannes-mk: Mabe move this code closer to the `return` statement.
Then we have one block for the annoucements and one for the amendingLaws.

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/ReleaseService.java
```diff
@@ -16,21 +18,40 @@
...
+        loadAnnouncementPort.loadAnnouncement(new LoadAnnouncementPort.Command(query.eli()));
+    announcementOptional.flatMap(
+        announcement -> {
+          announcement.setReleasedByDocumentalistAt(Instant.now());
+
+          return updateAnnouncementPort.updateAnnouncement(
+              new UpdateAnnouncementPort.Command(announcement));
+        });
+
     return amendingLawOptional.flatMap(
```
hannes-mk, line 53: Does this now always return the amendingLaw, even in the "annoucmenets" case?

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/TargetLawService.java
```diff
@@ -61,6 +67,14 @@ public Optional<String> applyTimeMachine(TimeMachineUseCase.Query query) {
 
   @Override
   public Optional<String> updateTargetLaw(UpdateTargetLawUseCase.Query query) {
+    var updatedNorm =
+        Norm.builder().document(XmlMapper.toDocument(query.newTargetLawXml())).build();
+    if (updatedNorm.getEli().isPresent() && updatedNorm.getEli().get().equals(query.eli())) {
+      updateNormPort.updateNorm(
+          new UpdateNormPort.Command(
+              Norm.builder().document(XmlMapper.toDocument(query.newTargetLawXml())).build()));
```
hannes-mk: Is this the same as `updatedNorm` above?

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/AmendingLawService.java
malte-laukoetter: I'm not aware that we have any agreement on using (or not using) it

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/ReleaseService.java
malte-laukoetter: it is just ensuring the announcement is kept in sync with the existing amending law. So yes it will always return the amending law. That the announcement is also updated should be transparent to the uses.
The release service will later be changed for the changes to the domain model.

PR #278, opened 2024-04-22
Container name refactoring

reckseba:
There are three published container packages:
- [ris-norms/backend](https://github.com/digitalservicebund/ris-norms/pkgs/container/ris-norms%2Fbackend)
- [ris-norms/frontend](https://github.com/digitalservicebund/ris-norms/pkgs/container/ris-norms%2Ffrontend) (outdated)
- [ris-norms](https://github.com/digitalservicebund/ris-norms/pkgs/container/ris-norms) (outdated)

Regarding the names I'd assume `ris-norms`to be the one containing both, but that's not true anymore.

This PR renames `ris-norms/backend` -> `ris-norms`.

[This](https://github.com/digitalservicebund/ris-norms-infra/pull/5) is the according PR in [ris-norms-infra](https://github.com/digitalservicebund/ris-norms-infra).


VictorDelCampo (commented):
Hey Basti you made the necessary changes and in my opinion the goal is reached.

But maybe you have more time and can think about:

- Does it make sense to have the creation of the frontend+backend image in the backend-jobs? I just realized that we do still have some "backend" naming for the image (steps "save image" and the following unnamed one creating the cache) and the question popped out in my mind

Also at the end we should also manually removed the old images pushed here so that we only have the `ris-norms` image.

reckseba:
Same here. I totally agree. Very nice to see that everybody on the team is willing to clean up even if it works right now. :100:

PR #295, opened 2024-04-23
Unpin chainguard images, use 'latest'

hannes-mk:
This allows for updating chainguard versions automatically, i.e. without them getting flagged by talisman.
All images we're using are reported as "Verified Publishers" in the docker registry:

- Chainguard/JRE https://hub.docker.com/r/chainguard/jre
- Chainguard/Redis https://hub.docker.com/r/chainguard/redis
- Chainguard/Nginx https://hub.docker.com/r/chainguard/nginx


hannes-mk:
@VictorDelCampo This is quite similar to not pinning the action we're using ([this PR](https://github.com/digitalservicebund/ris-norms/pull/274)). Would you agree to the reasoning why we don't need pinning?

VictorDelCampo:
@hannes-mk I guess you are proposing this because of the high CVE we had recently? I do see therefore the closest benefit since we would have had the fix version without doing something....

Nevertheless I am not 100% sure anymore of the decision of "trusting the creators" even with the Github badge, since Github is also writing:
_Although pinning to a commit SHA is the most secure option, specifying a tag is more convenient and is widely used. If you’d like to specify a tag, then be sure that you trust the action's creators. The ‘Verified creator’ badge on GitHub Marketplace is a useful signal, as it indicates that the action was written by a team whose identity has been verified by GitHub. Note that there is risk to this approach even if you trust the author, because a tag can be moved or deleted if a bad actor gains access to the repository storing the action._
Meaning we can't never be sure even if the creators can be trusted but on other side pinning with a digest would also not be enough if we haven't audited what the content is....

Would be great if we involve more devs. Maybe @malte-laukoetter or @SebastianRossa or @reckseba ?

And by the way, I think we don't need the frontend Dockerfile anymore...

hannes-mk:
Thanks, @VictorDelCampo ! 
I'll bring it up in a tech daily.

hannes-mk:
In the tech daily, we decided to go this way; especially as https://digitalservicebund.slack.com/archives/C01JUUDHG5Q/p1713161252315109?thread_ts=1712044976.739109&cid=C01JUUDHG5Q gave sort of a Platform perspective "yes".

PR #296, opened 2024-04-23
Replace amending law and target law services with norm service

malte-laukoetter:


andreasphil (commented):
Great work @malte-laukoetter!

PR #299, opened 2024-04-24
Return HTML of target law at time boundary - RISDEV-3843

hannes-mk:
Note that this only implements returning HTML on a valid `atIsoDate` parameter.
It does not actually apply the time machine. We have a separate task for that.


backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/controller/NormController.java
```diff
@@ -136,7 +138,17 @@ public ResponseEntity<String> getNormRender(
       @PathVariable final String version,
       @PathVariable final String language,
       @PathVariable final String subtype,
-      @RequestParam(defaultValue = "false") boolean showMetadata) {
+      @RequestParam(defaultValue = "false") boolean showMetadata,
+      @RequestParam Optional<String> atIsoDate) {
+
+    if (atIsoDate.isPresent()) {
+      try {
+        java.time.format.DateTimeFormatter.ISO_DATE_TIME.parse(atIsoDate.get());
```
hannes-mk: import the DateTimeFormatter

PR #308, opened 2024-04-30
Error tooltip

andreasphil:
Implements RISDEV-3800:

- Callouts can be dismissed
- Tooltip component for showing small floating callouts next to components
- Tooltips integrated into the button

### DIFF
--- a/frontend/src/composables/useElementId.ts
+++ b/frontend/src/composables/useElementId.ts
@@ -1,17 +1,21 @@
 import { Ref, readonly, ref } from "vue"
 
-function createUseElementId() {
+function createUseElementId(): (prefix?: string) => Readonly<Ref<string>> {
   /**
    * Used for keeping track of existing IDs in the unlikely event that
    * we generate a duplicate one.
    */
   const currentIds = new Set<string>()
 
-  function getIdentifier() {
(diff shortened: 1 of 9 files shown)


frontend/src/composables/useElementId.ts
```diff
@@ -0,0 +1,33 @@
+import { Ref, readonly, ref } from "vue"
+
+function createUseElementId() {
+  /**
+   * Used for keeping track of existing IDs in the unlikely event that
+   * we generate a duplicate one.
+   */
+  const currentIds = new Set<string>()
+
+  function getIdentifier() {
```
malte-laukoetter: I'm quite unsure how good it is that the id now starts with a number and includes a `.` as both is (while technically supported) not recommended as there are a couple of pitfalls that one could fall into with them: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id

malte-laukoetter (approved):

frontend/src/composables/useElementId.ts
andreasphil: It doesn't contain dots (those are "multiplied away") but good catch about the numbers. I'll fix that!

PR #313, opened 2024-05-03
Docs: Refer to general LDML.de norm instead of specific "amending law"

hannes-mk:
RISDEV-0000


hannes-mk:
@SebastianRossa as you were highly involved: Fine with you?

PR #315, opened 2024-05-03
Test: Fix call to intermediate value

hannes-mk:
RISDEV-0000


hannes-mk:
@andreasphil I guess the test mixed up two variables.

andreasphil (approved):

PR #318, opened 2024-05-06
Add E2E Tests for MetaData features (RISDEV-3900)

hannes-mk:
Part of this is fixing the current test/default data


hannes-mk:
FYI: @andreasphil

PR #319, opened 2024-05-06
e2e tests for temporal data - RISDEV-3848

hamo225:


VictorDelCampo:
Hey @hamo225, this is my review:

1. I think we need a new e2e test for the side navigation where we test it separate? Or we just keep also the navigation within this test but then I would expect to be two describe-layers:
    - test navigation to time boundaries using navigation starting on `/amending-laws/eli/bund/bgbl-1/2017/s419/2017-03-15/1/deu/regelungstext-1` and just check if you arrived at .`../temporal-data`
    - test the actual page including:
       - heading
       - Artikel 3Inkraftrete
       - the managing of the data itself meaning add/update/delete
2. For testing the heading I would use if possible something more readable like:
```
   await expect(
      page.getByRole("heading", { level: 1, name: "Artikel 1" }),
    ).toBeVisible()
```
3. Why do you need to do 
```
    await page.waitForResponse(
      (response) =>
        response.url().includes("timeBoundaries") && response.status() === 200,
    )
```
   Is the await saveButton.click() not enough for waiting? But this could be related to the next point

4. Why do you do the API calls to test the response JSON? In my opinion the E2E tests are done user-facing meaning what the user would se. I would just add/update/delete data, save, reload page and see if the data is there

frontend/e2e/navigate-to-temporal-data-page-and-test-content.spec.ts
```diff
@@ -1,34 +1,96 @@
...
-  }) => {
-    const dateInput = page.locator('[data-testid="date-input-field"]')
-    await expect(dateInput).toHaveCount(1)
-    const inputValue = await dateInput.inputValue()
+    const saveButton = page.locator("text=Speichern")
+    let dateInputs = page.locator('[data-testid="date-input-field"]')
+
+    await expect(dateInputs).toHaveCount(1)
+    const inputValue = await dateInputs.inputValue()
     await expect(inputValue).toBe("16.03.2017")
```
reckseba: No `await` needed, because `toBe` does not return a promise.

frontend/e2e/navigate-to-temporal-data-page-and-test-content.spec.ts
```diff
@@ -1,34 +1,96 @@
...
+    await expect(dateInputs).toHaveCount(3)
+    const dateInputToEdit = page
+      .locator('[data-testid="date-input-field"]')
+      .nth(1)
+    await dateInputToEdit.fill("03.06.2023")
+    await saveButton.click()
+    await page.reload()
+
+    const editedValue = await dateInputs.nth(1).inputValue()
+    await expect(editedValue).toBe("03.06.2023")
```
reckseba: No `await` needed, because `toBe` does not return a promise.

frontend/e2e/navigate-to-temporal-data-page-and-test-content.spec.ts
```diff
@@ -1,34 +1,96 @@
...
+    let dateInputs = page.locator('[data-testid="date-input-field"]')
+
+    await expect(dateInputs).toHaveCount(1)
+    const inputValue = await dateInputs.inputValue()
     await expect(inputValue).toBe("16.03.2017")
+
+    //   add new time boundaries
+    const newDateInput = page.locator('[data-testid="new-date-input-field"]')
+    await newDateInput.fill("01-05-2023")
+    await newDateInput.fill("02-06-2023")
```
reckseba: Did you fill by using dashes (01-05-2023) on purpose to check that this works? My first intuition would be to start with the happy case with the correct German dot format (01.05.2023). I'd put input validation/rejection/acceptance in a separate test.

frontend/e2e/navigate-to-temporal-data-page-and-test-content.spec.ts
```diff
@@ -1,34 +1,96 @@
...
+      (response) =>
+        response.url().includes("timeBoundaries") && response.status() === 200,
+    )
+
+    await page.reload()
+
+    await expect(dateInputs).toHaveCount(3)
+    //edit time boundaries
+
+    await expect(dateInputs).toHaveCount(3)
```
reckseba: Duplication of line 54?

frontend/e2e/navigate-to-temporal-data-page-and-test-content.spec.ts
```diff
@@ -1,34 +1,96 @@
...
+
+    //   add new time boundaries
+    const newDateInput = page.locator('[data-testid="new-date-input-field"]')
+    await newDateInput.fill("01-05-2023")
+    await newDateInput.fill("02-06-2023")
+    await saveButton.click()
+
+    await page.waitForResponse(
+      (response) =>
+        response.url().includes("timeBoundaries") && response.status() === 200,
```
reckseba: As you already have the response at hand, you could check for string equality with the expected response body: `[{"date":"2017-03-16","eventRefEid":"meta-1_lebzykl-1_ereignis-2"},{"date":"2023-05-01","eventRefEid":"meta-1_lebzykl-1_ereignis-3"},{"date":"2023-06-02","eventRefEid":"meta-1_lebzykl-1_ereignis-4"}]`

frontend/e2e/navigate-to-temporal-data-page-and-test-content.spec.ts
```diff
@@ -1,34 +1,96 @@
...
+    await expect(dateInputs).toHaveCount(3)
+    //edit time boundaries
+
+    await expect(dateInputs).toHaveCount(3)
+    const dateInputToEdit = page
+      .locator('[data-testid="date-input-field"]')
+      .nth(1)
+    await dateInputToEdit.fill("03.06.2023")
+    await saveButton.click()
+    await page.reload()
```
reckseba, line 81: Why here no `waitForResponse` like you did before? Was that not the reason to prevent a race condition -> flaky tests? Btw could check for response body `[{"date":"2017-03-16","eventRefEid":"meta-1_lebzykl-1_ereignis-2"},{"date":"2023-06-03","eventRefEid":"meta-1_lebzykl-1_ereignis-3"},{"date":"2023-06-02","eventRefEid":"meta-1_lebzykl-1_ereignis-4"}]`

frontend/e2e/navigate-to-temporal-data-page-and-test-content.spec.ts
```diff
@@ -1,34 +1,96 @@
...
+    const editedValue = await dateInputs.nth(1).inputValue()
+    await expect(editedValue).toBe("03.06.2023")
+
+    //   delete time boundaries
+
+    for (let i = 2; i > 0; i--) {
+      const deleteButton = page.locator(`[data-testid="delete-button-${i}"]`)
+      await deleteButton.click()
+      await saveButton.click()
+      await page.reload()
```
reckseba, line 91: Same `waitForResponse` as above.

1st expectation: `[{"date":"2017-03-16","eventRefEid":"meta-1_lebzykl-1_ereignis-2"},{"date":"2023-06-03","eventRefEid":"meta-1_lebzykl-1_ereignis-3"}]`

2nd expectation `[{"date":"2017-03-16","eventRefEid":"meta-1_lebzykl-1_ereignis-2"}]`

Of course then this nice loop wouldn't work anymore :D

frontend/e2e/navigate-to-temporal-data-page-and-test-content.spec.ts
```diff
@@ -1,34 +1,96 @@
...
+      dateInputs = page.locator('[data-testid="date-input-field"]')
+      await expect(dateInputs).toHaveCount(i)
+    }
+
+    const deleteButton = page.locator(`[data-testid="delete-button-0"]`)
+    const isDisabled = await deleteButton.isDisabled()
+    expect(isDisabled).toBe(true)
+
+    await page.request.put(
```
reckseba: What about this PUT request? You are not expecting anything from it. And it does not change anything on the state.

frontend/e2e/navigate-to-temporal-data-page-and-test-content.spec.ts
hamo225: good point, lets revisit when we add validation

frontend/e2e/navigate-to-temporal-data-page-and-test-content.spec.ts
hamo225: Removed this. The playwright e2e tests should have this focus of what we expect a user to experience. Meaning we do not really need to test the payloads. I have instead added checks that the inputs display the expected values.

frontend/e2e/navigate-to-temporal-data-page-and-test-content.spec.ts
hamo225: we can discuss

frontend/e2e/navigate-to-temporal-data-page-and-test-content.spec.ts
hamo225: The PUT is not part of the user interaction. Its for the test, to set up and reset the environment. I have moved this into a function that will be called at the start and end of the test. This whole setup for the tests will change soon i hope.

PR #322, opened 2024-05-07
Docker playwright

reckseba:


frontend/README.md
```diff
@@ -39,3 +39,19 @@ If the standalone TypeScript plugin doesn't feel fast enough to you, Volar has a
...
+- Make sure the backend runs
+- Make sure the frontend runs
+- Install Playwright
+- Let Playwright now where to connect to
+- Run the E2E Test
+
+```bash
+npx --yes playwright install --with-deps chrome firefox webkit
+
+E2E_BASE_URL=http://localhost:5173 npm run test:e2e
```
VictorDelCampo: You wouldn't need to pass that variable if you have the .env file. There is an example in the frontend folder

frontend/README.md
```diff
@@ -39,3 +39,19 @@ If the standalone TypeScript plugin doesn't feel fast enough to you, Volar has a
    1. Run `Extensions: Show Built-in Extensions` from VSCode's command palette
    2. Find `TypeScript and JavaScript Language Features`, right click and select `Disable (Workspace)`
 2. Reload the VSCode window by running `Developer: Reload Window` from the command palette.
+
+## Setup Playwright for E2E Tests
```
VictorDelCampo, line 43: You could also add how to run just 1 specific test:

npm run test:e2e -- [testfile]

And also maybe add how to debug an e2e test:

npm run test:e2e -- [testfile] --debug

docker-compose.yaml
```diff
@@ -1,7 +1,15 @@
 services:
+  nginx:
```
VictorDelCampo, line 2: Should we somehow make the usage of the nginx a standard for developers using the compose setup? meaning only the nginx should be access from the outside and the webapp can only be access through the docker network?

reckseba: Sounds appropriate but didn't want to disrupt to many things at the same time. Don't know if there are other implications. How about leaving it optional for now and we discuss it in a tech daily if anybody objects?

VictorDelCampo: Fine for me

VictorDelCampo (approved):

PR #328, opened 2024-05-10
It shall not be possible to add the same TimeBoundary Date twice

reckseba:


backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/constraints/UniqueTimeBoundariesDatesConstraintValidator.java
```diff
@@ -0,0 +1,18 @@
...
+import java.util.HashSet;
+import java.util.List;
+
+/** Validates if TimeBoundary dates are unique */
+public class UniqueTimeBoundariesDatesConstraintValidator
+    implements ConstraintValidator<UniqueTimeBoundariesDatesConstraint, List<TimeBoundarySchema>> {
+  @Override
+  public boolean isValid(
+      List<TimeBoundarySchema> timeBoundaries, ConstraintValidatorContext context) {
+    return new HashSet<>(timeBoundaries.stream().map(TimeBoundarySchema::getDate).toList()).size()
```
VictorDelCampo: I love your approach @reckseba. It is using for sure a HashSet where when instantiating duplicates will automatically be removed but in my opinion it is not that readable.

What do you think about:

`return timeBoundaries.stream().map(TimeBoundarySchema::getDate).distinct().count() == timeBoundaries.size()`

This way we only rely on stream operations and don't need to create an intermediate in -memory data structure (HashSet)

But it is just my opinion, I don't think we won't ever had any performance issues because of this. It is for you to decide. Otherwise the rest is great!

backend/src/test/java/de/bund/digitalservice/ris/norms/integration/adapter/input/restapi/TimeBoundaryControllerIntegrationTest.java
```diff
@@ -269,6 +269,62 @@ void itCallsUpdateTimeBoundariesDateMalformed() throws Exception {
           .andExpect(status().isBadRequest());
     }
 
+    @Test
```
VictorDelCampo, line 272: Should we also be testing the actual response content or just the bad request status? And why did you decide to write an integration test and not a controller unit test if the goal was to test a validation happening only in the controller?

VictorDelCampo (commented):
I really like your approach Basti but maybe you could have a look at my comments?

PR #334, opened 2024-05-13
Implement node selecting component

malte-laukoetter:


frontend/src/components/RisLawPreview.spec.ts
```diff
@@ -10,4 +10,123 @@ describe("RisLawPreview", () => {
...
+        content:
+          "<div><span class='longTitle'>Test Title</span><div class='akn-mod' data-eId='hauptteil-1_art-1_abs-1_untergl-1_listenelem-1_inhalt-1_text-1_ändbefehl-1' data-GUID='148c2f06-6e33-4af8-9f4a-3da67c888510'>MOD</div></div>",
+      },
+      attrs: {
+        "onClick:akn:mod": handler,
+      },
+    })
+
+    await waitFor(() => screen.getByRole("button"))
+    await fireEvent.click(screen.getByRole("button", { name: "MOD" }))
```
andreasphil, line 28: Question: Did you use `fireEvent` here for a specific reason? Usually we use [userEvent](https://testing-library.com/docs/user-event/intro/).

frontend/src/components/RisLawPreview.vue
```diff
@@ -24,18 +26,177 @@ withDefaults(
...
+
+const emit = defineEmits<{
+  /**
+   * Event fired when clicking on an HTML-element for the specified LDML.de element.
+   *
+   * Event handlers, a11y properties and so on are only created for elements for which an event handler is registered.
+   *
+   * E.g. `@click:akn:article` will fire for every click on a part of the preview that is showing an article.
+   */
+  [key: `click:akn:${string}`]: [
```
andreasphil, line 50: Question: Does this work with TypeScript, Intellisense etc.? We should do it like this either way but it would be really cool if the tooling was smart enough to pick this up

frontend/src/components/RisLawPreview.vue
```diff
@@ -24,18 +26,177 @@ withDefaults(
...
+              htmlElement,
+              (event) => {
+                emit(`click:akn:${aknElement}`, {
+                  eid,
+                  guid,
+                  originalEvent: event,
+                })
+              },
+              {
+                signal: abortController.signal,
```
andreasphil, line 151: Curious: what does the `AbortController` do here?

andreasphil (commented):
Great work @malte-laukoetter 🙌

frontend/src/components/RisLawPreview.spec.ts
malte-laukoetter: didn't knew of that so I'll change it :)

frontend/src/components/RisLawPreview.vue
malte-laukoetter: sadly it doesn't work with most of those things :/

frontend/src/components/RisLawPreview.vue
malte-laukoetter: Once the abort controller aborts the event listener ist removed. (So it basically replaces the stuff one would otherwise need to do with removeEventListener)

PR #335, opened 2024-05-13
Add endpoint   /norms/{eli}/elements?type=type1&type=type2&amendedBy={eli}

hannes-mk:
RISDEV-3890


backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/controller/ElementsController.java
```diff
@@ -0,0 +1,107 @@
...
+            @PathVariable final String subtype,
+            @RequestParam final ElementType type) {
+
+        // TODO Hannes: Where to put this?
+        // get norm
+        final String eli =
+                buildEli(agent, year, naturalIdentifier, pointInTime, version, language, subtype);
+
+
+        var norm = loadNormUseCase.loadNorm(new LoadNormUseCase.Query(eli));
```
hannes-mk: @andreasphil Here's how to make it work without any streams or maps. We can use it as a starting point.

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/controller/ElementsController.java
```diff
@@ -0,0 +1,107 @@
...
+
+        if (norm.isEmpty()) return ResponseEntity.notFound().build();
+
+        var articleNodes = NodeParser.getNodesFromExpression("//body//article", norm.get().getDocument());
+
+        var eids = articleNodes.stream().map(node -> NodeParser.getValueFromExpression("./@eId", node)).flatMap(Optional::stream).toList();
+
+        var elements = eids.stream().map(eid ->
+                (ElementsResponseEntrySchema) ElementsResponseEntrySchema.builder()
```
hannes-mk: @andreasphil Even in this version, I had to type cast the result of the builder. Strange.

PR #338, opened 2024-05-14
RISDEV-3932 implement data extraction from xml

malte-laukoetter:


frontend/src/services/ldmldeModService.ts
```diff
@@ -0,0 +1,62 @@
...
+  const temporalGroupNode = getNodeByEid(xml, temporalGroupEid)
+  if (!temporalGroupNode) return null
+
+  const eventRefEid = getStartEventRefEid(temporalGroupNode)
+  if (!eventRefEid) return null
+
+  const eventRefNode = getNodeByEid(xml, eventRefEid)
+  if (!eventRefNode) return null
+
+  return getEventRefDate(eventRefNode)
```
reckseba, line 61: @malte-laukoetter why not checking here as well?

frontend/src/services/ldmldeModService.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+                    <akn:timeInterval GUID="ca9f53aa-d374-4bec-aca3-fff4e3485179" eId="meta-1_geltzeiten-1_geltungszeitgr-1_gelzeitintervall-1" refersTo="geltungszeit" start="#meta-1_lebzykl-1_ereignis-2"/>
+                 </akn:temporalGroup>
+              </akn:temporalData>
+            </akn:meta>
+          </akn:act>
+        </akn:akomaNtoso>
+      `)
+
+      expect(
+        getTimeBoundaryDate(
```
reckseba, line 119: Shall we also test the branches where getTimeBoundaryDate returns null?

malte-laukoetter: We could but as they are quite trivial I didn't see the point in it ^^

frontend/src/services/ldmldeModService.ts
malte-laukoetter: It returns null (or maybe undefined I'm not 100% sure) either way so it would change (basically?) nothing

malte-laukoetter:
Thanks for looking over it :)

frontend/src/services/ldmldeModService.ts
reckseba: true

frontend/src/services/ldmldeModService.spec.ts
reckseba: Maybe test-coverage but not quite sure if sonar is calculating on that

PR #339, opened 2024-05-14
Refactor: Move buildEli into its own class

hannes-mk:
Remove duplication

### DIFF
--- a/backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/controller/AnnouncementController.java
+++ b/backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/controller/AnnouncementController.java
@@ -1,5 +1,6 @@
 package de.bund.digitalservice.ris.norms.adapter.input.restapi.controller;
 
+import static de.bund.digitalservice.ris.norms.utils.EliBuilder.buildEli;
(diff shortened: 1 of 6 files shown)


hannes-mk:
How about builder() ?

andreasphil:
As discussed: consider using a builder instead to avoid having such a long list of parameters. Just a suggestion though, looks good to me otherwise.

andreasphil (approved):

malte-laukoetter:
> How about builder() ?

as the method name? if you want something other than the current one i would go with `build()`

Also I agree with Andreas that it looks good to me! and that using the builder-pattern might make the many parameters a bit simpler to use :)

malte-laukoetter (approved):

reckseba:
Thanks for the invite. As the main goal was to remove duplication, I'd be okay with not refactoring it by using a builder pattern. The fact that we have more test coverage now, is already pretty nice.

reckseba (approved):
Thanks for the invite. As the main goal was to remove duplication, I'd be okay with not refactoring it by using a builder pattern. The fact that we have more test coverage now, is already pretty nice.

hannes-mk:
@reckseba Thanks, Sebastian! Your answer made me pause and reconsider priorities. I have decided to not fine-tune this one, but focus on other tasks. Will merge.

PR #342, opened 2024-05-17
RISDEV-3967 update target law xml

malte-laukoetter:


malte-laukoetter:
Sorry, this got a bit larger than i hoped for.

The interesting stuff happens in backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/UpdateNormService.java and backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/Norm.java Almost everything else is the creation of a couple of Domain-Classes to simplify the stuff that happens in these places:
* EId and EIdPart for handling eids
* Href for handling absolute and relative links
* TemporalGroup & ActiveModification for the respective LDML.de elements

PR #344, opened 2024-05-17
Add function to reset test data for e2e tests

SebastianRossa:
Adds the functionality of running SQL scripts from playwright e2e tests


frontend/e2e/database/db.ts
```diff
@@ -0,0 +1,38 @@
+import pg from "pg"
+
+export default class DB {
+  private DBConfig = {
+    user: "test",
```
andreasphil, line 5: Nitpick: I know these credentials are just our local DB stuff and most likely going to be always the same but it still feels like this should be in the env. Not a required change though, just something to consider.

frontend/e2e/database/resetDatabase.ts
```diff
@@ -0,0 +1,23 @@
...
+    console.log(
+      `---------> X Failed to connect to dataBase <--------- \n\n ${error}`,
+    )
+    process.exit(1)
+  }
+}
+
+async function resetDatabase() {
+  await establishDBConnection()
+  const sql = fs.readFileSync("e2e/database/initial_data.sql").toString()
```
andreasphil, line 19: Todo: Could we implement this instead by reading the XML files? Not a fan of these huge SQL files, they're impossible to maintain. Also, do you think it would make sense to share the data between backend and E2E tests? Or do we think those will be different?

frontend/e2e/navigate-to-publishing-page-and-test-release.spec.ts
```diff
@@ -1,5 +1,14 @@
 import { test, expect, Page } from "@playwright/test"
 import { readFileSync } from "fs"
+import resetDatabase from "@e2e/database/resetDatabase"
+
+test.beforeAll(async () => {
```
andreasphil, line 5: Todo: Is this something we expect to want to do in all test suites? If so, please move this to the [global setup and teardown](https://playwright.dev/docs/test-global-setup-teardown).

frontend/e2e/navigate-to-temporal-data-page-and-test-content.spec.ts
```diff
@@ -1,97 +1,89 @@
...
+    test(`adding, editing, deleting and saving time boundaries`, async ({
+      page,
+    }) => {
+      await page.goto(
+        "/amending-laws/eli/bund/bgbl-1/2017/s419/2017-03-15/1/deu/regelungstext-1/temporal-data",
+      )
+      // check contents of entry into force article html rendering
+      await expect(page.getByText("Artikel 3Inkrafttreten")).toBeVisible()
+
+      const saveButton = page.locator("text=Speichern")
```
andreasphil: Todo: This is not related to the SQL changes but some general feedback on the selectors: We should avoid using `page.locator` when possible. The things it's doing here can just as well be accomplished by a more readable and precise selector (e.g. `getByText` and `getByTestId`, or in this case even better: `getByRole("button", { name: "Speichern" })`). This is the [recommended best practice](https://playwright.dev/docs/locators#quick-guide) by Playwright whereas using `locator` is discouraged unless there is no other way to do it. Also see this [thread](https://digitalservicebund.slack.com/archives/C046VD44ZEH/p1715750128424729) in Slack for context.

frontend/e2e/navigate-to-temporal-data-page-and-test-content.spec.ts
```diff
@@ -1,97 +1,89 @@
...
+      let dateInputs = page.locator('[data-testid="date-input-field"]')
```
andreasphil: Todo: Related to the feedback above, there is no reason to use a `data-testid` here or in any other case in this file. You should be able to accomplish the same by selecting for textbox + label. If that is not possible then the issue is with our markup (because it's not accessible), not with the test.

andreasphil (changes requested):
Requesting some smaller changes but looking good generally. I'm adding a label not to merge this though while we wait for a decision on this topic.

frontend/e2e/database/db.ts
SebastianRossa: I generally aggree but since we have this information already in our docker-compose files and it can't be misused I'd leave it like that for now.

VictorDelCampo:
Agreed on Dev Weekly 02.10.2024

PR #349, opened 2024-05-21
Element preview to support at iso date risdev 3989

hannes-mk:


andreasphil (approved):
LGTM when we consider that we were planning to refactor this anyways.

PR #357, opened 2024-05-24
refactor: move loading individual element HTML to service

andreasphil:
RISDEV-3988

### DIFF
--- a/backend/src/test/java/de/bund/digitalservice/ris/norms/application/service/ElementServiceTest.java
+++ b/backend/src/test/java/de/bund/digitalservice/ris/norms/application/service/ElementServiceTest.java
@@ -1,10 +1,12 @@
 package de.bund.digitalservice.ris.norms.application.service;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
 import de.bund.digitalservice.ris.norms.application.port.input.LoadElementFromNormUseCase;
+import de.bund.digitalservice.ris.norms.application.port.input.LoadElementHtmlFromNormUseCase;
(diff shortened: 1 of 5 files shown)


backend/src/test/java/de/bund/digitalservice/ris/norms/application/service/ElementServiceTest.java
```diff
@@ -116,4 +119,107 @@ void itReturnsEmptyOptionalIfNoElementIsFound() {
...
+      var norm = new Norm(XmlMapper.toDocument(normXml));
+      when(loadNormPort.loadNorm(new LoadNormPort.Command(eli))).thenReturn(Optional.of(norm));
+      when(xsltTransformationService.transformLegalDocMlToHtml(any())).thenReturn("<div></div>");
+
+      // When
+      var html =
+          service.loadElementHtmlFromNorm(new LoadElementHtmlFromNormUseCase.Query(eli, eid));
+
+      // Then
+      assertThat(html).contains("<div></div>");
```
andreasphil, line 164: @hannes-mk Not sure if this is a useful test, but I also don't have a better idea right now ...

What I'm thinking:

- We want to test that the correct norm is loaded (ensured via the query in the `loadNorm` mock)
- We want to ensure it's converted to HTML
- We don't care about the specific HTML as this is covered by the tests of the converter, integration tests of the controller, and E2E tests

So based on that it would be fine but it still feels ... incorrect 🤨

hannes-mk: I stumbled across something similar in the Controller tests.

In this test we say: do return whatever comes back from the transformation service. Do not add anything and do not remove anything.

And this does not sound interesting or test worthy.
Let's keep it in mind for the general question of "what actually to test".

PR #358, opened 2024-05-24
Update frontend save

hamo225:


frontend/src/services/normService.ts
```diff
@@ -88,7 +88,9 @@ export async function previewNorm(
     body: amendingLawXml,
   })
 }
-
+//  post json and response xml
+// good but declarative
+// use composables ?
```
malte-laukoetter: there are some comments left here

malte-laukoetter (approved):

PR #359, opened 2024-05-24
Risdev 4009 connect update mod details with time machine

malte-laukoetter:


backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/controller/NormController.java
```diff
@@ -350,7 +353,8 @@ public ResponseEntity<String> updateMod(
       @PathVariable final String language,
       @PathVariable final String subtype,
       @PathVariable final String eid,
-      @RequestBody @Valid final ModUpdateSchema modUpdateSchema) {
+      @RequestBody @Valid final ModUpdateSchema modUpdateSchema,
```
VictorDelCampo: I would then rename the request schema to `UpdateModRequestSchema`

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/mapper/UpdateModResponseMapper.java
```diff
@@ -0,0 +1,27 @@
+package de.bund.digitalservice.ris.norms.adapter.input.restapi.mapper;
+
+import de.bund.digitalservice.ris.norms.adapter.input.restapi.schema.UpdateModResponseSchema;
+import de.bund.digitalservice.ris.norms.application.port.input.UpdateModUseCase;
+import de.bund.digitalservice.ris.norms.domain.entity.Norm;
+import org.springframework.stereotype.Service;
+
+/** Mapper class for converting to a {@link UpdateModResponseSchema}. */
+@Service
```
VictorDelCampo: I would say this is a `@Service` but rather a simple class with a static method? Like the other mappers?

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/mapper/UpdateModResponseMapper.java
```diff
@@ -0,0 +1,27 @@
...
+  /**
+   * Creates a {@link UpdateModResponseSchema} instance from the {@link Norm} entities.
+   *
+   * @param result The {@link UpdateModUseCase.Result} of the update use case.
+   * @return A new {@link UpdateModResponseSchema} instance.
+   */
+  public static UpdateModResponseSchema fromResult(final UpdateModUseCase.Result result) {
+    return UpdateModResponseSchema.builder()
+        .amendingNormXml(result.amendingNormXml())
+        .targetNormXml(result.targetNormXml())
```
VictorDelCampo: Why the target norm xml is needed? Is this the ZF0? is for the time machine? Oh wait we need to pass the ZF0 xml right? Then please rename to `zf0TargetNormXml` maybe?

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/schema/PreviewRequestSchema.java
```diff
@@ -0,0 +1,17 @@
...
+@Getter
+public class PreviewRequestSchema {
+  /** The xml of the norm. */
+  private String norm;
+
+  /**
+   * The xmls of norms which are referenced by the norm (e.g. in passiveModifications) and should
+   * not be loaded from the database.
+   */
+  private Map<String, String> customNorms = Map.of();
```
VictorDelCampo: Do we need to have a Map with pairs of ELI + XML or is the XML just enough?

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/controller/RenderingController.java
```diff
@@ -15,27 +23,51 @@
...
+      consumes = {APPLICATION_JSON_VALUE},
       produces = {TEXT_HTML_VALUE})
   public ResponseEntity<String> getHtmlPreview(
-      @RequestBody final String xml, @RequestParam boolean showMetadata) {
+      @RequestBody final PreviewRequestSchema previewRequestSchema,
+      @RequestParam(defaultValue = "false") boolean showMetadata,
+      @RequestParam Optional<Instant> atIsoDate) {
+    final var norm =
+        Norm.builder().document(XmlMapper.toDocument(previewRequestSchema.getNorm())).build();
+    final Map<String, Norm> customNorms =
```
VictorDelCampo: Depending on what you decide about if we need the pairs ELI + XML this would be much easier.

backend/src/main/java/de/bund/digitalservice/ris/norms/application/port/input/UpdateModUseCase.java
```diff
@@ -40,5 +42,28 @@ record Query(
...
+  /**
+   * The results of updating an amending command. Includes both the updated amending norm and the
+   * updated norm targeted by the mod.
+   *
+   * @param amendingNormXml The xml of the amending norm, in which the akn:mod and
+   *     akn:activeModifications are updated.
+   * @param targetNormXml The xml of the norm targeted by the akn:mod, in which the
+   *     akn:passiveModifications are updated.
+   */
+  record Result(String amendingNormXml, String targetNormXml) {}
```
VictorDelCampo: Nice! Query and Result of a use case, i like it

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/schema/PreviewRequestSchema.java
malte-laukoetter: Oh true, we can get the eli from the xml so no need for a map: a Set/Array/List is totally enough, will change that!
I would like to support sending multiple custom norms to keep this endpoint general as norms can reference multiple other norms and to keep the possibility to preview a completely custom combination of norms. (But feel free to challenge that)

VictorDelCampo: I agree. The time machine will need this functionality someday when we generate the ZF1 and ZF2 because we could have the case that a ZF0 contains mods from different amending laws

PR #361, opened 2024-05-27
Refactor Article Controller: Introduce UseCase and move the heavy lifting there

hannes-mk:


backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/controller/ArticleController.java
```diff
@@ -293,47 +292,12 @@ public ResponseEntity<String> getArticleRender(
...
-        .flatMap(List::stream)
-        .filter(article -> article.getEid().isPresent() && article.getEid().get().equals(eid))
-        .findFirst()
-        .map(article -> XmlMapper.toString(article.getNode()))
-        .map(
-            xml ->
-                this.transformLegalDocMlToHtmlUseCase.transformLegalDocMlToHtml(
-                    new TransformLegalDocMlToHtmlUseCase.Query(xml, false)))
+    return loadArticleHtmlUseCase
+        .loadArticleHtml(new LoadArticleHtmlUseCase.Query(eli, eid, atIsoDate.orElse(null)))
```
hannes-mk, line 300: Should we indicate "not required" with `null`, now that we're using `null` in the use case?

hannes-mk: I have no idea what my comment meant. Maybe: not using `Optional<...>` in the endpoint's list of parameters.
If that's the case, I don't think it would be a better idea.

hannes-mk:
@malte-laukoetter What do you think of this refactoring?

andreasphil (approved):

PR #363, opened 2024-05-27
refactor: simplify controller for getting element HTML

andreasphil:
RISDEV-3899

### DIFF
--- a/backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/controller/ElementController.java
+++ b/backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/controller/ElementController.java
@@ -13,7 +13,6 @@ import de.bund.digitalservice.ris.norms.domain.entity.TextualMod;
 import de.bund.digitalservice.ris.norms.utils.EliBuilder;
 import de.bund.digitalservice.ris.norms.utils.NodeParser;
 import java.time.Instant;
-import java.time.format.DateTimeFormatter;
 import java.util.*;
 import org.jetbrains.annotations.NotNull;
 import org.springframework.core.convert.converter.Converter;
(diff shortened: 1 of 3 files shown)


andreasphil:
@hannes-mk I realized we could leave the date parsing to Spring by using `Instant` as the parameter type (rather than `String` + parse that ourselves). If we don't need to attempt the parsing ourselves we can simplify the controller a lot. Only downside is that we can't have a 400 status code then. I think that's fine but would be curious to hear your feedback.

hannes-mk:
> @hannes-mk I realized we could leave the date parsing to Spring by using `Instant` as the parameter type (rather than `String` + parse that ourselves). If we don't need to attempt the parsing ourselves we can simplify the controller a lot. Only downside is that we can't have a 400 status code then. I think that's fine but would be curious to hear your feedback.

You mean the HTML path parameter could be `Instant` and then the endpoint would accept ISO, but reject invalid ISO?
Works for me. Let's have a quick chat so I learn how it would behave.

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/controller/ElementController.java
```diff
@@ -104,29 +103,21 @@ public ResponseEntity<String> getElementHtmlPreview(
...
-          .loadElementHtmlAtDateFromNorm(
-              new LoadElementHtmlAtDateFromNormUseCase.Query(
-                  eli, eid, Instant.parse(atIsoDate.get())))
-          .map(ResponseEntity::ok)
-          .orElse(ResponseEntity.notFound().build());
-    }
-
-    return loadElementHtmlFromNormUseCase
-        .loadElementHtmlFromNorm(new LoadElementHtmlFromNormUseCase.Query(eli, eid))
+    return atIsoDate
```
hannes-mk, line 112: I'm not sure I like this construct' use of `map`. 
Let's compare the alternatives.

hannes-mk (approved):
Nothing wrong, but I'd like to learn more about the alternatives.

PR #370, opened 2024-05-30
Error states on the Amending  Law overview page RISDEV-3803

hamo225:


frontend/src/services/normService.ts
```diff
@@ -51,6 +53,38 @@ export async function getNormHtmlByEli(
...
 }
 
+/**
+ * Load the rendered html version of a norm from the api using useFetch
+ *
+ * @param eli Eli of the norm
+ * @param showMetadata Whether to include metadata in the rendered HTML
+ * @param at Date indicating which modifications should be applied before the HTML gets rendered and returned
+ */
+export function useGetNormHtmlByEli(
```
malte-laukoetter, line 63: `useNormHtml`?

malte-laukoetter: oh that is what the existing composable is already called. Maybe we should atleast merge the new composable (`useNormHtmlByEli` and this one)

frontend/src/composables/useNormHtml.ts
```diff
@@ -28,3 +36,34 @@ export function useNormHtml(
 
   return readonly(normHtml)
 }
+
+export function useNormHtmlByEli(
```
malte-laukoetter: maybe add an `@deprecated` to `useNormHtml`

frontend/src/composables/useNormHtml.ts
```diff
@@ -28,3 +36,34 @@ export function useNormHtml(
...
+  const atRef = ref(toValue(at))
+
+  watch(
+    () => [toValue(eli), toValue(at)],
+    () => {
+      eliRef.value = toValue(eli)
+      atRef.value = toValue(at)
+    },
+    { immediate: true },
+  )
```
malte-laukoetter: we should be able to remove this watcher and the two refs and be able to just use `eli` and `at` immediately.

frontend/src/composables/useNormHtml.ts
```diff
@@ -28,3 +36,34 @@ export function useNormHtml(
...
+
+  const { data, error, isFetching } = useGetNormHtmlByEli(eliRef, false, atRef)
+
+  const normHtml = computed(() => data.value ?? undefined)
```
malte-laukoetter: do we need this? Could we not just return `data` directly?

frontend/src/composables/useNormHtml.ts
```diff
@@ -28,3 +36,34 @@ export function useNormHtml(
...
+
+  return {
```
malte-laukoetter: thinking of it: is there any reason to return just these fields and not the whole thing `useGetNormHtmlByEli` returns?

And if no is there any reason to even keep this composable at all, its just calling `useGetNormHtmlByEli` at that point.

PR #371, opened 2024-05-30
Create 404 page RISDEV-3943

hamo225:


frontend/src/services/apiService.ts
```diff
@@ -1,5 +1,12 @@
 import { ofetch } from "ofetch"
 import { createFetch } from "@vueuse/core"
+import type { Router } from "vue-router"
+
+let routerInstance: Router | null = null
+
+export const initializeApiService = (router: Router) => {
```
malte-laukoetter, line 7: this feels like the api service should be a composable and use `useRouter` to get the router. But as the places using it are not all composables (yet) I think this solution works as well :)

malte-laukoetter (approved):
looks good to me :)

PR #376, opened 2024-06-03
Automatic ZF0 creation (textko)

VictorDelCampo:
RISDEV-4010

### DIFF
--- a/backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/LoadZf0Service.java
+++ b/backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/LoadZf0Service.java
@@ -1,0 +1,119 @@
+package de.bund.digitalservice.ris.norms.application.service;
+
+import de.bund.digitalservice.ris.norms.application.port.input.LoadZf0UseCase;
+import de.bund.digitalservice.ris.norms.application.port.input.UpdatePassiveModificationsUseCase;
+import de.bund.digitalservice.ris.norms.application.port.output.LoadNormByGuidPort;
+import de.bund.digitalservice.ris.norms.domain.entity.FRBRExpression;
+import de.bund.digitalservice.ris.norms.domain.entity.FRBRManifestation;
+import de.bund.digitalservice.ris.norms.domain.entity.Norm;
+import de.bund.digitalservice.ris.norms.utils.exceptions.XmlProcessingException;
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+import java.util.Optional;
+import java.util.UUID;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import javax.xml.parsers.DocumentBuilder;
+import javax.xml.parsers.DocumentBuilderFactory;
+import javax.xml.parsers.ParserConfigurationException;
+import org.springframework.stereotype.Service;
+import org.w3c.dom.Document;
+import org.w3c.dom.Node;
+
+/**
+ * Service for loading ZF0 versions of norms. If not present in DB, it will be created using the
+ * amending law and target law.
+ */
+@Service
+public class LoadZf0Service implements LoadZf0UseCase {
+
+  private final UpdateNormService updateNormService;
+  private final LoadNormByGuidPort loadNormByGuidPort;
+
+  public LoadZf0Service(
+      final UpdateNormService updateNormService, final LoadNormByGuidPort loadNormByGuidPort) {
+    this.updateNormService = updateNormService;
+    this.loadNormByGuidPort = loadNormByGuidPort;
+  }
+
+  @Override
+  public Norm loadZf0(Query query) {
+
+    final Norm amendingNorm = query.amendingLaw();
+    final Norm targetNorm = query.targetLaw();
+
+    final Optional<Norm> optionalZf0LawDB =
+        targetNorm
+            .getNextVersionGuid()
+            .flatMap(
+                guid -> loadNormByGuidPort.loadNormByGuid(new LoadNormByGuidPort.Command(guid)));
+
+    if (optionalZf0LawDB.isPresent()) {
+      return optionalZf0LawDB.get();
+    }
+
+    final Norm zf0Norm = Norm.builder().document(cloneDocument(targetNorm.getDocument())).build();
+
+    final String announcementDateAmendingLaw =
+        amendingNorm.getFBRDateVerkuendung().orElseThrow().format(DateTimeFormatter.ISO_LOCAL_DATE);
+
+    updateFRBRExpression(zf0Norm, targetNorm, announcementDateAmendingLaw);
+    updateFRBRManifestation(zf0Norm, announcementDateAmendingLaw);
+    updateNormService.updatePassiveModifications(
+        new UpdatePassiveModificationsUseCase.Query(zf0Norm, amendingNorm, targetNorm.getEli()));
+
+    return zf0Norm;
+  }
+
+  private void updateFRBRExpression(
+      final Norm zf0Norm, final Norm targetNorm, final String announcementDateAmendingLaw) {
+    final FRBRExpression frbrExpression = zf0Norm.getMeta().getFRBRExpression();
+
+    // 1.FRBRalias (vorherige-version-id / aktuelle-version-id / nachfolgende-version-id)
+    final UUID previousVersionId =
+        targetNorm.getMeta().getFRBRExpression().getFRBRaliasCurrentVersionId();
+    frbrExpression.setFRBRaliasPreviousVersionId(previousVersionId);
+    final UUID currentVersionId =
+        targetNorm.getMeta().getFRBRExpression().getFRBRaliasNextVersionId();
+    frbrExpression.setFRBRaliasCurrentVersionId(currentVersionId);
+    frbrExpression.setFRBRaliasNextVersionId(UUID.randomUUID());
+
+    // 2. new eli of zfo
+    final Matcher matcher =
+        Pattern.compile("\\d{4}-\\d{2}-\\d{2}").matcher(frbrExpression.getEli());
+    final String zf0Eli = matcher.replaceFirst(announcementDateAmendingLaw);
+    frbrExpression.setEli(zf0Eli);
+
+    // 3. FRBRDate --> announcement date of amending law + @name="aenderung"
+    frbrExpression.setFBRDate(announcementDateAmendingLaw, "aenderung");
+  }
+
+  private void updateFRBRManifestation(
+      final Norm zf0Norm, final String announcementDateAmendingLaw) {
+    final FRBRManifestation frbrManifestation = zf0Norm.getMeta().getFRBRManifestation();
+
+    // 1.replace date of eli parts
+    final Matcher matcher =
+        Pattern.compile("\\d{4}-\\d{2}-\\d{2}").matcher(frbrManifestation.getEli());
+    final String zf0Eli = matcher.replaceFirst(announcementDateAmendingLaw);
+    frbrManifestation.setEli(zf0Eli);
+
+    // 2. FRBRdate --> current system date + @name="generierung"
+    frbrManifestation.setFBRDate(LocalDate.now().toString(), "generierung");
+  }
+
+  private Document cloneDocument(Document originalDocument) {
+    try {
+      final Node originalRootNode = originalDocument.getDocumentElement();
+      final DocumentBuilderFactory dbf = DocumentBuilderFactory.newInstance();
+      final DocumentBuilder db = dbf.newDocumentBuilder();
+
+      final Document clonedDocument = db.newDocument();
+      final Node clonedRootNode = clonedDocument.importNode(originalRootNode, true);
+      clonedDocument.appendChild(clonedRootNode);
+      return clonedDocument;
+    } catch (ParserConfigurationException e) {
+      throw new XmlProcessingException(e.getMessage(), e);
+    }
+  }
+}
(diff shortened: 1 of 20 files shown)


backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/LoadZf0Service.java
```diff
@@ -0,0 +1,119 @@
...
+    frbrExpression.setFRBRaliasPreviousVersionId(previousVersionId);
+    final UUID currentVersionId =
+        targetNorm.getMeta().getFRBRExpression().getFRBRaliasNextVersionId();
+    frbrExpression.setFRBRaliasCurrentVersionId(currentVersionId);
+    frbrExpression.setFRBRaliasNextVersionId(UUID.randomUUID());
+
+    // 2. new eli of zfo
+    final Matcher matcher =
+        Pattern.compile("\\d{4}-\\d{2}-\\d{2}").matcher(frbrExpression.getEli());
+    final String zf0Eli = matcher.replaceFirst(announcementDateAmendingLaw);
```
malte-laukoetter, line 84: this could be moved to a method for calculating the zf0 eli

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/FRBR.java
```diff
@@ -0,0 +1,76 @@
...
+
+  /**
+   * Returns the Eli as {@link String} from the FRBRThis of the specific FRBR level.
+   *
+   * @return An Eli
+   */
+  public String getEli() {
+    final String xpath = "./FRBRthis/@value";
+    return NodeParser.getValueFromExpression(xpath, node)
+        .orElseThrow(() -> new MandatoryNodeNotFound(xpath, this.nodeName, this.normEli));
```
malte-laukoetter: i'm a bit unsure how nice it is to carry these information around everywhere just for the error messages. It might be possible to get them from `node.getOwnerDocument()`?

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/FRBR.java
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  /**
+   * Updates both the date and the name attibutes of FBRDate
+   *
+   * @param date - the new date
+   * @param name - the new name
+   */
+  public void setFBRDate(final String date, final String name) {
```
malte-laukoetter, line 57: if the `get` returns a date I think the `set` should also get a date (this could also remove the need to format the date in the `LoadZf0UseCase`)

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/FRBRExpression.java
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  /**
+   * Retrieves the UUID of the previous version.
+   *
+   * @return the uuid
+   */
+  public Optional<UUID> getFRBRaliasPreviousVersionId() {
+    return NodeParser.getNodeFromExpression("./FRBRalias[@name='vorherige-version-id']", getNode())
+        .map(m -> m.getAttributes().getNamedItem(VALUE_ATTIBUTE).getNodeValue())
```
malte-laukoetter:
```suggestion
    return NodeParser.getValueFromExpression("./FRBRalias[@name='vorherige-version-id']/@value", getNode())
```

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/FRBRExpression.java
```diff
@@ -0,0 +1,108 @@
...
+
+  /**
+   * Retrieves the UUID of the current version.
+   *
+   * @return the uuid
+   */
+  public UUID getFRBRaliasCurrentVersionId() {
+    final String xpath = "./FRBRalias[@name='aktuelle-version-id']";
+    return NodeParser.getNodeFromExpression(xpath, getNode())
+        .map(m -> m.getAttributes().getNamedItem(VALUE_ATTIBUTE).getNodeValue())
```
malte-laukoetter: see previous suggestion

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/FRBRExpression.java
```diff
@@ -0,0 +1,108 @@
...
+
+  /**
+   * Retrieves the UUID of the next version.
+   *
+   * @return the uuid
+   */
+  public UUID getFRBRaliasNextVersionId() {
+    final String xpath = "./FRBRalias[@name='nachfolgende-version-id']";
+    return NodeParser.getNodeFromExpression(xpath, getNode())
+        .map(m -> m.getAttributes().getNamedItem(VALUE_ATTIBUTE).getNodeValue())
```
malte-laukoetter: see previous suggestion

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/FRBRExpression.java
```diff
@@ -0,0 +1,108 @@
...
+   *
+   * @param uuid the new uuid
+   */
+  public void setFRBRaliasPreviousVersionId(final UUID uuid) {
+    NodeParser.getNodeFromExpression("./FRBRalias[@name='vorherige-version-id']", getNode())
+        .orElseGet(
+            () -> {
+              final Element newElement =
+                  getNode().getOwnerDocument().createElement("akn:FRBRalias");
+              newElement.setAttribute("eId", "meta-1_ident-1_frbrexpression-1_frbralias-1");
```
malte-laukoetter, line 44: we have `Norm::createElementWithEidAndGuid` but I don't have an idea how to best use it here (or in other classes for parts of norms). But it could simplify the code here quite a bit

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/Meta.java
```diff
@@ -0,0 +1,44 @@
...
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import lombok.experimental.SuperBuilder;
+import org.w3c.dom.Node;
+
+/** Class representing the akn:metadata */
+@Getter
+@AllArgsConstructor
+@SuperBuilder(toBuilder = true)
+public class Meta {
```
malte-laukoetter, line 13: I'm thinking about moving the getters for the sub elements of the meta node (eg. analysis, time boundaries,...) into this class, but I'm really unsure how good of an idea that would be... It would reduce the code in the Norm class but one would need to know to find them in the meta class

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/Norm.java
```diff
@@ -124,6 +131,18 @@ public Optional<Proprietary> getProprietary() {
...
+  /**
+   * Returns a {@link Meta} instance from a {@link Document} in a {@link Norm}.
+   *
+   * @return the meta node as {@link Meta}
+   */
+  public Meta getMeta() throws MandatoryNodeNotFound {
+    final String xpath = "//act/meta";
+    return NodeParser.getNodeFromExpression(xpath, document)
+        .map(node -> new Meta(node, getEli()))
+        .orElseThrow(() -> new MandatoryNodeNotFound(xpath));
```
malte-laukoetter: the more often i see this the more I'm questioning if we should move it to `NodeParser` as `getMandatoryNodeFromExpression`

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/LoadZf0Service.java
malte-laukoetter: or keep it for now, I spend some time on wednsday experimenting with an ELI class (mostly for removing the need for that many `@PathParameter` things in the controllers) and this seems like something that could be moved there (as a `setPointInTime`)

malte-laukoetter (commented):
Love the new FRBR classes <3

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/LoadZf0Service.java
VictorDelCampo: I will leave it there for now, since you seem to have some thoughts in the repetitive eli params. Thanks

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/FRBR.java
VictorDelCampo: This was something I thought about a lot... and I was about to move the getEli to the FRBRExpression, since it would logically belong there but the ELI is something we use everywhere and it is nice to just do norm.getEli()

About passing it around this was my best approach, since the exception would me meaningless if we don't know on which law (eli) the error happened.

But to get it from the owner document would mean you need to pass the Document to a Norm again, so that you can call the getEli() method... don't really think this would be a good approach.

Another idea would be to set the eli on a process (API call, service, bla bla) level so that you can always access it, doesn't matter where. But this would require much more work specially because we don't always use the ELI from the API endpoint but rather from a norm we loaded from the db

If we don't have a better approach now, I would say let's pass the eli around for now. What do you say?

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/FRBR.java
VictorDelCampo: I would prefer to return a String in the getFRBRDate than to pass a LocalDate since we don't have any requirements now for using a LocalDate like compare them or something like that. For now we just get a date from the XML, in the ISO string format and set it in another FRBRDate. What do you think?

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/FRBRExpression.java
VictorDelCampo: We could for sure create later on a factory class to create different kind of nodes by passing the necessary info like parent node, tag name and the map list of attributes+values. But for now is in my opinion overengineered

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/Meta.java
VictorDelCampo: This would be for me next step after merging into main. And since our domain is lean to the LDML.de standard it would be a nice practice for everybody to find out from which entity class to take what information. Don't you think?

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/Norm.java
VictorDelCampo: I like it. I will do it. We then need to pass the xpath, the document or node and for sure the eli so that we can't then throw the custom exception in the NodeParser, you agree? If so I would implement it

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/FRBRExpression.java
malte-laukoetter: yes, sure, for now it works and we can keep it like it is :)

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/Meta.java
malte-laukoetter: I would agree, as long as we keep something like `getEli` in the norms class :)

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/Norm.java
malte-laukoetter: sounds good :)

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/FRBR.java
malte-laukoetter: Yeah the information is really helpful to have here! 

I also like your idea of keeping it as information on the process but that really seems like a lot of work and thought that we would need to invest into that.

But I would actually prefer to calculate the eli from the document again instead of passing it around (we need to keep infinite loops in mind here), as it would keep the api simpler and we can encapsulate all this logic to one place (otherwise we would need to add these fields to all of these classes where we want to have mandatory nodes). Maybe it would actually be a better idea to use the xpath for getting the eli directly on the document, even if that duplicates some code, instead of converting the document to a norm.

Also do we need the `nodeName` or could that just be `node.getNodeName()`?

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/FRBR.java
malte-laukoetter: Works for me as well, as long as getter and setter work with the same type :)

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/Norm.java
VictorDelCampo: Done

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/FRBR.java
VictorDelCampo: With the nodeName jep we don't need it. Done. And with the getEli I put it close to the exception which is now only thrown in the NodeParser, with a new method for getting mandatory nodes (and values)

PR #377, opened 2024-06-03
Input validation risdev 3801

hamo225:


frontend/src/components/controls/RisDateInput.vue
```diff
@@ -117,17 +156,48 @@ watch(inputCompleted, (is) => {
...
-    placeholder="TT.MM.JJJJ"
-    :readonly="isReadOnly"
-    @blur="onBlur"
-    @keydown.delete="backspaceDelete"
-    @maska="onMaska"
-  />
+  <div
+    :class="
+      props.labelPosition === 'left'
+        ? `|| 'gap-8'} flex items-center gap-4`
```
malte-laukoetter: it feels like there is missing some logic here (in regards to the `|| 'gap-8'}`)

frontend/src/components/controls/RisDateInput.vue
```diff
@@ -117,17 +156,48 @@ watch(inputCompleted, (is) => {
...
+        : 'grid gap-2'
+    "
+  >
+    <!-- eslint-disable-next-line vuejs-accessibility/label-has-for -->
```
malte-laukoetter: can we nest the input within this label? That should solve this warning.

frontend/src/components/controls/RisDateInput.vue
```diff
@@ -4,23 +4,40 @@ import dayjs from "dayjs"
...
+    /** Label of the form field. */
+    label?: string
+    /** Validation error and message to display. */
+    validationError?: ValidationError
+    /** Label position: 'above' or 'left' */
+    labelPosition?: "above" | "left"
+    /** Custom classes for the container */
+    containerClass?: string
+    /** Custom classes for the label */
+    labelClass?: string
```
malte-laukoetter: do we actually need this or is this just the same class list for every `RisDateInput` anyway?

malte-laukoetter (changes requested):

frontend/src/components/controls/RisDateInput.vue
hamo225: Not needed anymore. Removing.

frontend/src/components/controls/RisDateInput.vue
hamo225: Good point

frontend/src/components/controls/RisDateInput.vue
hamo225: I have removed both.

PR #392, opened 2024-06-07
Expose documenttype properties in controller (RISDEV-4098)

hannes-mk:
Btw. not changing anything in the non- atDate controller as it's about to be removed (https://digitalservicebund.atlassian.net/browse/RISDEV-4109)


backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/schema/ProprietaryResponseSchema.java
```diff
@@ -18,4 +21,28 @@ public class ProprietaryResponseSchema {
   public static class Fna {
     private String value;
   }
+
```
hannes-mk, line 24: @VictorDelCampo FYI: I have not changed the general setup of the response schema for my task. This stays an open topic.

At least we now have some more metadata in here to lead whatever abstraction we may use in the future.

PR #394, opened 2024-06-07
Move updating activeMods to its own use-case

reckseba:
RISDEV-3824


reckseba:
@malte-laukoetter I came to the conclusion to only refactor updateActiveModifications to its own use-case as updating the mod part of the amendingLaw would result in updateModUseCase which already exists.

malte-laukoetter (approved):

PR #395, opened 2024-06-07
E2E test validator

SebastianRossa:


frontend/e2e/validation.spec.ts
```diff
@@ -0,0 +1,60 @@
+import { test, expect } from "@playwright/test"
+
+test.describe("validation errors", () => {
```
andreasphil: Todo: this tests validation on a specific page, right? So it should be in that page's test file or at least be named more precisely

frontend/e2e/validation.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+  test("Wrong character range", async ({ page }) => {
+    await page.goto(
+      "http://localhost:5173/amending-laws/eli/bund/bgbl-1/2017/s419/2017-03-15/1/deu/regelungstext-1/articles/hauptteil-1_art-1/edit/hauptteil-1_art-1_abs-1_untergl-1_listenelem-2_inhalt-1_text-1_%C3%A4ndbefehl-1",
```
andreasphil: Todo: you can use a relative URL so we don't hardcode host and port. Same for the other URLs

frontend/e2e/validation.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+    )
+    await page
+      .getByLabel("zu ersetzende Textstelle")
+      .fill(
+        "hauptteil-1_para-20_abs-1_untergl-1_listenelem-2_inhalt-1_text-1/9-36.xml",
+      )
+    page.once("dialog", (dialog) => {
```
andreasphil: Question: I'm not sure what this is doing, but I assume this is going to be removed anyways once the result is reported with something other than alert boxes, right?

andreasphil (changes requested):

andreasphil (approved):

malte-laukoetter:
merged in https://github.com/digitalservicebund/ris-norms/commit/6bca9f847137d1131c350a17b419b78a66121084

PR #400, opened 2024-06-10
RISDEV-3807 implement states on change command editor

malte-laukoetter:


frontend/src/components/RisModForm.vue
```diff
@@ -161,11 +165,27 @@ function modTypeLabel(modType: ModType | "") {
...
+      <div class="relative">
+        <RisTooltip
+          v-slot="{ ariaDescribedby }"
+          :visible="isUpdatingFinished"
+          :title="
+            updateError ? 'Fehler beim Speichern' : 'Speichern erfolgreich'
+          "
+          alignment="right"
+          attachment="top"
+          :variant="updateError ? 'error' : 'success'"
```
andreasphil, line 178: Question: Is the "show success message" a general pattern we're using? Even though it's possible with the tooltip I haven't heard about this from design. But might've missed it :)

frontend/src/components/RisModForm.vue
```diff
@@ -161,11 +165,27 @@ function modTypeLabel(modType: ModType | "") {
...
+        >
+          <RisTextButton
+            :aria-describedby="ariaDescribedby"
```
andreasphil, line 181: Nitpick: in case you didn't know, Vue allows you to use a [shorthand](https://vuejs.org/guide/essentials/template-syntax.html#same-name-shorthand) for this now. Not a required change though if you prefer the more explicit version

frontend/src/services/apiService.ts
```diff
@@ -69,6 +69,18 @@ export const useApiFetch = createFetch({
         })
       }
 
+      // this error is sometimes throws when previous requests are automatically aborted as
+      // some of the data changed and refetch is true. It seems to only be throws when the request
+      // is aborted before it was actually send.
+      // We ignore this error as it (for some odd reason) isn't replaced once the second request finishes
+      // successfully
+      if (fetchContext.error.name === "AbortError") {
```
andreasphil, line 77: Question: I created unit tests for `useApiFetch`, is this a behavior we can/should test there?

frontend/src/components/RisModForm.vue
malte-laukoetter: We have a design for a success message in figma. But now that you mention it it's not actually using the tooltip and I'm unsure if it is applicable here 🤔

frontend/src/components/RisModForm.vue
malte-laukoetter: i didn't know (or forgot) about it thanks for mentioning it!

frontend/src/services/apiService.ts
malte-laukoetter: hm... I'm not sure if i'm able to mock the specific behaviour that is causing this, but i could give it a try, and at least i can add a test for `AbortError`s in general.

frontend/src/components/RisModForm.vue
malte-laukoetter: I would keep it for now anyway as it is (in my opinion) still better to have some feedback that it worked than none at all

andreasphil: Fine with me, but could you check with design about this and ask if this is something we should use everywhere? Thanks!

PR #403, opened 2024-06-12
Loading & error states on metadata editor

andreasphil:
RISDEV-3808


frontend/src/services/normService.ts
```diff
@@ -145,3 +145,27 @@ export function usePutNormXml(
...
+ *
+ * @param eli ELI of the norm
+ * @param options Optional additional filters and queries. This is only a
+ *  placeholder for now and can not have a value.
+ * @param [fetchOptions={}] Optional configuration for fetch behavior
+ * @returns Reactive fetch wrapper
+ */
+export function useNormService(
+  eli: MaybeRefOrGetter<string>,
+  options?: never,
```
malte-laukoetter, line 160: remove it if we don't need it for now?

frontend/src/services/proprietaryService.ts
```diff
@@ -41,8 +38,43 @@ export function useProprietaryService(
...
+ * @param eli ELI of the norm
+ * @param options Optional additional filters and queries
+ * @param [fetchOptions={}] Optional configuration for fetch behavior
+ * @returns Reactive fetch wrapper
+ */
+export function usePutProprietary(
+  updateData: MaybeRefOrGetter<Proprietary | null>,
+  eli: Parameters<typeof useProprietaryService>["0"],
+  options: Parameters<typeof useProprietaryService>["1"],
+  fetchOptions: Parameters<typeof useProprietaryService>["2"],
```
malte-laukoetter, line 75: for a put the current options do not have an effect and the stuff that needs to be passed to `fetchOptions` should be handled here (atleast in my understanding). So do we really need/want those parameters here? I would try to keep the parameters for these methods as simple as possible and only extend them if we actually need to support more stuff

frontend/src/views/AmendingLawAffectedDocumentEditor.vue
```diff
@@ -1,25 +1,35 @@
...
 const amendingLawEli = useEliPathParameter()
 const affectedDocumentEli = useEliPathParameter("affectedDocument")
-const amendingLaw = useAmendingLaw(amendingLawEli)
+
+const {
+  data: amendingLaw,
+  isFetching: amendingLawIsLoading,
+  error: amendingLawError,
+} = useNormService(amendingLawEli, undefined, {
+  immediate: true,
```
malte-laukoetter, line 21: `immediate` is `true` by default

frontend/src/services/proprietaryService.ts
```diff
@@ -41,8 +38,43 @@ export function useProprietaryService(
...
+): ReturnType<typeof useProprietaryService> {
+  return useProprietaryService(eli, options, fetchOptions)
```
malte-laukoetter, line 77: i would expect this to set `{ refetch: false }` so there is a lower risk of forgetting to do this. But we can discuss that.

malte-laukoetter: this would also mean that the fetchOptions do not need to be set by the views as they are already "correctly" defined

frontend/src/views/AmendingLawAffectedDocumentEditor.vue
andreasphil: Good catch, I'll change it

frontend/src/services/proprietaryService.ts
andreasphil: Agree, I'll change it

frontend/src/services/proprietaryService.ts
andreasphil: I would prefer if we could stick to the `useFetch` defaults. The reason being that I would like all services to behave the same, i.e. we should avoid having some services refetch automatically whereas others don't. Sticking with the `useFetch` default behavior seems to be the easiest way to achieve that, even if it is a bit more verbose. Would that be fine with you?

(At least for the services that is, I don't mind if composables set the refetch)

malte-laukoetter: I would have expected all gets to have `refetch` set to `true` and all other endpoints to have `immediate` set to `false` (the only exception to this so far is the render endpoint as it is a post endpoint but acts more like a get). But I'm also fine with having this in the composable layer.

frontend/src/services/proprietaryService.ts
andreasphil: I checked again and the options do have an effect and this is used (see `AmendingLawAffectedDocumentRahmenEditor.vue` for an example). So I'm keeping it as it is

PR #404, opened 2024-06-12
Refactor get/set simple metadata with an enum class

VictorDelCampo:
After having introduced 4 metadata already I have realized that they all behave in the same way within the MetadatenDs

- Getting a value to a given date.
- Setting a value to a given date.

I have therefore introduced an enum class within the MetadatenDs entity. With this implementation, whenever we have a new simple metadata the only places we need to extend are:

- Are new entry in the enum class with the xpath
- For GET, add new method `getMetadatumAt()` in `Proprietary` (since it could be that it is present in `MetadatenDe` and not only `MetadatenDs`, this part can be further optimized), and extend the `ProprietaryResponseMapper` 
- For PUT just add a new line in ProprietaryService like     `metadatenDs.setSimpleProprietaryMetadata(MetadatenDs.SimpleMetadatum.NEW-METADATUM, query.atDate(), query.metadata().fna());`


andreasphil (approved):
LGTM, thanks @VictorDelCampo :)

PR #405, opened 2024-06-13
Loading & error states on metadata editor (continued)

andreasphil:
RISDEV-3808

Follow up to #403

### DIFF
--- a/frontend/src/services/normService.ts
+++ b/frontend/src/services/normService.ts
@@ -9,6 +9,7 @@ import { computed, MaybeRefOrGetter, toValue, unref } from "vue"
  *
  * @param eli Eli of the amending law
  * @param options Fetch options for the request
+ * @deprecated
  */
 export async function getNormByEli(
   eli: string,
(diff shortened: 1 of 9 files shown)


frontend/src/services/normService.ts
```diff
@@ -150,22 +153,84 @@ export function usePutNormXml(
...
+ */
+export function useGetNormHtml(
+  eli: Parameters<typeof useNormService>["0"],
+  options?: Parameters<typeof useNormService>["1"],
+  fetchOptions?: Parameters<typeof useNormService>["2"],
+): UseFetchReturn<string> {
+  return useNormService(eli, options, {
+    ...fetchOptions,
+    beforeFetch(c) {
+      c.options.headers = { ...c.options.headers, Accept: "text/html" }
```
malte-laukoetter, line 233: as we have the option to pass the options we can also directly set the header in there. But I'm a bit unsure if thats a better option.

andreasphil: Not sure what you mean, can you give me an example?

malte-laukoetter:
```ts
return useNormService(eli, {
  ...options,
  headers: {
    ...options.headers,
    Accept: "text/html"
  }
}, fetchOptions)
```

PR #416, opened 2024-06-17
Refactor validators

VictorDelCampo:
I found the validator so hard to read that I have a refactor proposal for it.

**Context information**

What can be modify in the UI:
- The destination href containing
  - The target eId
  - The character range
-  The new text that will replace the old one

Assumptions
- Active mod is valid and it contains an eId
- ZF0 is valid itself and
  - It does not contain duplicate eIds
  - the passive mods:
     - have a source href and the source href have an eId
     - contain the one passive mod inside than has been translated from the active mod being modified. And this one has a destination href (what is being set in the UI) and the destination href has an eId

What is to validate:
- destination href has an existing target eId in ZF0
- destination href has a character range with the following conditions
   - the format is valid, meaning two numbers separated with comma and the first number is lower than the second one.
   - the range is within the range of the length of the text content from the target node
   - the range resolves to the old text being targeted

**Refactor changes in this PR**
VALIDATORS
- I have separated the "big" validator we had into two:
  - `SingleModValidator` --> with only the validation currently needed in our workflow of update mod
  - `AmendingLawValidator` --> containing the validator of a whole amending law that validate all mods (which is reusing the ModificationValidator)
- I have strongly simplified the `SingleModValidator` following the assumptions above and also simplified the error messages, since we are in the specific use case of updating an active mode. Meaning the context is already in the frontend... meaning no need for printing eId or eli

EXCEPTIONS
- I have created a new `ValidationException`, that is to be used when doing validation processes. Also extended the ControllerAdvice to handle these ones. The `XmlContentException` is still there but in my opinion to be used when working with an XML in general.

VALIDATION SERVICE AND VALIDATOR INTERFACE
- There is now a `Validator` interface to be implemented by any custom validator we write. The validators themselves are no longer services but components and are "managed" by the `ValidationService`, that will be used everywhere we want to use a custom validator.

**Possible extensions**
- Maybe in the future we could think about chaining validators.

**Open discussion**
- Problematic could be the creation of a new package `validator` under the application layer. Had to extend the architecture fitness test for that. Not really sure if this is ok, since the application layer was only divided between the `port` and `service` packages.


reckseba (approved):

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/NormService.java
```diff
@@ -32,7 +33,7 @@ public class NormService
   private final LoadNormPort loadNormPort;
   private final LoadNormByGuidPort loadNormByGuidPort;
   private final UpdateNormPort updateNormPort;
-  private final ModificationValidator modificationValidator;
+  private final ValidationService validationService;
```
malte-laukoetter: Could we not "just" use a `SingleModValidator` here and remove the need for both `ValidatorService` and `ValidatorName`? I don't see the advantage that the `ValidatorService` provides us with at the moment 🧐

backend/src/main/java/de/bund/digitalservice/ris/norms/application/validator/Validator.java
```diff
@@ -0,0 +1,15 @@
...
+/** Interface to be implemented by any custom validator. */
+public interface Validator {
+
+  /**
+   * Validates the passed data within the custom implementation.
+   *
+   * @param args - variable-sized input parameters.
+   * @throws ValidationException whenever a validation step fails
+   */
+  void validate(Object... args) throws ValidationException;
```
malte-laukoetter: Can we use a generic instead of `Object...` here? That would allow us to use the type checking to check if our validator is called correctly and we do not need to rely on runtime type checking to ensure that the validator is called correctly. (And additionally it would be simpler to understand what each parameter is supposed to be)

backend/src/main/java/de/bund/digitalservice/ris/norms/application/validator/SingleModValidator.java
```diff
@@ -0,0 +1,80 @@
...
+ * A custom validator for validating single mods that are being updated on the amending law. This
+ * validator uses the ZF0 version of the target law targeted by the mod
+ */
+@Component
+public class SingleModValidator implements Validator {
+
+  public static final ValidatorName NAME = ValidatorName.SINGLE_MOD;
+
+  @Override
+  public void validate(final Object... args) throws ValidationException {
```
malte-laukoetter: Could you please add a doc-comment explaining what `args[0]` and `args[1]` is supposed to contain.

backend/src/main/java/de/bund/digitalservice/ris/norms/application/validator/Validator.java
```diff
@@ -0,0 +1,15 @@
+package de.bund.digitalservice.ris.norms.application.validator;
+
+import de.bund.digitalservice.ris.norms.utils.exceptions.ValidationException;
+
+/** Interface to be implemented by any custom validator. */
+public interface Validator {
```
malte-laukoetter: We could also use the existing interface for validators from [spring](https://docs.spring.io/spring-framework/docs/6.1.9/javadoc-api/org/springframework/validation/Validator.html) or [jakarta](https://jakarta.ee/specifications/platform/9/apidocs/jakarta/validation/validator). That would "for free" provide us with some convenient ways of calling the validator like using the `@Valid` annotation and the option to use more complex validation errors.

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/NormService.java
VictorDelCampo: We could do that yes, but I think that encapsulating the logic of which validator to use can be deferred to the `ValidationService`. Later on we could also have the case that we need to chain several validators and this could be done by the `ValidationService`. Also we keep our backend package structure cleaner by having only services within the `service` package.

backend/src/main/java/de/bund/digitalservice/ris/norms/application/validator/Validator.java
VictorDelCampo: That was my first implementation but we still need to have a variable size of input parameters. Now we have the `AmendingLawValidator` only passing a `Norm` object and the `SingleModValidator` passing both a `Norm` object as well as a `Mod` object. We would need to use, for now, two generics for the `SingleModValidator` but the `AmendingLawValidator` only needs one

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/NormService.java
VictorDelCampo: We could yes but I think it is better if we deferred the logic of which validator to use to a service. Maybe later we want to chain validators and this could be done in the `ValidationService`.

backend/src/main/java/de/bund/digitalservice/ris/norms/application/validator/Validator.java
VictorDelCampo: Oh i didn't know about that. But then we could need to create custom classes for the input parameters passed to the validator right? Like the `SingleModValidator` would need a class containing a Norm and also a Mod. I think this validator interface is a good fit for domain entities that you already have. Or you think it is worth the time investigating a bit further?

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/NormService.java
malte-laukoetter: How would you add chaining logic here? I'm really unsure how that would simplify any any work by using the `ValidationService` instead of calling two (or more) Validators. But having the `ValidationService` definitely adds some complexity.

I'm also not sure I understand how this change affects the package structure

backend/src/main/java/de/bund/digitalservice/ris/norms/application/validator/Validator.java
malte-laukoetter: Could we use something like an `Arguments` sub-class within the validator to bundle `Norm` and `Mod` for this case? (So similar to the `Query´ or `Command` things) I would really like to have this type safety here as it seems very simple to use the validators wrongly otherwise.

backend/src/main/java/de/bund/digitalservice/ris/norms/application/validator/Validator.java
malte-laukoetter: I agree that at the moment this `Validator` interface would work totally fine.
But if we ever feel the need to extend it further / support more things I think we should investigate using an existing solution. I would expect that there are a lot of things we haven't even thought of that they have already solved for us as validation is a pretty common problem. Eg. how to validate multiple things at once and return messages for all of them and not just the first problem.

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/NormService.java
VictorDelCampo: I had an idea in mind with the chaining but I just realized it was using the `Validator` interface... so forget that wrong argumentation.

I created the `ValidationService` after having refactored the validators (meaning separating the two) and, at least in my head, they are not real services but rather smaller components doing just 1 thing: validating. So that is why I created the `ValidationService` to act as the "manager" of the validators.

Every future custom validator will implement the `Validator` interface but will not act as a service, or at least as the services we had till now (meaning call coming from the controller and using a use case which is implemented in a service). But well you could also say the `ValidationService` is not implementing a use case but we could introduce one or two. Also calling a validator a service feels wrong but it may just be my feeling.

But yeah this argumentation would not be a reason enough, now that I try to argument the idea, to having the `ValidationService`.  But maybe in the future we will extend what the ValidationService can do, like for example seting flags for throwing exceptions or do a dry-run. 

But please if you don't see any benefit or future application, keep challenging it and I would remove it.

backend/src/main/java/de/bund/digitalservice/ris/norms/application/validator/Validator.java
VictorDelCampo: Sure we could but then the `ValidationService`... if needed, would need to be re-worked because calling the same validate method is now working because of the varargs of the input parameters.

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/NormService.java
VictorDelCampo: Oh forgot, the thing with the package structure is we had before only `port` and `service` and I introduced now the `validator` package, where we could put all validators annotated with @Component (I know is just the generic annotation)

backend/src/main/java/de/bund/digitalservice/ris/norms/application/validator/Validator.java
VictorDelCampo: I understand. Let's keep this in mind for when we extend our validation process

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/NormService.java
malte-laukoetter: I don't really see why a validator shouldn't be a service, but thats not a hill i will die on ^^ So I'm totally fine with having them separated and using the `@component` annotation. (I actually see an advantage with having them in another folder to keep the service folder smaller)

I'm still not convinced that the benefits you see in the `ValidatorService` outweigh the problems it could create (also for type-safety as you mentioned in the other comment) but as we just discussed we can move all this to a later time :)

VictorDelCampo:
After several discussion rounds I have re-refactor the validator and together with @malte-laukoetter 
 decided to remove both `Validator` interface and `ValidationService` and rather inject the custom validators directly where needed (at least for now)

I have also extended the log messages we produce when a validation exception is thrown so that when debugging we can always identify with norm is affected (eli) and which eId was being handled (like the target node eId and the eId of the passive mod, for when validating the character range)

malte-laukoetter (approved):

PR #423, opened 2024-06-19
Metadata success state

andreasphil:


frontend/src/composables/useNormXml.ts
```diff
@@ -37,12 +37,14 @@ export function useNormXml(
...
-  const getNormXml: UseFetchReturn<string> = useGetNormXml(eli)
+  const getNormXml: UseFetchReturn<string> = useGetNormXml(eli, undefined, {
+    refetch: true,
+  })
   watch(getNormXml.data, () => {
     data.value = getNormXml.data.value
   })
 
-  const putNormXml = usePutNormXml(newXml, eli)
+  const putNormXml = usePutNormXml(newXml, eli, undefined, { refetch: true })
```
malte-laukoetter, line 45: i'm not sure why refetch is set to true here, but i haven't really thought about it

PR #424, opened 2024-06-19
Optimize E2E tests for metadata

andreasphil:
RISDEV-4221

### DIFF
--- a/frontend/e2e/amending-law-metadata-editor-rahmen.spec.ts
+++ b/frontend/e2e/amending-law-metadata-editor-rahmen.spec.ts
@@ -356,6 +356,131 @@ test.describe("metadata view", () => {
   })
 })
 
+test.describe("displaying, editing, and saving", () => {
+  let sharedPage: Page
+
+  async function restoreInitialState() {
+    const data: Proprietary = {
+      fna: "210-5",
+      art: "regelungstext",
+      typ: "gesetz",
+      subtyp: "Rechtsverordnung",
+      bezeichnungInVorlage: "Testbezeichnung nach meiner Vorlage",
+      artDerNorm: "SN,ÜN",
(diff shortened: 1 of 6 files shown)


frontend/e2e/amending-law-metadata-editor-rahmen.spec.ts
```diff
@@ -0,0 +1,842 @@
...
+      await sharedPage.unrouteAll()
+    })
+  })
+
+  test.describe("Dokumenttyp", () => {
+    let control: Locator
+
+    test.beforeAll(() => {
+      control = sharedPage.getByRole("combobox", { name: "Dokumenttyp" })
+    })
```
malte-laukoetter, line 438: I'm a bit unsure about moving this up here, especially as it is only one line that is needed for the setup and one now needs to scroll around to understand what kind of element `control` is. I would have just duplicated it in all test cases.

VictorDelCampo: It is not always just one line ;) (see "art der norm") I proposed @andreasphil to move the control app to a `beforeAll` so that we don't need to duplicate it every time (always when using the string for the name you have the danger to misspell it somewhere).

But if it is about which kind of element is that, we could name the variables better for every metadatum, like for example `checkboxSN`, or `dropdownDocumentType`

I do understand the meaning of duplicating stuff for better readability of a test, but for these ones where we have a describe per metadatum and if we better-name the variables, I think the readability would still be given

andreasphil: Thanks for the feedback, I think both are valid so let's leave that to whoever implements the test :) I improved the naming though to be more descriptive, the `control` was mostly for copy-pastability but you're right that it's not very helpful.

PR #425, opened 2024-06-19
Federführung backend + E2E tests

VictorDelCampo:
While adding the get/set for "Federführung" I realized that we need the functionalities that were till now implemented only for `MetadatenDs` also accessible in `MetadatenDe`.

Therefore I decided to create an abstract class `Metadaten` that is extended by the other two, having this way the common functionalities at one place (getting/setting values of metadata) and with a generic so that we can pass the concrete enum class with the different metadata.

Also added two variables to the abstract class `startAttribute` and `endAttribute` which are initialized by the construtor/builder of the child classes so use the correct attributes naming depending on the Metadaten-Block (Ds or De).

In addition I have extended the method for setting a metadatum so that we can also automatically cope with nested nodes like `./federfuehrung/federfuehrend`, so that the parent node is first retrieved (or created) and looping to the depth till we find the last child node that should get the text content.

The constructors in MetadatenDs and MetadatenDe with the @Builder annotation (after removing the @SuperBuilder at the class level) is needed in order to not having to specify every time in the tests `startAttribute` and `endAttribute`

### DIFF
--- a/backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/Metadaten.java
+++ b/backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/Metadaten.java
@@ -1,0 +1,170 @@
+package de.bund.digitalservice.ris.norms.domain.entity;
+
+import de.bund.digitalservice.ris.norms.utils.NodeCreator;
+import de.bund.digitalservice.ris.norms.utils.NodeParser;
+import java.time.LocalDate;
+import java.util.Comparator;
+import java.util.List;
+import java.util.Optional;
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import org.w3c.dom.Element;
+import org.w3c.dom.Node;
+
+interface MetadataInterface {
+  String getXpath();
+}
+
+/**
+ * Abstract class with common functionalities between meta:legalDocML.de_metadaten_ds and
+ * meta:legalDocML.de_metadaten
+ *
+ * @param <T> the concrete enum class implementing {@link MetadataInterface}
+ */
+@Getter
+@AllArgsConstructor
+public abstract class Metadaten<T extends MetadataInterface> {
+
+  private final Node node;
+  private final String startAttribute;
+  private final String endAttribute;
+
+  /**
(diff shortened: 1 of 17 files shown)


backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/Metadaten.java
```diff
@@ -0,0 +1,170 @@
...
+import java.time.LocalDate;
+import java.util.Comparator;
+import java.util.List;
+import java.util.Optional;
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import org.w3c.dom.Element;
+import org.w3c.dom.Node;
+
+interface MetadataInterface {
```
reckseba, line 14: We now have a "Metadata" interface next to a "Metadaten" abstract. Language wise I was quite confused and thought that this must have slipped through. Letting that sink in, I now understand why. Actually no recommendation here.

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/Metadaten.java
```diff
@@ -0,0 +1,170 @@
...
+            .toList();
+    if (!valuesWithStartOrAndEnd.isEmpty()) {
+      return valuesWithStartOrAndEnd.stream().findFirst();
+    } else {
+      return valuesWithoutStartAndEnd.stream().findFirst();
+    }
+  }
+
+  private void createNewNode(
+      final T metadatum,
```
reckseba: Wanna rename this one to `simpleMetadatum` like the others?

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/Metadaten.java
VictorDelCampo: I could have put the interface in another file but it is just needed for the abstract class to be able to access the xpath of both types of enums within its scope. That is why it belongs somehow to the abstract class, which is specifying in the generic that the enum/class passed must implement that interface

PR #431, opened 2024-06-24
First iteration exception handling in backend

VictorDelCampo:
This PR represents a first iteration proposal for the exception handling in the backend. To be used as an idea for discussions.

**Goal**: Have a common handling of exceptions in our backend that can be integrated into our hexagonal architecture.

**Strategy**: 
1. Place exceptions in the architecture layers they belong to. If an exception is used (thrown or handled) by several layers it will be placed into the newly created `common/exception` package.
2. Propagate exceptions throughout the application till they are handled by  our `@ControllerAdvice`s. This annotation is used in Spring to handle exceptions across all the controllers. This is in my opinion the best way in a backend application that consists solely of an API supporting the frontend.
3. Try/catch exceptions within the business logic if they should not be thrown but handled in order to "activate" another flow (like in the `LoadZf0Service` when checking if we do have already a ZF0)

**Main changes**:
- The following layers:
    - `application` layer contains now `ValidationException` and `TransformationException`.
    - 'utils' package contains now only `MandatoryNodeNotFoundException` and `XmlProcessingException`.
    - `common` package contains now `NormNotFoundException`
- I removed the `XmlContentException`, which was being used in the `AmendingLawValidator` and replace it there with the `ValidationException`. It was also being used in the `XsltTransformationService` for which I created a new `TransformationException`
- I have adapted three output ports (`LoadNormPort`, `LoadNormByGuidPort` and `UpdateNormPort`) so that they don't return an `Optional<Norm>` anymore but rather always a `Norm` and throw the `NormNotFoundException` if it was not found.
- Since that exception is being propagated i have also updated the use cases (input ports) that are using those output ports.

`Open questions`:
1. Should we maybe just put ALL exceptions into the `common/exception` package since they are all "handled" in the `@ControllerAdvice` which is located in the `adapter` package?

If there is much discussion going on in this PR (or Slack) I would schedule a discussion meeting.

⚠️ I would write an **ADR** once we have gotten to a common strategy/concept for the exception handling in the backend. ⚠️


backend/src/test/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/controller/NormControllerTest.java
```diff
@@ -48,7 +49,7 @@ void itReturnsNorm() throws Exception {
       var norm =
           Norm.builder()
               .document(
-                  XmlMapper.toDocument(
+                  XmlProcessor.toDocument(
```
malte-laukoetter, line 52: Nitpicky thing: could you put a rename like XmlMapper -> XmlProcessor into a separate commit next time, that would make reading the diff a bit less noisy ^^

malte-laukoetter (approved):
Sounds like a good plan 👍 

I'm not really sure if we do need this `common` layer at all or could just add the exceptions to the individual layers by e.g.
* allow the adapter (or only the CustomExceptionHandler) to access the exceptions irregardless of the layer they reside in
* or rethrow the exceptions (as application-layer-exceptions) in the application layer (my preferred way; I don't think this adds too much boilerplate and I think according to the architecture idea they are technically different exceptions and should be mapped between, similarly to data)

This way we can keep the separations between the packages and avoid adding a `common`-package, which could otherwise lead us to place more and more stuff there as it exists already

And if we keep the common package could you still extend the arch-tests to ensure that it is only used for exceptions?

SebastianRossa:
In my opinion we could use Optionals for our output ports. In the application layer we can than decide to throw an exception. My view here would be that to throw an exception is a business logic decision. (disruption of further code execution and immediately return an api response vs. collect all places of malformed xml and return them all at once)
- testing values is faster than throwing exceptions
- try catch blocks inhibit JVM optimisation

If we wanna stay with exception throwing in the lower levels (e.g. database) we should use exception translation to fit the abstraction of our application layer. See Effective Java Item 73
- method/rest api might break if lower level exceptions change
- e.g. we don't fetch laws from our database but from the portal api

Use standard exceptions instead of custom ones? (if no individual return object is needed) See Effective Java Item 72
- api easier to learn because programmers are familiar with them
- easier to read
- smaller memory footprint because less classes need to be loaded
- e.g. IllegalArgumentException when a href is malformed
This might not apply here since we often want to return an individual api response pointing to malformed xml

In general we should use exceptions as little as possible and prefer state-checking methods and Optionals to be sure to catch the real reason of the Exception.

reckseba:
I'm okay with the proposal and I think that I could find myself productive working in that environment. Though:

- Not sure. Maybe I'd need to code an example.
- 114 files touched is quite a number and hard to grasp. I must admit I haven't read it in full and I'm not sure if I captured the essence of it all, besides all the re- naming, structuring, factoring. (One commit that cuts to the chase might have helped me - But i get that this is hard when the coding flow hits you)
- One super low level finding: sometimes packages are `exceptions` (plural) and sometimes singular. On purpose?
- @malte-laukoetter concerns about a place where things stack up, seem reasonable
- @SebastianRossa concerns about inhibiting JVM optimization by over utilizing try/catch, too (Reminds me of finally reading Effective Java) Though I'm quite surprised that he has such strong feelings about it. It's nice to shed light on the topic from this perspective.

PR #436, opened 2024-06-25
RISDEV-4240 prototype reference documentation

malte-laukoetter:
**DO NOT MERGE - This is just a prototype and has neither tests nor documentation and is just there to demonstrate how it could work. We should not take this as is even when we want this functionality as this was quickly put together**

https://github.com/user-attachments/assets/e42f93cb-6280-4d1a-a346-30d7f4bd420b

# Technical Notes

Untitled-2024-03-04-1546

We modify the xml directly in the frontend. Mostly we even work directly on the parsed xml document and only convert it back to an xml string when needed. This means that most interaction happens through side effects. Working directly on the xml has a bunch of advantages here (see below) but this needs to be carefully done and contained to not get too many of the negative effects of side-effects.

### Block selecting over html element borders

`<div>Hallo <a>Test</a></div>`
Refs need to be surroundable by an `<akn:ref>`-Element. Therefore a selection like `llo <a>Te` should not be possible.
To stop this we change the selection to always stay within the same element: https://github.com/digitalservicebund/ris-norms/blob/5718756a53e0b2cb7c9ca55c2aed61999e8e0c28/frontend/src/components/RisLawPreview.vue#L256

### React to finishing with a selection
There is no event like `selectionfinished` but only `seletionchange` and `selectionstart`. Therefore there is no simple way to detect when a selection is done. For the prototype we use a debounce and say the selection is finished when it hasn't changed for a while. It might be possible to also use something like `keyup` but we might need to keep keyboard based selection in mind for this.

### Creating a character range to find the selection in the xml

So far the selection was only created in the html-render of the norm. To create an akn:ref element we need to find the selected range in the xml.
For this we use the eid of the surrounding element (in the html its in the `data-eid` property) and a character count (only counting non-whitespace characters to not need to deal with indentation)

### No nested refs

For the prototype we decided agains allowing to mark nested refs (refs within refs). This allows to still select text within a ref and use copy and past to copy the text to somewhere.

## Creating a new ref from a text selection

For creating the akn:ref element we create a `Range` with the node (identified by the eid) and the character count from the selection (for this we need to add the whitespace again). We then create the  `akn:ref` element and set it's properties and then use `range.surroundContents(refElement)` to place the ref in the existing xml document.
https://github.com/digitalservicebund/ris-norms/blob/5718756a53e0b2cb7c9ca55c2aed61999e8e0c28/frontend/src/views/AmendingLawTextMetadataEditor.vue#L172

## Modifying a akn:ref

We modify only properties of akn:ref elements. If we want to change the selected range we need to delete the existing ref and create a new one.

The modifications are done in the RisRefEditor component (https://github.com/digitalservicebund/ris-norms/blob/5718756a53e0b2cb7c9ca55c2aed61999e8e0c28/frontend/src/components/RisRefEditor.vue) This is passed the `Node` of the parsed xml that represents the akn:ref that will be edited by this component. **Therefore modifying some property of the node has side-effects as it is modifying the xml document used by all RisRefEditor's and the view**

This was advantageous for the prototype here as it allows us to have the logic for updating the properties in the RisRefEditor without needing to move this logic to the view. For a proper implementation I would reevaluate this as I think it would be best to keep all the code affecting one xml-document instance within one file to not have side-effects over multiple files.

## Deleting a akn:ref
We replace the akn:ref node by its child nodes https://github.com/digitalservicebund/ris-norms/blob/5718756a53e0b2cb7c9ca55c2aed61999e8e0c28/frontend/src/views/AmendingLawTextMetadataEditor.vue#L264

## Automatic scrolling

See here: https://github.com/digitalservicebund/ris-norms/blob/5718756a53e0b2cb7c9ca55c2aed61999e8e0c28/frontend/src/views/AmendingLawTextMetadataEditor.vue#L356 It might be a good idea to evaluate if `nearest` is the best mode

## Keyboard navigation in the form

We added support to navigate to the next/previous input field of the same type with the up and down arrows. We where able to identify the elements by using `document.querySelectorAll` and then finding our current element and calling `focus()` on the previous / next element of the results. To update the selected element after a focus change we listen to the `@focusin` event on the `RisRefEditor`.

## Copy paste a whole row

When copying we store a JSON (as string) representation in the clipboard. And when pasting we try to parse the clipboard content for json, if it is successful we fill the fields with the data otherwise we do a normal text paste.
We only copy a row using cmd-c if no text is highlighted so its still possible to copy text.

## Highlighting refs

We use css with the selector `:deep(.akn-ref)` in the view. This selects the elements in the rendered html of the RisPreview.

## Keeping the preview updated

As we are modifying the xml-document directly we can no longer rely on vue to automatically detect changes and update eg. the preview of the xml and the html-render.
We therefore use manually thrown events to tell the view when to convert the xml-document to a string to update the previews.

## Saving to the backend

We convert the xml-document to a string and send it to the backend to replace the existing xml of the norm.

## Why are we doing all of this in the frontend?

The only calls to the backend are currently for the initial loading of data, for updating the html-render and for saving. Everything else happens client side. (We also quickly evaluated if we can move the xlst for rendering the html to the frontend, this is also possible at the moment but would need some minor modifications of the xslt so it works with xslt 1.0)

This ensures that the interactions quickly lead to visible changes without needing to wait for requests. This is especially imporant as we always load the visible information from the xml-document and do not cache it somewhere. Therefore the new information would (if we modify the xml after every edit in the backend) only visible after some time. Especially when creating many refs quickly (by highlighting) this could lead to the application being/feeling very slow.


andreasphil:
Assigning myself as a reviewer just so I remember to have a closer look 😇

The recording already looks super nice!

PR #441, opened 2024-06-27
Risdev 4024 exception handling reloaded

SebastianRossa:


backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/controller/ProprietaryController.java
```diff
@@ -4,12 +4,12 @@
 
 import de.bund.digitalservice.ris.norms.adapter.input.restapi.mapper.ProprietaryResponseMapper;
 import de.bund.digitalservice.ris.norms.adapter.input.restapi.schema.ProprietarySchema;
+import de.bund.digitalservice.ris.norms.application.exception.NormNotFoundException;
```
VictorDelCampo: Did you remove the try/catch in the endpoints of this controller (and the other controllers, if present) for the `NormNotFoundException`? We don't need them anymore because we are handling now this exception also in the controller advice.

doc/adr/0011-concept-exception-handling-backend.md
```diff
@@ -0,0 +1,33 @@
+# 10. How to use exceptions in the backend
+
+Date: 2024-06-28
+
+## Status
+
+Accepted
+
+## Context
+So far we used different approaches when to use Optionals, throw checked exceptions or unchecked exceptions. This also led
```
VictorDelCampo, line 10: I would remove the topic with optionals since the ADR is related to the exception handling. That we throw application exceptions in some cases when we get an optional from the db layer is in my opinion not part of this ADR

doc/adr/0011-concept-exception-handling-backend.md
```diff
@@ -0,0 +1,33 @@
...
+to using exceptions from other layers. We want to streamline our approach.
```
VictorDelCampo, line 11: I would also add: "...to using exceptions from other layers, which was violating the hexagonal architecture approach"

doc/adr/0011-concept-exception-handling-backend.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Decision
+Layer-Specific Exceptions
+* By having each layer (ports, adapters, and application) define its own exceptions, we achieve a clear separation of concerns. This ensures that exceptions are relevant to the specific context in which they occur.
```
VictorDelCampo, line 15: The ports are not a layer by themselves, but live within the application layer. Meaning the layers that we have are: application (also containing the domain) and the adapter layer (containing both input - api - and also output - db)

doc/adr/0011-concept-exception-handling-backend.md
```diff
@@ -0,0 +1,33 @@
...
+* Avoiding a common exception package prevents mixing unrelated exceptions and promotes a more organized codebase.
```
VictorDelCampo, line 16: Maybe also add a short sentence what we had the idea of a common package for exceptions --> so that we can reuse exceptions in different layers

doc/adr/0011-concept-exception-handling-backend.md
```diff
@@ -0,0 +1,33 @@
...
+
+Custom Runtime Exceptions
+* Throwing custom runtime exceptions in the application layer allows us to encapsulate domain-specific logic. These exceptions can be tailored to the norms project’s requirements.
+* Using @ControllerAdvice to handle these exceptions ensures consistent error handling across the application. It centralizes exception handling logic and provides a single point of control.
```
VictorDelCampo, line 20: I would maybe add here the main argument for this approach and that is that our backend app is a merely API supporting app, so that we can centralized the handling of the exceptions in the controller layer

doc/adr/0011-concept-exception-handling-backend.md
```diff
@@ -0,0 +1,33 @@
...
+
+Output Ports
+* Returning optionals from output ports aligns with functional programming principles. Optionals allow for more elegant handling of absence or optional results.
```
VictorDelCampo, line 23: As mentioned above the decision about keeping the optionals should not be part of this ADR about exception handling in the backend right?

doc/adr/0011-concept-exception-handling-backend.md
```diff
@@ -0,0 +1,33 @@
...
+* Translating exceptions thrown in output ports at the application layer maintains a clean boundary between layers. It avoids leaking implementation details and promotes a clear contract.
```
VictorDelCampo, line 24: I would maybe add here, since we don't have any exceptions being thrown in the output ports, that in case we want to throw an exception we would translate it from the output port to an application exception

doc/adr/0011-concept-exception-handling-backend.md
```diff
@@ -0,0 +1,33 @@
...
+* The output port is not allowed not have a dependency on the application layer’s exceptions.
```
VictorDelCampo, line 25: Type, twice a "not"

doc/adr/0011-concept-exception-handling-backend.md
```diff
@@ -0,0 +1,33 @@
...
+
+Input Ports
+* Allowing input ports to depend on application layer exceptions acknowledges that input ports may need to handle specific application-related errors.
```
VictorDelCampo, line 28: And this is by default right? We don't need to allow it extra but since the input ports are implemented in the application layer of course they can access application exceptions

doc/adr/0011-concept-exception-handling-backend.md
SebastianRossa: My perspective is that Optionals are a kind of checked exception in this case and therefore should be part of this discussion.

doc/adr/0011-concept-exception-handling-backend.md
SebastianRossa: see above

doc/adr/0011-concept-exception-handling-backend.md
SebastianRossa: The thought here was that we define the rules in our architecture tests. We could have decided otherwise.

PR #443, opened 2024-06-28
Implement GET metadata at time boundary for single element

reckseba:
RISDEV-4281

### DIFF
--- a/backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/controller/ProprietaryController.java
+++ b/backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/controller/ProprietaryController.java
@@ -4,6 +4,7 @@ import static org.springframework.http.MediaType.APPLICATION_JSON_VALUE;
 
 import de.bund.digitalservice.ris.norms.adapter.input.restapi.mapper.ProprietaryResponseMapper;
 import de.bund.digitalservice.ris.norms.adapter.input.restapi.schema.ProprietarySchema;
+import de.bund.digitalservice.ris.norms.adapter.input.restapi.schema.ProprietarySingleElementSchema;
 import de.bund.digitalservice.ris.norms.application.port.input.LoadProprietaryFromNormUseCase;
 import de.bund.digitalservice.ris.norms.application.port.input.UpdateProprietaryFromNormUseCase;
 import de.bund.digitalservice.ris.norms.domain.entity.Eli;
@@ -99,6 +100,34 @@ public class ProprietaryController {
                       proprietarySchema.getOrganisationsEinheit())));
 
       return ResponseEntity.ok(ProprietaryResponseMapper.fromProprietary(proprietary, atDate));
+
+    } catch (NormNotFoundException e) {
+      return ResponseEntity.notFound().build();
+    }
+  }
+
+  /**
+   * Return specific metadata of a single element at a given date within the {@link Proprietary}
+   * node.
+   *
+   * @param eli Eli of the ZF0 version of the target law where the metadata are saved.
+   * @param eid the eId of the single element within the ZF0
+   * @param atDate the time boundary at which to return the metadata
+   * @return the specific metadata returned in the form of {@link ProprietarySingleElementSchema}
+   */
+  @GetMapping(
+      path = "/{eid}/{atDate}",
+      produces = {APPLICATION_JSON_VALUE})
+  public ResponseEntity<ProprietarySingleElementSchema> getProprietaryAtDate(
+      final Eli eli, @PathVariable final String eid, @PathVariable final LocalDate atDate) {
+
+    try {
+      var proprietary =
+          loadProprietaryFromNormUseCase.loadProprietaryFromNorm(
+              new LoadProprietaryFromNormUseCase.Query(eli.getValue()));
+
+      return ResponseEntity.ok(
+          ProprietaryResponseMapper.fromProprietarySingleElement(proprietary, eid, atDate));
 
     } catch (NormNotFoundException e) {
       return ResponseEntity.notFound().build();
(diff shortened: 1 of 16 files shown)


backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/controller/ProprietaryController.java
```diff
@@ -104,4 +105,32 @@ public ResponseEntity<ProprietarySchema> updateProprietaryAtDate(
...
+
+    try {
+      var proprietary =
+          loadProprietaryFromNormUseCase.loadProprietaryFromNorm(
+              new LoadProprietaryFromNormUseCase.Query(eli.getValue()));
+
+      return ResponseEntity.ok(
+          ProprietaryResponseMapper.fromProprietarySingleElement(proprietary, eid, atDate));
+
+    } catch (NormNotFoundException e) {
```
VictorDelCampo, line 132: We will soon merge the branch with the exception handling in the backend and we won't need to catch the NormNotFoundException as it will be handled in the controller advice. Sebastian will remove the try/catching of the old endpoints so I would say why not also remove it here. I don't think we would have problems till Sebastian's branch is merged or do we have an integration controller test that is testing the not found exception and what http code is returned?

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/Proprietary.java
```diff
@@ -225,4 +225,18 @@ public Optional<String> getOrganisationsEinheit(final LocalDate date) {
...
+
+  /**
+   * Returns the ("Art der Norm") which can be "SN or ÄN or ÜN" of the document from the MetadatenDs
+   * block for a single element at a specific date.
+   *
+   * @param eid the eId of the single element
+   * @param date the specific date of the time boundary.
+   * @return "Art der Norm" or empty if it doesn't exist.
+   */
+  public Optional<String> getSingleElementArtDerNorm(final String eid, final LocalDate date) {
```
VictorDelCampo: Should we just maybe overload the old `getArtDerNorm` method and not have a new name?

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/MetadatenDs.java
```diff
@@ -91,4 +92,20 @@ public void setAttributeOfSimpleMetadatum(
...
+  /**
+   * It returns the value for a specific metadata from a single element referenced by its eid for a
+   * specific date.
+   *
+   * @param metadatumSingleElement the enum metadatum
+   * @param eid the eid of the single element
+   * @param date the selected date
+   * @return value of the requested metadata element or empty if it doesn't exist
+   */
+  public Optional<String> getMetadatenDsSingleElement(
```
VictorDelCampo: Not happy with this naming. What about aligning the name to what we already had? The abstract class `Metadaten` is defining `getSimpleMetadatum` for extracting the value of a simple metadatum (meaning no child element) and maybe we need to rename that method to `getFrameSimpleMetadatum` ?  So that we can name this new method `getSingleElementSimpleMetadatum`?

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/Einzelelement.java
```diff
@@ -0,0 +1,34 @@
...
+  public Einzelelement(final Node node) {
+    super(node, "start", "end");
+  }
+
+  /**
+   * The list of all simple metadata within the meta:legalDocML.de_metadaten_ds block. They consist
+   * of a single string property.
+   */
+  public enum Metadata implements MetadataInterface {
+    ART_DER_NORM("./artDerNorm");
```
VictorDelCampo, line 21: Should we maybe create a general enum with metadata (and xpath) present in several metadata blocks, like for example the FNA in MetadatenDe and MetadatenDs and artDerNorm in both MetadatenDs and Einzelelemente? Or you think is more readable to duplicate the metadata definition with their xpaths in every metadata block?

backend/src/test/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/controller/ProprietaryControllerTest.java
```diff
@@ -266,4 +266,91 @@ void itReturnsNotFoundIfNormIsNotFound() throws Exception {
           .andExpect(status().isNotFound());
     }
   }
+
+  @Nested
+  class getProprietaryEinzelelementAtDate {
+    @Test
+    void returns404IfNormNotFound() throws Exception {
```
VictorDelCampo, line 273: I just saw now this test. Meaning we can't remove the try/catch of the controller. It would be great if we can merge this branch before Sebastian is done, so that he can remove the try/catch there

VictorDelCampo (changes requested):
It looks already really nice. I just have some proposals for naming and possible further optimizations. If you wanna talk about them, lets sync later :)

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/controller/ProprietaryController.java
reckseba: Yes, there is.

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/Proprietary.java
reckseba: Wonderful idea!

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/MetadatenDs.java
reckseba: Agree!

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/Einzelelement.java
reckseba: I don't get this one. We might need to discuss.

backend/src/test/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/controller/ProprietaryControllerTest.java
reckseba: Ok for me

PR #454, opened 2024-07-02
RISDEV-4297 connect frontend and backend

malte-laukoetter:


backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/controller/NormController.java
```diff
@@ -235,7 +237,8 @@ public ResponseEntity<UpdateModResponseSchema> updateMod(
       produces = {APPLICATION_JSON_VALUE})
   public ResponseEntity<UpdateModsResponseSchema> updateMods(
       final Eli eli,
-      @RequestBody @Valid final Map<String, UpdateModRequestSchema> updateModsRequestSchema,
+      @RequestBody @Valid
+          final Map<String, UpdateModsRequestSchema.ModUpdate> updateModsRequestSchema,
```
VictorDelCampo, line 240: Why such a complicated request schema? If I am not wrong the JSON would look like:

```
{
  "mod1": {
    "timeBoundaryEid": "eid1"
  },
  "mod2": {
    "timeBoundaryEid": "eid2"
  },
  "mod3": {
    "timeBoundaryEid": "eid3"
  }
}
```
I do get your point of having this clarify of "we use the mod eid to update the temporary eid of it" but I find the Map code in the backend (both in the controller as well as in the use case and service) too overcomplicated.

If you don't like
```
[
  {
    "modEid": "mod1",
    "timeBoundaryEid": "eid1"
  },
  {
    "modEid": "mod2",
    "timeBoundaryEid": "eid2"
  },
  {
    "modEid": "mod3",
    "timeBoundaryEid": "eid3"
  }
]
```
what about:
```
[
  {
    "mod1": "eid1"
  },
  {
    "mod2": "eid2"
  },
  {
    "mod3": "eid3"
  }
]
```
?

malte-laukoetter: I changed it so the UseCase works with a collection but kept the request body as is (your first example is correct). I think this gives us the best of both worlds.

PR #456, opened 2024-07-03
Rename first field under Federführung to Ressort

reckseba:
RISDEV-4309


reckseba:
@VictorDelCampo I don't like it as in the end it is `federfuehrend` in the xml. Would you say we do it as proposed or only naming the field "Ressort" and the rest stays as it is?

VictorDelCampo:
> @VictorDelCampo I don't like it as in the end it is `federfuehrend` in the xml. Would you say we do it as proposed or only naming the field "Ressort" and the rest stays as it is?

@reckseba I would name the variables everywhere as the labels in the frontend, meaning:
- current `normgeber` would be now `staat`
- current `federfuehrung` should be called `ressort`

The case with `ressort` being mapped to `federfuehrung/federfuehrend` is not ideal but this metadatum (which as per definition by the old DES-N schema, the one defining the data structure for the metadata, a combination of `ressort` and `organisationsEinheit`) is a special case because it was already present in LDML.de but completely refined as the documentalists need. Therefore I would say we live for now with the naming conflict

As of `normgeber` being renamed to `staat` I would do raise to PM the issue that the metadatum we created in the DS block should also be renamed to `staat`. What do you say?

reckseba:
@VictorDelCampo ok, can merge?

PR #461, opened 2024-07-05
Update XML editor content after changes

reckseba:
RISDEV-4191

### DIFF
--- a/frontend/src/composables/useNormRender.ts
+++ b/frontend/src/composables/useNormRender.ts
@@ -23,7 +23,12 @@ export function useNormRender(
       if (!toValue(normXml)) return INVALID_URL
 
       const searchParams = new URLSearchParams()
-      searchParams.set("showMetadata", toValue(showMetadata) ? "true" : "false")
+      if (renderHtml) {
+        searchParams.set(
+          "showMetadata",
+          toValue(showMetadata) ? "true" : "false",
+        )
+      }
 
       const atValue = toValue(at)
       if (atValue) {
--- a/frontend/src/composables/useNormRender.ts
+++ b/frontend/src/composables/useNormRender.ts
@@ -69,7 +69,11 @@ export function useNormRenderXml(
         searchParams.set("atIsoDate", atValue.toISOString())
       }
 
-      return "renderings?" + searchParams.toString()
+      return (
+        "renderings" +
+        (searchParams.size > 0 ? "?" : "") +
+        searchParams.toString()
+      )
     }),
     {
       headers: {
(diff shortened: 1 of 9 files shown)


reckseba:
@malte-laukoetter would you say that this is a valid attempt? Let's discuss. If it's fine I'd provide the tests.

malte-laukoetter:
> @malte-laukoetter would you say that this is a valid attempt? Let's discuss. If it's fine I'd provide the tests.

yes looks like a good solution, I'm fine with reusing the endpoint for the html render also for the — let's call it xml-render — of the norm at a specific point in time ^^

frontend/src/composables/useNormRender.ts
```diff
@@ -7,12 +7,14 @@ import { UseFetchReturn } from "@vueuse/core"
...
  * @param normXml XML of the norm that should be rendered
  * @param showMetadata Enable or disable metadata list at the beginning of the document
+ * @param renderHtml By default (true) renders HTML. When false then XML.
  * @param customNorms The XMLs of norms which are referenced by the norm (e.g. in passiveModifications) and should be used instead of the data stored.
  * @param at Passive modifications coming into effect before this date should be applied before rendering the HTML
  */
 export function useNormRender(
   normXml: MaybeRefOrGetter<string | undefined>,
   showMetadata: MaybeRefOrGetter<boolean> = false,
+  renderHtml: MaybeRefOrGetter<boolean> = true,
```
malte-laukoetter: I'm unsure if this additional flag is the best solution here - or if we want to just include a second method. Also with this many parameters (and multiple once that are booleans) I would find it nice if we could move to an options object for all the optional parameters.

reckseba: Alright! I wavered between the two options. I removed the parameter and introduced a second method instead.

If this is fine, I'll provide the missing FE & BE tests on Monday.

reckseba:
@malte-laukoetter I provided the necessary tests. Though the e2e test is failing and I actually have no clue why. Do you?

malte-laukoetter:
> @malte-laukoetter I provided the necessary test. Though the e2e test is failing and I actually have no clue why. Do you?

The xml editor is using virtual scrolling to not need to render the complete xml all the time. Therefore the new text might not be rendered on the page and can therefore not be found by playwright. I don't think i ever found a good solution to that problem :/ Maybe you have an idea?

malte-laukoetter (approved):

reckseba:
> > @malte-laukoetter I provided the necessary test. Though the e2e test is failing and I actually have no clue why. Do you?
> 
> The xml editor is using virtual scrolling to not need to render the complete xml all the time. Therefore the new text might not be rendered on the page and can therefore not be found by playwright. I don't think i ever found a good solution to that problem :/ Maybe you have an idea?

That's why.. You're right. Thanks! 

And yes. I had and idea: There is this search feature. I used that one. Good?

PR #480, opened 2024-07-18
Risdev 4411 update api spec

SebastianRossa:


doc/backend/norms-api-v1.yaml
```diff
@@ -659,17 +659,31 @@ components:
 
     Mod:
       type: object
+      required:
```
VictorDelCampo, line 662: the timeBoundaryEid is also required right?

PR #486, opened 2024-07-24
Rendering XML snippets

VictorDelCampo:
In the quotedStructure team we re-use the rendering XML-->HTML endpoint to render the content within quotedStructure into the RisModForm.

This worked without problems till we created new synthetic data with quotedStructures containing several children. This lead to an error when trying to:
1. Parse that XML snippet to a `w3c.dom.Document` (which is in my opinion only needed for complete a `Norm`)
2. Even also after skipping the `render()` method within the `RenderingController`, the transformation of the snippet using XSLT also failed.

The same endpoint was already being used for a snippet of XML but the snippet was having only 1 root element (article rendering).

Therefore I introduced a new optional query parameter to explicitly denotate that it is an XML snippet, so that the render() method is skipped (circumventing 1.) and to add a root `akn:akomaNtoso` element to the snippet.

### DIFF
--- a/backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/XsltTransformationService.java
+++ b/backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/XsltTransformationService.java
@@ -39,10 +39,16 @@ public class XsltTransformationService implements TransformLegalDocMlToHtmlUseCa
       transformer.setOutputProperty(OutputKeys.ENCODING, "UTF-8");
       transformer.setParameter(SHOW_METADATA, query.showMetadata());
       transformer.setParameter("outputMode", "html");
-
+      String inputXml = query.xml();
+      if (query.snippet()) {
+        inputXml =
+            "<akn:akomaNtoso xmlns:akn=\"http://Inhaltsdaten.LegalDocML.de/1.6/\">"
+                + inputXml
+                + "</akn:akomaNtoso>";
+      }
       StringWriter output = new StringWriter();
       transformer.transform(
-          new StreamSource(new ByteArrayInputStream(query.xml().getBytes())),
+          new StreamSource(new ByteArrayInputStream(inputXml.getBytes())),
           new StreamResult(output));
       return output.toString();
     } catch (IOException | TransformerException e) {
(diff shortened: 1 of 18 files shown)


frontend/src/composables/useNormRender.ts
```diff
@@ -7,12 +7,14 @@ import { UseFetchReturn } from "@vueuse/core"
...
  * @param customNorms The XMLs of norms which are referenced by the norm (e.g. in passiveModifications) and should be used instead of the data stored.
  * @param at Passive modifications coming into effect before this date should be applied before rendering the HTML
  */
 export function useNormRenderHtml(
   normXml: MaybeRefOrGetter<string | undefined>,
   showMetadata: MaybeRefOrGetter<boolean> = false,
+  snippet: MaybeRefOrGetter<boolean> = false,
   at?: MaybeRefOrGetter<Date | undefined>,
   customNorms?: MaybeRefOrGetter<string[] | undefined>,
 ): UseFetchReturn<string> {
```
malte-laukoetter, line 23: with that amount of optional parameters i would start thinking about moving them to an options object, so that the places calling this method only need to provide the parameters necessary (and it is less confusing which parameter is what _"was snippet the second or third parameter?"_) But no need to change it here now :)

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/XsltTransformationService.java
```diff
@@ -39,10 +39,16 @@ public String transformLegalDocMlToHtml(TransformLegalDocMlToHtmlUseCase.Query q
       transformer.setOutputProperty(OutputKeys.ENCODING, "UTF-8");
       transformer.setParameter(SHOW_METADATA, query.showMetadata());
       transformer.setParameter("outputMode", "html");
-
+      String inputXml = query.xml();
+      if (query.snippet()) {
+        inputXml =
+            "<akn:akomaNtoso xmlns:akn=\"http://Inhaltsdaten.LegalDocML.de/1.6/\">"
```
malte-laukoetter, line 45: l like this, it also already solves the eventual problem that snippets do not include the namespace <3

malte-laukoetter (approved):

frontend/src/composables/useNormRender.ts
VictorDelCampo: I understand. I will change it now

PR #489, opened 2024-07-25
RISDEV-4434 create page for reference documentation no functionality

malte-laukoetter:


frontend/src/components/references/RisModSelectionPanel.vue
```diff
@@ -0,0 +1,57 @@
...
+} = useNormRenderHtml(props.normXml)
+
+function handleAknModClick({ eid }: AknElementClickEvent) {
+  selectedModEId.value = eid
+}
+</script>
+
+<template>
+  <div class="flex flex-col">
+    <div v-if="normRenderLoading" class="flex justify-center">
```
andreasphil, line 34: Suggestion: I know replacing the UI with the spinner is our usual pattern but in this case it makes the UI feel a bit janky when showing a spinner for a split-second after every highlight or removing of a highlight. Some possible alternatives that come to mind:

- Debounce the spinner
- Show the spinner with a semi-transparent overlay on top of the UI instead of replacing it (possibly combined with debouncing)
- Show a small spinner inline next to the heading

frontend/src/components/references/RisModSelectionPanel.vue
```diff
@@ -0,0 +1,57 @@
...
+      :selected="selectedModEId ? [selectedModEId] : []"
+      :content="normRender ?? ''"
+      @click:akn:mod="handleAknModClick"
+    />
+  </div>
+</template>
+
+<style scoped>
+:deep(.akn-mod) {
+  @apply block border border-[#4299F7] bg-[#E7E7E766] px-2;
```
andreasphil: Question: should we add these colors to our Tailwind config rather than hardcoding/arbitrary variants? Or do we know they won't be used anywhere else?

frontend/src/components/references/RisModSelectionPanel.vue
```diff
@@ -0,0 +1,57 @@
...
+      class="ds-textarea flex-grow p-2"
+      :selected="selectedModEId ? [selectedModEId] : []"
+      :content="normRender ?? ''"
+      @click:akn:mod="handleAknModClick"
+    />
+  </div>
+</template>
+
+<style scoped>
+:deep(.akn-mod) {
```
andreasphil, line 52: Question: I'm not a huge fan of using `:deep`. I think it's fine for the preview component because of the `v-html`, but I would prefer if we didn't have this all across the app because it makes it harder to understand where styles come from, and harder to change the preview component as it also needs to work with all kinds of unpredictable styling overrides. Maybe we can find a different way to solve this?

andreasphil: Side note: If we break out of the scoping anyways, maybe `style module` or simply `style` would be cleaner?

frontend/src/components/references/RisRefEditor.vue
```diff
@@ -0,0 +1,23 @@
...
+}>()
+</script>
+
+<template>
+  <div>Typ Select</div>
+  <div>ELI Input</div>
+  <RisTextButton
+    size="small"
+    variant="ghost"
+    label="Löschen"
```
andreasphil, line 24: Question: do we need to generate a more dynamic/precise label here? I don't know how this will be logically grouped with the other inputs in this row but I'm not sure simply saying "Löschen" will be enough from the accessibility perspective to make the connection to what is being deleted.

frontend/src/services/xmlService.ts
```diff
@@ -1,5 +1,5 @@
 /**
- * Convert the text representation of an xml document to a Document
+ * Convert the text representation of a xml document to a Document
```
andreasphil, line 2: Nitpick: I think `an` is actually correct here 😀

frontend/src/components/references/RisModSelectionPanel.vue
```diff
@@ -0,0 +1,57 @@
+<script setup lang="ts">
+import RisLawPreview, {
+  AknElementClickEvent,
+} from "@/components/RisLawPreview.vue"
+import { useNormRenderHtml } from "@/composables/useNormRender"
+import RisLoadingSpinner from "@/components/controls/RisLoadingSpinner.vue"
+import RisCallout from "@/components/controls/RisCallout.vue"
+
+const props = defineProps<{
+  normXml: string
```
andreasphil, line 13: Todo: docs

frontend/src/components/references/RisRefSelectionPanel.vue
```diff
@@ -0,0 +1,140 @@
...
+        @mousedown="handleSelectionStart"
+        @mouseup="handleSelectionEnd"
+        @click:akn:ref="handleAknRefClick"
+      />
+    </div>
+  </div>
+</template>
+
+<style scoped>
+:deep(.akn-quotedStructure .akn-quote-startQuote) {
```
andreasphil, line 145: Same here as above regarding the `deep` selectors

frontend/src/composables/useAknTextSelection.ts
```diff
@@ -0,0 +1,106 @@
...
+        selection.anchorNode,
+        selection.anchorOffset,
+        selection.anchorNode,
+        0,
+      )
+      break
+  }
+}
+
+export function useAknTextSelection(
```
andreasphil, line 79: todo: docs

frontend/src/lib/htmlRangeToLdmlDeRange.ts
```diff
@@ -0,0 +1,129 @@
...
+    return null
+  }
+
+  const childIndex = Array.from(node.parentElement.childNodes).indexOf(node)
+
+  return getNodeByEid(ldmlDocument, eId)?.childNodes.item(childIndex) ?? null
+}
+
+/**
+ * Finds the offset that is the corresponds to the same text range in the LDML.de node.
```
andreasphil, line 51: todo: IMO this could need some additional explanation about the algorithm that is used for finding the offset

frontend/src/lib/htmlRangeToLdmlDeRange.ts
```diff
@@ -0,0 +1,129 @@
...
+      ldmlOffsetWithNormalizedWhitespace++
+    }
+    ldmlOffset++
+  }
+
+  return ldmlOffset
+}
+
+/**
+ * Converts a Range within a html-render of a LMDL.de xml to a range within that xml document.
```
andreasphil, line 87: todo: same here, some explanation about the algorithm would be appreciated

frontend/src/lib/ref.ts
```diff
@@ -0,0 +1,50 @@
...
+  if (!parentElement.ownerDocument) {
+    return
+  }
+
+  const aknRef = parentElement.ownerDocument.createElementNS(
+    "http://Inhaltsdaten.LegalDocML.de/1.6/",
+    "ref",
+  )
+  aknRef.setAttribute("eId", getNextRefEId(parentElement))
+  aknRef.setAttribute("GUID", uuidv4())
```
andreasphil, line 47: Question: is this a temporary GUID or will this also be the GUID in the backend? If it's the latter, it would feel better to me to have one actor in charge of generating IDs (in our case: the backend) to ensure data integrity and uniqueness (even though I admit that the likelihood of conflicts is low).

frontend/src/views/References.vue
```diff
@@ -0,0 +1,221 @@
...
+    }
+
+    const newNode = xmlStringToDocument(newValue).firstChild
+    if (!newNode) {
+      return
+    }
+
+    parentNode.replaceChild(newNode, selectedModQuotedContent.value)
+    // we mutate the norm Document (as a side effect), so we need to trigger a recompute of all things that use it
+    triggerRef(amendingNormDocument)
```
andreasphil, line 102: I didn't know about `triggerRef`, nice! 💡

andreasphil (changes requested):

frontend/src/components/references/RisModSelectionPanel.vue
malte-laukoetter: The previous solution was to have all the different highlighting styles within the preview component itself (and properties to turn them on or off), but that became quite messy with the amount of different highlighting we want to have on different pages. 

The scoped style still (even with the deep) ensures that nothing outside of the component is affected by the style (but other than without the deep child components can be affected). So it still has advantages above a simple `style`. With `style module` I'm unsure how to achieve this kind of styling.

So yes it's not ideal but I couldn't find a better solution.

frontend/src/lib/ref.ts
malte-laukoetter: It's the final one. But as we are using UUIDs here (and especially v4 once) we do not have any data integrity or uniqueness problems as they are designed with the explicit purpose that lots of distributed actors can generate them and there will not be any conflicts. So it is totally fine to create the GUID on the frontend.

frontend/src/components/references/RisRefEditor.vue
malte-laukoetter: I'll look into this when I implement the other inputs of the component

frontend/src/services/xmlService.ts
malte-laukoetter: atleast intellij complaint about "an" but I don't know 🤷

frontend/src/components/references/RisModSelectionPanel.vue
malte-laukoetter: I think we should but they are currently different from the colors already defined for mods in other views so I don't know what to call them. Will talk to design about this ^^

frontend/src/components/references/RisRefEditor.vue
malte-laukoetter: I have grouped the inputs into sections which have the text content of the akn:refs as names

PR #500, opened 2024-08-02
Debug preview issues

andreasphil:
We're currently experiencing issues with the HTML preview on staging which can't be reproduced locally. This PR is for debugging (and hopefully fixing) the issue.

**Edit:** The issue was caused by a dependency update. This PR reverts the update (see commits)


andreasphil:
Related: https://github.com/getsentry/sentry-javascript/issues/13187

PR #505, opened 2024-08-05
Call references documentation from frontend RISDEV-4482

hamo225:


frontend/e2e/amending-law-references.spec.ts
```diff
@@ -11,6 +11,67 @@ test("navigate to amending law overview", async ({ page }) => {
...
+      },
+    })
+  })
+
+  await page.goto(
+    "/amending-laws/eli/bund/bgbl-1/1002/10/1002-01-10/1/deu/regelungstext-1/affected-documents/eli/bund/bgbl-1/1002/1/1002-01-10/1/deu/regelungstext-1/references",
+  )
+
+  // Check if the alert is displayed
+  const alert = page.locator('div[role="alert"]')
```
andreasphil, line 63: Todo: Please [don't use locators](https://digitalservicebund.atlassian.net/wiki/spaces/VER/pages/884310020/Frontend+conventions+and+patterns#Use-specific-queries.-Avoid-locator-and-querySelector.) if alternatives are available. Here `getByRole("alert")` should work.

frontend/e2e/amending-law-references.spec.ts
```diff
@@ -11,6 +11,67 @@ test("navigate to amending law overview", async ({ page }) => {
...
+  await expect(alert).toBeVisible()
+  await expect(alert).toContainText(
+    "Die automatische Referenzierung konnte nicht durchgeführt werden.",
+  )
+
+  const closeButton = alert.locator('button[aria-label="Schließen"]')
```
andreasphil, line 69: Todo: Please [don't use locators](https://digitalservicebund.atlassian.net/wiki/spaces/VER/pages/884310020/Frontend+conventions+and+patterns#Use-specific-queries.-Avoid-locator-and-querySelector.) if alternatives are available. Here `getByRole("button", { name: "Schließen" })` should work.

frontend/src/composables/useFetchReferences.ts
```diff
@@ -0,0 +1,31 @@
+import { MaybeRefOrGetter, toValue, computed } from "vue"
```
andreasphil, line 1: To be consistent with our other services, can you change this like

- File name: `services/referencesService.ts`
- Function name: `useGetReferences`

Implementation LGTM

frontend/src/views/References.vue
```diff
@@ -60,10 +62,56 @@ function handleSave(xml: string) {
...
+
+interface ApiError {
+  response?: {
+    status?: number
+    message?: string
+  }
+}
+
+function isNotFoundError(error: ApiError): boolean {
+  console.log(error?.response?.status)
```
andreasphil, line 91: Todo: remove `console.log` pls. Now that we have sentry this will just pop up in our logs and make people wonder if something is wrong with the app.

frontend/src/views/References.vue
```diff
@@ -60,10 +62,56 @@ function handleSave(xml: string) {
...
+}
+
+interface ApiError {
+  response?: {
+    status?: number
+    message?: string
+  }
+}
+
+function isNotFoundError(error: ApiError): boolean {
```
andreasphil, line 90: Note: This should no longer be necessary (see previous comment) but if it was, this too should be a computed prop:

```
const referencesNotFound = computed(() => error.value?.response?.status === 404);
```

The reason is that we should [avoid calling functions](https://cdruc.com/vue-function-templates) in templates when possible. I don't think it would be a huge issue here, just in general a bad habit that we shouldn't get used to 😛 (I know it can be tempting)

As a side-note, should we remove the `ApiError` interface here and just leave the error response `any`? Since we'll be implementing more API error handling soon, I'd like to avoid having these localized solutions all over the codebase. If we need the the ApiError interface, please move it to the `types` folder.

frontend/src/views/References.vue
```diff
@@ -60,10 +62,56 @@ function handleSave(xml: string) {
   newNormXml.value = xml
   save()
 }
+
+const isFetchingReferences = ref()
```
andreasphil, line 66: Not sure I understand this entire block from line 66-97. Why wrap the `useFetchReferences` in a function and sync it's output to these variables? Especially since it's called unconditionally in `onMounted`. Wouldn't it be a lot simpler to implement it like this:

```ts
const { isFetching: isFetchingReferences, error: referencesError } = useFetchReferences(amendingNormEli);

const showReferencesError = computed(() => 
  referencesError.value && referencesError.value.response?.status !== 404
)
```

frontend/src/views/References.vue
```diff
@@ -60,10 +62,56 @@ function handleSave(xml: string) {
...
+onMounted(() => {
+  fetchReferences()
+})
 </script>
 
 <template>
   <div class="flex h-[calc(100dvh-5rem-1px)] flex-col bg-gray-100">
+    <RisAlert
+      v-if="
+        showErrorAlert && referencesError && !isNotFoundError(referencesError)
```
andreasphil, line 104: Todo: this can then be `v-if="showReferencesError"`

frontend/src/views/References.vue
```diff
@@ -60,10 +62,56 @@ function handleSave(xml: string) {
...
+}
+
+onMounted(() => {
+  fetchReferences()
+})
 </script>
 
 <template>
   <div class="flex h-[calc(100dvh-5rem-1px)] flex-col bg-gray-100">
+    <RisAlert
```
andreasphil, line 102: Todo: use `RisCallout`, `RisAlert` will be removed soon (see also https://github.com/digitalservicebund/ris-norms/commit/08d0aedebdaf59afdf75d2d02a0c5addd5513726)

```
<RisCallout
  v-if="showReferencesError"
  :title="Die automatische etc."
  variant="warning"
  allow-dismiss
/>
```

Suggestion: also switch to the warning variant—while technically an error, for the user's point of view I think it's more like a warning since it's not blocking them.

frontend/src/views/References.vue
```diff
@@ -96,8 +144,12 @@ function handleSave(xml: string) {
           <ModSelectionPanel
             v-if="amendingNormXml"
             v-model="selectedModEId"
-            class="overflow-hidden"
             :norm-xml="amendingNormXml"
+            :class="[
+              'overflow-hidden',
```
andreasphil, line 149: Todo: Can you also change the pointer to be crossed out? Otherwise people might just think the application is stuck

andreasphil (changes requested):

andreasphil:
Requested changes will be implemented in a follow-up PR

PR #511, opened 2024-08-08
Keyboard navigation in reference docs

andreasphil:


hamo225 (commented):
LGTM

frontend/src/components/RisLawPreview.vue
```diff
@@ -264,18 +284,100 @@ watch(
   },
   { deep: true },
 )
+
+/* -------------------------------------------------- *
+ * Focus handling                                     *
+ * -------------------------------------------------- */
+
+// NB: If we need this in more places in the future, consider extracting
+// this into a composable to reuse it in other places!
```
malte-laukoetter, line 293: i think i would have already moved this into a composable, in my experience that also makes testing a bit simpler ^^

frontend/src/components/RisLawPreview.vue
```diff
@@ -264,18 +284,100 @@ watch(
...
+
+const ids = useElementId()
+
+/**
+ * Used for keeping track of the elements that have been made interactive. Needed
+ * for focus handling.
+ */
+let interactiveEls: HTMLElement[] = []
```
malte-laukoetter, line 301: i needed to think a (short) moment about what Els means as my mind first went to elis

andreasphil: Wow I never thought about this but I think that's a habit from the [backbone days](https://backbonejs.org/#View-el) 😆  I see how that can be confusing. Will keep it in mind for next time

frontend/src/components/RisLawPreview.vue
andreasphil: I could not really decide on the right abstraction here. Thought it would be easier once we have more than one use case 😅

PR #518, opened 2024-08-12
Subpage for Outline Elements RISDEV-4527

hamo225:


frontend/e2e/amending-law-metadata-editor.spec.ts
```diff
@@ -139,7 +139,7 @@ test.describe("sidebar navigation", () => {
     page,
   }) => {
     await page.route(
-      /elements\?type=article&type=conclusions&type=preamble&type=preface&amendedBy=eli/,
+      /elements\?type=(article|conclusions|preamble|preface|book|chapter|part|subtitle|title)(&type=(article|conclusions|preamble|preface|book|chapter|part|subtitle|title))*&amendedBy=eli/,
```
andreasphil: Question: what does this do? This is not a route that exists in the backend.

frontend/e2e/amending-law-metadata-editor.spec.ts
```diff
@@ -158,7 +158,7 @@ test.describe("sidebar navigation", () => {
 
   test("shows an empty state when no elements are found", async ({ page }) => {
     await page.route(
-      /elements\?type=article&type=conclusions&type=preamble&type=preface&amendedBy=eli/,
+      /elements\?type=(article|conclusions|preamble|preface|book|chapter|part|subtitle|title)(&type=(article|conclusions|preamble|preface|book|chapter|part|subtitle|title))*&amendedBy=eli/,
```
andreasphil: Question: see above

frontend/src/router.ts
```diff
@@ -91,6 +91,12 @@ const routes: readonly RouteRecordRaw[] = [
         name: "AmendingLawMetadataEditorElement",
         component: () => import("@/views/AmendingLawMetadataEditorElement.vue"),
       },
+      {
+        path: ":timeBoundary/:eid",
```
andreasphil: Question: Are you sure this works? I don't think we're allowed to have two identical paths. (see `AmendingLawMetadataEditorElement`)

frontend/src/views/AmendingLawMetadataEditor.vue
```diff
@@ -206,7 +216,13 @@ const {
               v-for="element in elements"
               :key="element.eid"
               :to="{
-                name: 'AmendingLawMetadataEditorElement',
+                name:
```
andreasphil: Suggestion: this would be a good candidate to ["pre-compute"](https://vuejs.org/guide/essentials/computed.html#computed-caching-vs-methods) i.e. something like

```ts
const elementLinks = computed(() => {
  return elements.value.map(el => {
    // return the el itself + the route it should link to
  })
})
``` 

... and then iterate over `elementLinks` instead.

Advantages: It's faster and the template is a bit more tidy

andreasphil (changes requested):

frontend/src/router.ts
hamo225: I decided to revert the change, I extended the original pattern to include the new element types (book, chapter, part, subtitle, title)

frontend/e2e/amending-law-metadata-editor.spec.ts
hamo225: I decided to revert the change, I extended the original pattern to include the new element types (book, chapter, part, subtitle, title)

frontend/e2e/amending-law-metadata-editor.spec.ts
hamo225: same as above

PR #519, opened 2024-08-13
RISDEV-4460 improve rendering of element structure

malte-laukoetter:


andreasphil:
@malte-laukoetter Great work (as usual)!

I just noticed two things:

In our design the `num` is inline with the elements it belongs to:

Design

I didn't see that in the app:

App

---

And: There are some glitches where borders collapse and/or elements don't extend to the full width:

Bildschirmfoto 2024-08-14 um 09 32 05

I bet the markup can be quite tricky/complex but do you see a way to fix those? For reference I looked at [this mod](https://norms.ris.dev.ds4g.net/amending-laws/eli/bund/bgbl-1/1002/10/1002-01-10/1/deu/regelungstext-1/articles/hauptteil-1_para-1/edit/hauptteil-1_para-1_abs-1_untergl-1_listenelem-1_inhalt-1_text-1_%C3%A4ndbefehl-1) in Chrome.

malte-laukoetter:
Thanks for catching that, i actually added the special num handling but it was also butchered a bit with the transpiling problem (thats the reason for the non full width num). But it should all be fixed now ^^

malte-laukoetter:
Bug tickets for the transpiling problems: https://github.com/vuejs/core/issues/11613 & https://github.com/evanw/esbuild/issues/3877

PR #521, opened 2024-08-14
Return new elements in endpoint

andreasphil:


backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/ElementService.java
```diff
@@ -78,16 +78,16 @@ public static ElementType fromLabel(String label) throws UnsupportedElementTypeE
       // preliminary and untested due to lack of test data. Will be updated in RISDEV-4530.
```
malte-laukoetter, line 78: is this comment still current?

andreasphil: good catch, I missed that 😇 will remove it, thanks!

PR #523, opened 2024-08-15
Risdev-4559 string index out of bounds exception range 0 1 out of bounds for length 0

hannes-mk:
Return early if targetNode is empty.

Returning early may not be the best way to handle this kind of error, but it follows the pattern of other pieces of code in that function (e.g. line 146: 
`if (mod.getSecondQuotedText().isEmpty()
        || mod.getOldText().isEmpty()
        || mod.getNewText().isEmpty()) return;`

Fixes RISDEV-4559

### DIFF
--- a/backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/TimeMachineService.java
+++ b/backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/TimeMachineService.java
@@ -152,12 +152,14 @@ public class TimeMachineService implements ApplyPassiveModificationsUseCase {
     if (mod.containsRef()) {
       final Node secondQuotedTextNode = mod.getSecondQuotedText().get();
 
-      // Get the target node's current text content and remove unnecessary whitespaces, for the
+      // Get the target node's current text content and remove unnecessary
+      // whitespaces, for the
       // character counting from mod to work
       final String targetText = targetNode.getTextContent().trim().replaceAll("\\s+", " ");
       if (targetText.length() == 0) return;
 
-      // Get the start and end index of the old text, within the cleaned old text of the target node
+      // Get the start and end index of the old text, within the cleaned old text of
+      // the target node
       final int startIndex = targetText.indexOf(oldText);
       final int endIndex = startIndex + oldText.length();
 
@@ -176,7 +178,8 @@ public class TimeMachineService implements ApplyPassiveModificationsUseCase {
         targetNode.appendChild(beforeTextNode);
       }
 
-      // Import content of quotedText from amending law (which include akn:refs) using importNode
+      // Import content of quotedText from amending law (which include akn:refs) using
+      // importNode
       // because of different documents
       NodeParser.nodeListToList(secondQuotedTextNode.getChildNodes())
           .forEach(
--- a/backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/TimeMachineService.java
+++ b/backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/TimeMachineService.java
@@ -161,8 +161,8 @@ public class TimeMachineService implements ApplyPassiveModificationsUseCase {
       final int endIndex = startIndex + oldText.length();
 
       // Split the text into three parts: before, replacement, after
-      final String beforeText = targetText.substring(0, startIndex);
-      final String afterText = targetText.substring(endIndex + 1);
+      final String beforeText = targetText.length() > 0 ? targetText.substring(0, startIndex) : "";
+      final String afterText = targetText.length() > 0 ? targetText.substring(endIndex + 1) : "";
 
       // Clear the target node's content
       while (targetNode.hasChildNodes()) {
--- a/backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/TimeMachineService.java
+++ b/backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/TimeMachineService.java
@@ -156,7 +156,7 @@ public class TimeMachineService implements ApplyPassiveModificationsUseCase {
       // whitespaces, for the
       // character counting from mod to work
       final String targetText = targetNode.getTextContent().trim().replaceAll("\\s+", " ");
-      if (targetText.length() == 0) return;
+      if (targetText.isEmpty()) return;
 
       // Get the start and end index of the old text, within the cleaned old text of
       // the target node
--- a/backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/TimeMachineService.java
+++ b/backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/TimeMachineService.java
@@ -155,14 +155,15 @@ public class TimeMachineService implements ApplyPassiveModificationsUseCase {
       // Get the target node's current text content and remove unnecessary whitespaces, for the
       // character counting from mod to work
       final String targetText = targetNode.getTextContent().trim().replaceAll("\\s+", " ");
+      if (targetText.length() == 0) return;
 
       // Get the start and end index of the old text, within the cleaned old text of the target node
       final int startIndex = targetText.indexOf(oldText);
       final int endIndex = startIndex + oldText.length();
 
       // Split the text into three parts: before, replacement, after
-      final String beforeText = targetText.length() > 0 ? targetText.substring(0, startIndex) : "";
-      final String afterText = targetText.length() > 0 ? targetText.substring(endIndex + 1) : "";
+      final String beforeText = targetText.substring(0, startIndex);
+      final String afterText = targetText.substring(endIndex + 1);
 
       // Clear the target node's content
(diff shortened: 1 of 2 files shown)


hannes-mk:
@VictorDelCampo Pinging you for two topics:

1. "Returning early" as a means to handle empty targetNode texts. This seems to be a pattern in the function, but I wonder if it should be (maybe we agreed to simply ignore invalid modifications for now, maybe we actually want to throw an exception)
1. As the method that failed was private, I used reflection to still test it. This stirred some emotion and good input. Bottom line: why is it private?

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/TimeMachineService.java
```diff
@@ -155,6 +155,7 @@ private void applyQuotedText(final ModData modData, Node targetNode) {
       // Get the target node's current text content and remove unnecessary whitespaces, for the
       // character counting from mod to work
       final String targetText = targetNode.getTextContent().trim().replaceAll("\\s+", " ");
+      if (targetText.length() == 0) return;
```
reckseba: I'd prefer isEmpty()

backend/src/test/java/de/bund/digitalservice/ris/norms/application/service/TimeMachineServiceTest.java
```diff
@@ -34,24 +37,24 @@ void returnUnchangedIfNoPassiveModifications() {
...
+                                                        <akn:FRBRWork eId="meta-1_ident-1_frbrwork-1" GUID="3385defa-f0e5-4c6d-a2d4-17388afd5d51">
+                                                            <akn:FRBRnumber eId="meta-1_ident-1_frbrwork-1_frbrnumber-1" GUID="b82cc174-8fff-43bf-a434-5646de09e807" value="s593" />
+                                                            <akn:FRBRname eId="meta-1_ident-1_frbrwork-1_frbrname-1" GUID="374e5873-9c62-4e3d-9dbe-1b865ba0b327" value="BGBl. I" />
+                                                            <akn:FRBRdate eId="meta-1_ident-1_frbrwork-1_frbrdate-1" GUID="5a628f8c-65d0-4854-87cc-6fd01a2d7a9a" date="1964-08-05" name="verkuendungsfassung" />
+                                                         </akn:FRBRWork>
+                                                    </akn:identification>
+                                                  </akn:meta>
+                                               </akn:act>
+                                            </akn:akomaNtoso>
+                                            """))
```
reckseba: Why adding whitespace here?

backend/src/test/java/de/bund/digitalservice/ris/norms/application/service/TimeMachineServiceTest.java
```diff
@@ -137,18 +140,18 @@ void applyPassiveModificationsWhereTargetNodeEqualsNodeToChange() {
...
+                                                                    Angaben, die die
+                                                                    Feststellung der abgefragten Datensätze ermöglichen, sowie die abfragende Stelle zu protokollieren. Die Auswertung der
+                                                                    Protokolldaten
+                                                                    ist nach dem Stand der Technik zu gewährleisten. Die protokollierten Daten dürfen nur für Zwecke der
+                                                                    Datenschutzkontrolle, der
+                                                                    Datensicherung oder zur Sicherstellung eines ordnungsgemäßen Betriebs der Datenverarbeitungsanlage verwendet werden.
+                                                                    Die
+                                                                    Protokolldaten sind nach Ablauf von fünf Jahren zu löschen.
+                                    """);
```
reckseba: Why adding whitespace here?

backend/src/test/java/de/bund/digitalservice/ris/norms/application/service/TimeMachineServiceTest.java
```diff
@@ -286,9 +289,9 @@ void applyQuotedTextWithRefs() {
...
-                                        Kennezichen eines verbotenen Vereins oder einer Ersatzorganisation verwendet,</akn:p>
-                """);
+                            <?xml version="1.0" encoding="UTF-8"?><akn:p xmlns:akn="http://Inhaltsdaten.LegalDocML.de/1.6/" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" GUID="0ba9a471-e9ef-44c4-b5da-f69f068a4483" eId="hauptteil-1_para-20_abs-1_untergl-1_listenelem-2_inhalt-1_text-1">entgegen § 9 Absatz 1 <akn:ref GUID="514f37b3-5f75-4ee4-a110-6bad8c5a46c3" eId="hauptteil-1_para-20_abs-1_untergl-1_listenelem-2_inhalt-1_text-1_ref-1" href="eli/bund/bgbl-1/1001/1/1001-01-01/1/deu/regelungstext-1">Satz 2</akn:ref>, Absatz 2 oder 3
+                                                                    Kennezichen eines verbotenen Vereins oder einer Ersatzorganisation verwendet,</akn:p>
+                                            """);
```
reckseba: Why adding whitespace here?

backend/src/test/java/de/bund/digitalservice/ris/norms/application/service/TimeMachineServiceTest.java
```diff
@@ -311,4 +314,74 @@ void applyQuotedTextWithRefs() {
...
+      applyQuotedTextPrivateMethod.setAccessible(true);
+
+      // when
+      applyQuotedTextPrivateMethod.invoke(timeMachineService, modData, targetNode);
+
+      // then
+      assertThat(XmlMapper.toString(unchangedTargetNodeClone))
+          .isEqualTo(XmlMapper.toString(unchangedTargetNodeClone));
+    }
+  }
```
reckseba, line 337: In view of the discussion yesterday in the engineering discipline I'd suggest not to go that route and avoid testing private methods this way. How about testing the public one like we always do?

Following should do the trick:
```
// given
final var norm = NormFixtures.loadFromDisk("NormWithPassiveModifications.xml");
      norm.getNodeByEId("hauptteil-1_para-20_abs-1_untergl-1_listenelem-2_inhalt-1_text-1").get().setTextContent("");

final var amendingLaw = NormFixtures.loadFromDisk("NormWithQuotedTextModAndRefs.xml");

when(normService.loadNorm(any())).thenReturn(Optional.of(amendingLaw));

// when then
timeMachineService.applyPassiveModifications(
                      new ApplyPassiveModificationsUseCase.Query(norm, Instant.MAX));
```

hannes-mk:
@reckseba Thanks for your help! It has lead to a nice, small PR addressing exactly the issue at hand with no shady "test what you actually shouldn't".

Will merge once the pipeline succeeds.

VictorDelCampo:
> @VictorDelCampo Pinging you for two topics:
> 
> 1. "Returning early" as a means to handle empty targetNode texts. This seems to be a pattern in the function, but I wonder if it should be (maybe we agreed to simply ignore invalid modifications for now, maybe we actually want to throw an exception)
> 2. As the method that failed was private, I used reflection to still test it. This stirred some emotion and good input. Bottom line: why is it private?

Hey @hannes-mk sorry for the late reply, as you know I was out the whole last week. Answering your questions:

1. The documentalists want to work on the mods, but be able to save them without having finalize them. Meaning that the mods we save (both in the active mod as well as in the passive mods, from where we take the information for the time machine) may be incomplete. Therefore when the time machine runs, if will just ignore those incomplete mods and don't through any errors so that the flow is not stopped. We could think about logging the info but is not needed for now. Why the time machine needs to ingore those incomplete mods is because we run it every time we modify any of the fields for the mods.
2. I would not recommend testing private methods, they are private for a reason (mainly here for readability purposes). In our hexagonal architecture (Ports&Adapters) we have interfaces defining input and output ports. This service is implementing only 1 input port (usecase), meaning only that implemented method should be public, since this method represents the interfaces to be invoked by external clients (such as controllers, other services, or components from different layers). The rest of the methods in the service serve as helpers or internal operations to support that one public method. Why should they be private:
- Encapsulation: private methods hide the implementation from the "outside" world, preventing external code from relying on or misusing internal logic that might change in the future. The way a quotedText or a quotedStructure is applied is not relevant for the callers
- Interface boundary: when keeping only that method public, we create a clear boundary of what the service class is supposed to do from an external point of view. This aligns with the principles of the Ports&Adapters architecture, where the public API represents the only way to interact with the business logic (i.e., the method exposed by the input port). 
- Maintenance: if all methods were public, it would be harder to change the internal implementation without risking breaking the clients depending on those public methods.

hannes-mk:
Thanks for the extensive answer, @VictorDelCampo ! [edit: clarify]

hannes-mk:
In fact, your write-up was so good, @VictorDelCampo that I think I may now be able to nail down where I have conflicting intuitions. Will take it to the dev weekly. :pray:

PR #525, opened 2024-08-16
Add proposal for organising things by view

malte-laukoetter:
RISDEV-0000

### DIFF
--- a/doc/adr/0012-organize-feature-specific-things-by-page-in-frontend.md
+++ b/doc/adr/0012-organize-feature-specific-things-by-page-in-frontend.md
@@ -92,6 +92,8 @@ something and we need to introduce some further nesting there.
 It could also be the case that we add some larger features that are not view-specific. In that case we could still add
 some `features` folder to organize everything related to that feature.
 
+Also the urls of our application could change and with this proposale they would be tightly coupled.
+
 ## Further details
 
 <details>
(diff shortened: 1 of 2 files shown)


doc/adr/0012-organize-feature-specific-things-by-page-in-frontend.md
```diff
@@ -0,0 +1,567 @@
...
+
+At the moment all our big features have their own route[^1] in our application.
+
+## Decision
+
+We use the existing `components` and `composables` folders for shared components and composables.
+
+The `views` folder gets restructured to also contain things that are specific to the specific route of that view.
+
+The view folders are organized by the url paths of the view.
```
andreasphil: Comment: In that case, should we revisit again the decision to use folder-based routing? Last time I checked I was not super convinced of any of the existing solutions, but maybe things have changed in the meantime.

andreasphil (approved):
I like the proposal! We can implement that if the others agree too.

doc/adr/0012-organize-feature-specific-things-by-page-in-frontend.md
malte-laukoetter: I'm was also not convinced by them last time I checked, especially if the path parameters get a bit more complex. And with the elis in the paths I do not think that it would work for us at all.

andreasphil: Didn't think about the ELIs. Agree!

VictorDelCampo (commented):
I just wanted to comment that for somebody like me rarely working in the frontend I would really appreciate having the structure reflects the pages/urls we have so that I don't need to swim panicky in the big frontend lake.

Also wanted to add, thinking of the architecture tests we have in the backend, if you plan something similar, if possible, to disallow certain views/components using code they should not be using

andreasphil:
@malte-laukoetter Can we merge this? Doesn't seem like there are any concerns.

PR #526, opened 2024-08-16
Add ADR proposal for error responses

malte-laukoetter:
RISDEV-4013

### DIFF
--- a/doc/adr/0013-error-responses.md
+++ b/doc/adr/0013-error-responses.md
@@ -52,19 +52,15 @@ a `norm-not-found` error. In this case only the `norm-not-found` error will be s
 The response includes an additional `errors` array. The elements of the `errors` array can have the same fields as the
 global object. If no `type` is provided it is assumed that it is the same as the parent type.
 
-For both JSON and LegalDocML.de some additional fields are defined to reference parts of a JSON-object or LDML.de
-document:
+### Extension Members (additional fields)
 
-#### JSON
+We expect certain extension members to be needed by many errors. Therefore, here are some predefined extension members
+that can be used when needed and may not be used for other purposes:
 
 - `pointer`: a JSON-Pointer ([RFC-6901](https://www.rfc-editor.org/info/rfc6901)) referencing the field of the JSON
   object responsible for this error
-
-#### LegalDocML.de
-
-- `eid`: an eId (LegalDocML.de) referencing the part of the norm responsible for the error (preferred)
-- `guid`: a GUID (LegalDocML.de) referencing the part of the norm responsible for the error (optionally or as
-  alternative to the eid)
+- `eid`: an eId (LegalDocML.de) referencing the part of the norm responsible for the error
+- `guid`: a GUID (LegalDocML.de) referencing the part of the norm responsible for the error
 
 ### Displaying messages for the user
 
--- a/doc/adr/0013-error-responses.md
+++ b/doc/adr/0013-error-responses.md
@@ -75,7 +75,7 @@ Example error response for NormNotFound:
 {
  "type": "/errors/norm-not-found",
  "title": "Norm not found.",
- "detail": "The requested norm could not be found.",
+ "details": "The requested norm (eli/bund/bgbl-1/2017/s419/2017-03-15/1/deu/regelungstext-1) could not be found.",
  "instance": "/eli/bund/bgbl-1/2017/s419/2017-03-15/1/deu/regelungstext-1"
 }
 ```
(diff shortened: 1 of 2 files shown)


doc/adr/0013-error-responses.md
```diff
@@ -0,0 +1,115 @@
...
+
+[RFC-9457 (Problem Details for HTTP APIs)](https://www.rfc-editor.org/rfc/rfc9457) addresses how such information can be
+sent by an API.
+
+## Decision
+
+We follow [RFC-9457 (Problem Details for HTTP APIs)](https://www.rfc-editor.org/rfc/rfc9457) for our error responses.
+All error responses are JSON objects.
+
+The `type` field always starts with `/errors/`. Therefore, all errors are relative to our application.
```
hannes-mk: maybe "relative URLs"

doc/adr/0013-error-responses.md
```diff
@@ -0,0 +1,115 @@
...
+
+For errors related to a specific norm the `instance` field is filled with the eli of the norm starting with `/eli/`.
+
+Both `title` and `details` are written in english.
+
```
hannes-mk: Maybe add what is required and what is optional.

doc/adr/0013-error-responses.md
```diff
@@ -0,0 +1,115 @@
...
+### Multiple error messages in one response
+
+For errors that can have multiple messages (e.g. after a validation) we mostly follow the example from Section 3 of
+RFC-9457 for `HTTP/1.1 422 Unprocessable Content`.
+
+The multiple errors MUST be of the same general type. Otherwise, only the most severe error is reported. E.g. it is
+possible to send multiple `norm-not-valid` errors but not a `norm-not-valid` error at the same time as
+a `norm-not-found` error.
+
```
hannes-mk: so, in this case, only `norm-not-found` is returned

doc/adr/0013-error-responses.md
```diff
@@ -0,0 +1,115 @@
...
+#### LegalDocML.de
+
+- `eid`: an eId (LegalDocML.de) referencing the part of the norm responsible for the error (preferred)
+- `guid`: a GUID (LegalDocML.de) referencing the part of the norm responsible for the error (optionally or as
+  alternative to the eid)
+
+### Displaying messages for the user
+
+To create messages to display to the user the `type` must be used in combination with the extension members. The `title`
```
hannes-mk: "extension members": maybe "other fields"

doc/adr/0013-error-responses.md
```diff
@@ -0,0 +1,115 @@
...
+RFC-9457 for `HTTP/1.1 422 Unprocessable Content`.
+
+The multiple errors MUST be of the same general type. Otherwise, only the most severe error is reported. E.g. it is
+possible to send multiple `norm-not-valid` errors but not a `norm-not-valid` error at the same time as
+a `norm-not-found` error.
+
+The response includes an additional `errors` array. The elements of the `errors` array can have the same fields as the
+global object. If no `type` is provided it is assumed that it is the same as the parent type.
+
+For both JSON and LegalDocML.de some additional fields are defined to reference parts of a JSON-object or LDML.de
```
VictorDelCampo: I wasn't sure what you meant here with "JSON" but now I understand you mean the request payload. All these fields "pointer", "eid" and "guid" are extension members we already have in mind that we will be using.

I would maybe rephrase that part and maybe say something like possible/probable extension members, and then you name them and explain for what would they be used (specially JSON being the request payload :) )

doc/adr/0013-error-responses.md
VictorDelCampo: I guess the frontend would then use the `type` which will probably contain parameters matching the extension members to build the message with those right?

doc/adr/0013-error-responses.md
```diff
@@ -0,0 +1,115 @@
...
+
+Example error response for validation:
+
+```
+{
+ "type": "/errors/norm-not-valid",
+ "title": "Norm validation failed.",
+ "instance": "/eli/bund/bgbl-1/2017/s419/2017-03-15/1/deu/regelungstext-1"
+ "errors": [{
+    "type": "/errors/norm-not-valid/quoted-structure/target-and-up-to-are-not-siblings
```
VictorDelCampo: I guess you left out the `title` in the sub-error because you were just doing an example? This brings me back to the point from @hannes-mk saying to specify which properties would be required (like `type`) and which can be optional (`title`, `details`, `instance` and extension members)?

VictorDelCampo (commented):
Nice work Malte. I like this concept so much! It is going some steps further than what was proposed in the caselaw team. I just have some minor comments you could use to extend the ADR if you feel like to

doc/adr/0013-error-responses.md
malte-laukoetter: yes exactly thats the idea

doc/adr/0013-error-responses.md
malte-laukoetter: i moved this to its own section, so i hope it is clearer now?

doc/adr/0013-error-responses.md
malte-laukoetter: I adjusted the examples to better align with the text and added some sentences about which fields are required and recommended :)

VictorDelCampo:
@malte-laukoetter looks great! Just adapt the date and can be merged :)

PR #541, opened 2024-08-22
RISDEV-4486 character counting

malte-laukoetter:


backend/src/main/java/de/bund/digitalservice/ris/norms/utils/XmlMapper.java
```diff
@@ -27,22 +28,44 @@ public class XmlMapper {
...
    */
   public static Document toDocument(String xmlText) {
+    return toDocument(xmlText, false);
+  }
+
+  /**
+   * Maps a string containing xml to a {@link Document}.
+   *
+   * @param xmlText The input string containing xml to be mapped to a {@link Document}
+   * @param validateSchema Should the schmea of the XML be validated to be valid LDML.de? When this
```
VictorDelCampo, line 45: Small typo: schmea --> schema

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/CharacterRange.java
```diff
@@ -54,6 +59,121 @@ public boolean isValidCharacterRange() {
...
+   *
+   * <p>SIDE-EFFECT: This might split text-nodes into multiple text nodes to be able to return
+   * exactly the range. Please call Node::normalize after you are finished working with the nodes.
+   *
+   * @param node the node in which the range should be found
+   * @return a list of nodes that represent the given range
+   */
+  public List<Node> getNodesInRange(Node node) {
+    if (node.getTextContent().length() < getStart()) {
+      throw new IndexOutOfBoundsException();
```
VictorDelCampo: It would be nice if we would use the other constructor of that exception to pass a message giving specific information like what was the length of the node and what was the given start?

backend/src/main/java/de/bund/digitalservice/ris/norms/utils/XmlMapper.java
VictorDelCampo: Small typo: schmea --> schema

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/CharacterRange.java
```diff
@@ -54,6 +59,121 @@ public boolean isValidCharacterRange() {
...
+    }
+
+    if (node.getTextContent().length() < getEnd()) {
+      throw new IndexOutOfBoundsException();
```
VictorDelCampo: Same here as the comment before

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/CharacterRange.java
```diff
@@ -54,6 +59,121 @@ public boolean isValidCharacterRange() {
...
+  /**
+   * Finds the nodes that make up the given range within the given node.
+   *
+   * <p>SIDE-EFFECT: This might split text-nodes into multiple text nodes to be able to return
+   * exactly the range. Please call Node::normalize after you are finished working with the nodes.
+   *
+   * @param node the node in which the range should be found
+   * @return a list of nodes that represent the given range
+   */
+  public List<Node> getNodesInRange(Node node) {
```
VictorDelCampo, line 127: Does this method need to be public? As far as I can see the entry point for this functionality is the other public method `findTextInNode`?

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/CharacterRange.java
```diff
@@ -54,6 +59,121 @@ public boolean isValidCharacterRange() {
...
+   * range from start to end of the text content.
+   *
+   * <p>A call to Node::normalize undoes this.
+   *
+   * @param textNode A text node
+   * @param start start of the text range within the node
+   * @param end end (excluding) of the text range within the node
+   * @return a text node that represents exactly the range
+   */
+  private Node extractTextNodeForRange(Node textNode, int start, int end) {
```
VictorDelCampo, line 86: Why manipulate the passed textNode if the goal of the method is to extract and return only the text node for the given range? The passed node will be manipulated and splitted into several text nodes and this will modify the XML right?

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/CharacterRange.java
```diff
@@ -54,6 +59,121 @@ public boolean isValidCharacterRange() {
...
+
+          start -= textContentLength;
+          end -= textContentLength;
+
+          if (end <= 0) {
+            return nodesInRange;
+          }
+        }
+
+        throw new IndexOutOfBoundsException();
```
VictorDelCampo: Also here pass a message

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/CharacterRange.java
```diff
@@ -54,6 +59,121 @@ public boolean isValidCharacterRange() {
...
+
+    textNode.getParentNode().removeChild(textNode);
+
+    return rangeNode;
+  }
+
+  /**
+   * Finds the nodes that make up the given range within the given node.
+   *
+   * <p>SIDE-EFFECT: This might split text-nodes into multiple text nodes to be able to return
```
VictorDelCampo, line 121: This info would not be needed if you make the method private and it is only called by the other public one, which is always doing the normalize right? Maybe add the info about normalizing in that other public method?

VictorDelCampo (changes requested):
Looks good Malte. Just some comments to optimize it

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/CharacterRange.java
malte-laukoetter: Yes at the moment only that method uses this. But my idea is to use this method to figure out which parts of the xml need to be changed when applying a quoted text change command. Similar to this example from the test for this method: https://github.com/digitalservicebund/ris-norms/blob/35afbf26d92ffcde5c7b314732b2044383755be6/backend/src/test/java/de/bund/digitalservice/ris/norms/domain/entity/CharacterRangeTest.java#L227-L231. And for that this method needs then needs to be public.

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/CharacterRange.java
malte-laukoetter: Yes this does modify the original XML, but that is actually the idea. With this we can then take the text node returned from this method and eg. replace it with a new node. Which is just the thing we need for applying change commands. Also the side effect here is quite minor as it can be reversed by normalizing (Node::normalize) the xml.

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/CharacterRange.java
malte-laukoetter: see the other comments

VictorDelCampo (approved):

PR #544, opened 2024-08-23
RISDEV-4486 support existing elements in amended text

malte-laukoetter:


malte-laukoetter:
This includes a complete rewrite of the EIdConsistencyGuardian, the previous algorithm was not 100% what was written in the spec and only correct for the `akn:meta` area. So to use it on the whole norm i updated it according to the spec. As many of our sample norms did not 100% adhere to the standard i needed to adjust some things in many of them.
So overall this got more complex and larger than i hoped for. But now the EIdConsistenceGuardian can take any norm and fix all eIds in it, just call it after changing something (like adding a new node) and every element will get the correct eId.

backend/src/main/java/de/bund/digitalservice/ris/norms/utils/NodeParser.java
```diff
@@ -36,9 +37,15 @@ public static Optional<String> getValueFromExpression(String xPathExpression, No
         return Optional.empty();
       }
 
+      // Clone the node if the xpath is relative to the node. This drastically improves the
```
VictorDelCampo, line 40: This sounds very interesting... didn't read much into the docs but is it related to the clone not having a parent, meaning it must be cut-off from the whole document and the node becomes like a standalone mini-document with itself as the root? 

Crazy optimization Malte! Top! Could you maybe add this extra information as a comment?

backend/src/main/java/de/bund/digitalservice/ris/norms/utils/NodeParser.java
```diff
@@ -36,9 +37,15 @@ public static Optional<String> getValueFromExpression(String xPathExpression, No
...
+      // performance.
+      if (xPathExpression.startsWith("./")) {
+        sourceNode = sourceNode.cloneNode(true);
+      }
+
       // should be invoked on every method call since: An XPath object is not thread-safe and not
       // reentrant.
-      final XPath xPath = XPathFactory.newInstance().newXPath();
+      final XPath xPath = xPathFactory.newXPath();
```
VictorDelCampo, line 48: Could you also add the same `xPathFactory` instance to the other methods? I guess only the `XPath` object in not thread-safe and not the factory instance correct?

VictorDelCampo (changes requested):
As mentioned on Slack, only reviewed the NodeParser... tomorrow more!

backend/src/main/java/de/bund/digitalservice/ris/norms/utils/NodeParser.java
malte-laukoetter: Thats probably the reason why, as the clone has no parent the document the xpath is applied to is a lot smaller. I'll extend the comment a bit :) Also not my own idea was the first suggestions from stack overflow but it did provide a really significant speedup.

backend/src/main/java/de/bund/digitalservice/ris/norms/utils/NodeParser.java
malte-laukoetter: Oh yes I'll definitely do that. And yes only the XPath itself is not thread safe (and can be created in a very small fraction of the time it takes to create a factory).

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/EIdPartType.java
```diff
@@ -0,0 +1,216 @@
...
+import lombok.Getter;
+import org.w3c.dom.Node;
+
+/**
+ * The type of part of an EId.
+ *
+ * <p>See LDML.de 1.6 Section 7.1.3
+ */
+@Getter
+public enum EIdPartType {
```
VictorDelCampo, line 16: This is amazing. Did you double check the same chapter of the LDML.de 1.7? Are they any changes there? Do we need to adapt this class when also supporting 1.7?

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/EIdPartType.java
```diff
@@ -0,0 +1,216 @@
...
+    if (aknElement.getNodeName().equals("akn:article")) {
+      return Optional.of(forAknArticle(aknElement));
+    }
+
+    return Arrays.stream(EIdPartType.values())
+        .filter(partType -> partType.aknElements.contains(aknElement.getNodeName()))
+        .findFirst();
+  }
+
+  private static EIdPartType forAknArticle(Node aknArticleElement) {
```
VictorDelCampo, line 176: According to the specification the differentiation of which eId should be also for an akn:article depends solely on the `refersTo` attribute right? You did much more steps, i guess reading through the specification, but that are not needed right? If we don't have a refersTo attribute or if the value of the attribute does not help us then we should just through an exception right? Or did you implement all the further steps because of the test data we have?

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/EId.java
```diff
@@ -38,7 +38,23 @@ public String toString() {
    * @return the eId of the node or empty if no eId could be found.
    */
   public static Optional<EId> fromNode(Node node) {
-    return NodeParser.getValueFromExpression("./@eId", node).map(EId::new);
+    if (!node.hasAttributes()) {
```
VictorDelCampo, line 41: Did you change this implementation because the class Eid itself does not support instantiating an object of it with an empty string? If so should we adapt the Eid class? That just returns a null if no eid was found?

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/EId.java
```diff
@@ -50,4 +66,59 @@ public static Optional<EId> fromNode(Node node) {
   public static EId fromMandatoryNode(Node node) {
     return new EId(NodeParser.getValueFromMandatoryNodeFromExpression("./@eId", node));
   }
+
+  /**
+   * Creates the expected EId for the given node given that the parent nodes eId is correct.
```
VictorDelCampo: Type? ...parent node... instead of ...parent nodes... ? You just used the direct parent node right?

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/EId.java
```diff
@@ -50,4 +66,59 @@ public static Optional<EId> fromNode(Node node) {
...
+   *
+   * @param node the node for which the eId should be calculated
+   * @return the expected EId for that node
+   */
+  public static Optional<EId> forNode(Node node) {
+    final Optional<EId> parentEId = EId.fromNode(node.getParentNode());
+
+    final Optional<EIdPartType> eIdPartType = EIdPartType.forAknElement(node);
+
+    if (eIdPartType.isEmpty()) {
```
VictorDelCampo, line 83: Should you also just directly return an Optional.empty() if the parentEid is also empty?

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/EId.java
```diff
@@ -50,4 +66,59 @@ public static Optional<EId> fromNode(Node node) {
...
+        .or(() -> Optional.of(new EId(newEIdPart.value())));
+  }
+
+  private static String findEIdPosition(Node node, EIdPartType eIdPartType) {
+    if (node.getNodeName().equals("akn:li")
+        && node.getParentNode().getNodeName().equals("akn:ol")) {
+      return node.getAttributes().getNamedItem("value").getNodeValue().replace(".", "");
+    }
+
+    return NodeParser.getValueFromExpression("./num/marker", node)
```
VictorDelCampo: Shouldn't we be taking the eid of the marker always from the @name attribute? As far as I know the text content of the akn:marker also contains characters like ".", ")" and so on right?

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/EId.java
```diff
@@ -50,4 +66,59 @@ public static Optional<EId> fromNode(Node node) {
...
+        .or(() -> NodeParser.getValueFromExpression("./num/marker/@name", node))
+        .orElseGet(
+            () -> {
+              var position = 1;
+              var previousSibling = node.getPreviousSibling();
+              while (previousSibling != null) {
+                var eId = EId.fromNode(previousSibling);
+
+                if (eId.isPresent()) {
```
VictorDelCampo, line 109: Should we throw an exception here if we have a previous sibling but that sibling does not have an eId?

backend/src/main/java/de/bund/digitalservice/ris/norms/utils/EidConsistencyGuardian.java
```diff
@@ -50,28 +48,12 @@ public static Document correctEids(final Document currentXml) {
           "//lifecycle/eventRef",
           "eId",
           rootElement);
-      // Check eIds
-      updateElementEids(startNode, startNode.getAttribute("eId"), rootElement);
+      // Check and fix eIds
+      updateElementEids(rootElement, rootElement);
```
VictorDelCampo: Why do you need to pass the rootElement twice? i thought maybe the updateElementEids() would be call recursively but is not the case right?

backend/src/main/java/de/bund/digitalservice/ris/norms/utils/EidConsistencyGuardian.java
```diff
@@ -108,127 +90,56 @@ private static boolean isReferenceValid(
...
+    List<Node> nodesToUpdate = new ArrayList<>();
+    nodesToUpdate.add(node);
 
-  private static String determineCorrectEid(int currentIndex, String oldEid, String newParentEid) {
-    final StringBuilder newEid = new StringBuilder();
+    while (!nodesToUpdate.isEmpty()) {
+      var currentNode = nodesToUpdate.removeLast();
 
-    String currentElementId;
+      if (currentNode instanceof Element element) {
```
VictorDelCampo: I think testing with instanceOf Element is the same as testing if the node type is of element type?

backend/src/main/java/de/bund/digitalservice/ris/norms/utils/EidConsistencyGuardian.java
```diff
@@ -108,127 +90,56 @@ private static boolean isReferenceValid(
...
+        final var expectedEId = EId.forNode(currentNode);
+        final var currentEId = EId.fromNode(currentNode);
 
-    // Check first if eId contains parent eIds
-    if (oldEid.contains("_")) {
-      // 1. ParentEid is not the same
-      final int lastIndex = oldEid.lastIndexOf('_');
+        if (expectedEId != currentEId) {
+          if (expectedEId.isPresent()) {
```
VictorDelCampo, line 106: What do we do if the expectedEid is not present? Are we throwing an exception in the Eid class if the expectedEid could not be calculated? And also what about is the currentEid is empty? Also throw an exception? Because we could have the case that neither of them are present and then the resulted XML would not be valid right?

backend/src/main/java/de/bund/digitalservice/ris/norms/utils/EidConsistencyGuardian.java
```diff
@@ -108,127 +90,56 @@ private static boolean isReferenceValid(
...
-    final String elementEidNoIndex = currentElementId.substring(0, currentElementId.indexOf("-"));
-    final String currentElementIndex =
-        currentElementId.substring(currentElementId.indexOf("-") + 1);
-    if (Integer.parseInt(currentElementIndex) != (currentIndex)) {
-      newEid.append(elementEidNoIndex);
-      newEid.append("-");
-      newEid.append(currentIndex);
-    } else {
-      newEid.append(currentElementId);
+        nodesToUpdate.addAll(NodeParser.nodeListToList(currentNode.getChildNodes()).reversed());
```
VictorDelCampo: Why reversing the child nodes? isn't the result of getChildNodes() already in the order meaning the first one returned is the first child of the node?

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/TimeMachineService.java
```diff
@@ -117,88 +116,84 @@ public Norm applyPassiveModifications(ApplyPassiveModificationsUseCase.Query que
...
-
-      // Import afterText
-      if (StringUtils.isNotEmpty(afterText)) {
-        final Text afterTextNode = targetNode.getOwnerDocument().createTextNode(afterText);
-        targetNode.appendChild(afterTextNode);
+    if (modData.mod().getSecondQuotedText().isPresent()) {
+      try {
+        applyQuotedText(modData, targetNode.get());
+      } catch (IllegalArgumentException | IndexOutOfBoundsException exception) {
+        log.info(
```
VictorDelCampo, line 146: Shouldn't this be on error level? Or is it info level because the time machine runs always even if the mod is not complete/correct?

VictorDelCampo (changes requested):
It is an awesome implementation Malte, specially moving the Eid related stuff (determine the correct eId of an element using the parent and the EidPartType). Thanks for also adapting our wished-to-be-better quality of the test data

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/EIdPartType.java
malte-laukoetter: The spec isn't that clear which `refersTo` attributes are for a stammform and which are for a mantelform and `article`s within `quotedStrucuture`s actually do not need this attribute at all. So I looked into the Schematron rules and they depended upon the information from `meta:form`. (and are especially complicated for things like `vertragsgesetz`) So I decided to just use the same algorithm the rules use.

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/EId.java
malte-laukoetter: No, its just a lot faster to do it this way and I had some performance problems as i this method is called very often.

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/EIdPartType.java
malte-laukoetter: Oh no i didn't check 1.7 yet. I think I should do that especially as we are already using `rref`. But at general we will need to adjust this when supporting new ldml versions

VictorDelCampo: Yes the topic of rref was maybe to premature... Should we maybe keep track of all the features/parts in the code-base we should double check when supporting several LDML.de versions? Maybe a placeholder ticket?

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/EId.java
malte-laukoetter: In our current samples the marker has no text content, so the `@name` attribute is always used. This is different from the samples from the spec which does not include the `@name` attribute. But as i just looked at it again this is just some non adjusted example from an older version of ldml-de so I'll remove that line ^^

backend/src/main/java/de/bund/digitalservice/ris/norms/utils/EidConsistencyGuardian.java
malte-laukoetter: Yes, it just also convinces java that it is indeed an `Element`

backend/src/main/java/de/bund/digitalservice/ris/norms/utils/EidConsistencyGuardian.java
malte-laukoetter: There are some elements in the xml that do not need an eid so both being empty is totally fine. I was wondering what to do if we only have a current eid but no expected one and decided that it is best to keep the eid in that case to not delete eids from elements we might not support yet

backend/src/main/java/de/bund/digitalservice/ris/norms/utils/EidConsistencyGuardian.java
malte-laukoetter: Yes but (for some reason i dont remember) we iterate in a depth-first way (and therefore use a last-in-first-out strategy) so the first child needs to be added last

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/EId.java
malte-laukoetter: i dont think so as it could be fine that a previous sibling has no eid if it is an element that does not need an eid

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/EId.java
malte-laukoetter: no, see other comments about possibly empty eids

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/TimeMachineService.java
malte-laukoetter: As this is something we expect to happen regularly (for the reasons you mentioned) i was thinking that info is fine

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/EId.java
VictorDelCampo: You are right. And also the way the code is written, if that previous sibling didn't have an eId if would have gotten one in the loop run before, right? You can resolve the conversation

backend/src/main/java/de/bund/digitalservice/ris/norms/utils/EidConsistencyGuardian.java
VictorDelCampo: Alright. This means that we may have nodes without a corrected eId or even have some with no eId at all altough they should have one right?

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/EIdPartType.java
malte-laukoetter: I think thats a good idea, i created a template ticket here: https://digitalservicebund.atlassian.net/browse/RISDEV-4698

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/EId.java
malte-laukoetter: yes if it is used in the EIdConsistencyGuard it would have already gotten one

backend/src/main/java/de/bund/digitalservice/ris/norms/utils/EidConsistencyGuardian.java
VictorDelCampo: Oh i see you take the last element from `nodesToUpdate` when iterating. Alright! You could also just don't do the reversed() and take the first element from `nodesToUpdate` right?

backend/src/main/java/de/bund/digitalservice/ris/norms/utils/EidConsistencyGuardian.java
malte-laukoetter: No. All nodes that should have an eId will get the corrected one. Only the once that, according to `Eid::forNode`, should not have an eId will keep their eId if they already have one.

backend/src/main/java/de/bund/digitalservice/ris/norms/utils/EidConsistencyGuardian.java
malte-laukoetter: yes, i couldn't find a reason against doing it breath-first instead of depth-first so i changed that

backend/src/main/java/de/bund/digitalservice/ris/norms/utils/EidConsistencyGuardian.java
VictorDelCampo: Got it. Even if a parent node didn't have an eId it will be corrected in the iteration before. Thanks!

PR #550, opened 2024-08-26
Improve regex toolchain docu

reckseba:
Does this help?


backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/ReferenceService.java
```diff
@@ -127,6 +127,8 @@ private List<MatchInfo> findReferences(final String text) {
 
   record MatchInfo(String reference, int start, int end) {}
 
+  // This regex was crafted by a regex tool. Check /regex/README.md before editing. We need to keep
+  // both in sync.
```
hannes-mk: What does "both" refer to?

regex/README.md
```diff
@@ -1,14 +1,35 @@
-# regex toolchain
+# Regular Expression Toolchain for Reference Detection
+## Motivation and Context
+When displaying an amending law, potential references (AV) should be highlighted and available for interaction. Documentalists could highlight them manually, but we'd like to support them with a semi-automated reference detection.
```
hannes-mk: what's "AV"?

hannes-mk:
Maybe not for the documentation, but for me: Where is that regex used, eventually? 
In the frontend or backend? If only one of them: why is the "regex" machinery outside that folder?

hannes-mk:
Another thought: the regex is not tested, is it?

hannes-mk (approved):
I think your information is already better that what we have right now.

My other comments are meant as input to potentially making it even more informative.

regex/README.md
reckseba: Aktivverweis. But yes, that comes out of nowhere

reckseba:
> Maybe not for the documentation, but for me: Where is that regex used, eventually? In the frontend or backend? If only one of them: why is the "regex" machinery outside that folder?

In the backend in the ReferenceService. I thought it might be obvious if I reference the service like that. I can extend it to make it more clear.

> Another thought: the regex is not tested, is it?

Whats a tested regex for you? In a way it's tested as every text file represents a bucket which is tested against the according one.

hannes-mk:
I'll take the testing question to our dev weekly.

regex/README.md
```diff
@@ -16,7 +16,7 @@ When a certain case shall be captured, that is not yet handled, the idea is:
 The bitter pill of the current approach is, that one needs to check if the regex in Java code was altered in the meantime. Meaning if someone touched the regex there without changing it here, the next time a rendered regex will be inserted there, the changes will be overridden.
```
hannes-mk:
```suggestion
The bitter pill of the current approach is, that one needs to check if the regex in Java code was altered in the meantime. Meaning if someone touched the regex there without changing it here, the next time a rendered regex will be inserted there, the changes will be overwritten.
```

PR #551, opened 2024-08-26
Breaking regex on purpose

hannes-mk:
Let's see whether the pipeline does pick it up.

RISDEV-0000


reckseba:
Oh that's what you mean. Nah the regex is not tested in that way. Pipe failed because of trivy scan.

PR #554, opened 2024-08-27
Refactor article controller

andreasphil:
RISDEV-4322

### DIFF
--- a/backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/controller/ArticleController.java
+++ b/backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/controller/ArticleController.java
@@ -244,10 +244,8 @@ public class ArticleController {
       produces = {TEXT_HTML_VALUE})
   public ResponseEntity<String> getArticleRender(
       final Eli eli, @PathVariable final String eid, @RequestParam Optional<Instant> atIsoDate) {
-    return loadArticleHtmlUseCase
-        .loadArticleHtml(
-            new LoadArticleHtmlUseCase.Query(eli.getValue(), eid, atIsoDate.orElse(null)))
-        .map(ResponseEntity::ok)
-        .orElse(ResponseEntity.notFound().build());
+    return ResponseEntity.ok(
+        loadArticleHtmlUseCase.loadArticleHtml(
+            new LoadArticleHtmlUseCase.Query(eli.getValue(), eid, atIsoDate.orElse(null))));
   }
 }
--- a/backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/controller/ArticleController.java
+++ b/backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/controller/ArticleController.java
@@ -244,8 +244,8 @@ public class ArticleController {
       produces = {TEXT_HTML_VALUE})
   public ResponseEntity<String> getArticleRender(
       final Eli eli, @PathVariable final String eid, @RequestParam Optional<Instant> atIsoDate) {
-    return ResponseEntity.ok(
-        loadArticleHtmlUseCase.loadArticleHtml(
-            new LoadArticleHtmlUseCase.Query(eli.getValue(), eid, atIsoDate.orElse(null))));
+    var query = new LoadArticleHtmlUseCase.Query(eli.getValue(), eid, atIsoDate.orElse(null));
+    var articleHtml = loadArticleHtmlUseCase.loadArticleHtml(query);
+    return ResponseEntity.ok(articleHtml);
   }
 }
--- a/backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/controller/ArticleController.java
+++ b/backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/controller/ArticleController.java
@@ -5,12 +5,10 @@ import static org.springframework.http.MediaType.*;
 import de.bund.digitalservice.ris.norms.adapter.input.restapi.mapper.ArticleResponseMapper;
 import de.bund.digitalservice.ris.norms.adapter.input.restapi.schema.ArticleResponseSchema;
 import de.bund.digitalservice.ris.norms.application.exception.ArticleNotFoundException;
-import de.bund.digitalservice.ris.norms.application.exception.NormNotFoundException;
 import de.bund.digitalservice.ris.norms.application.port.input.*;
 import de.bund.digitalservice.ris.norms.domain.entity.Analysis;
 import de.bund.digitalservice.ris.norms.domain.entity.Eli;
 import de.bund.digitalservice.ris.norms.domain.entity.Href;
-import de.bund.digitalservice.ris.norms.domain.entity.Norm;
 import de.bund.digitalservice.ris.norms.domain.entity.TextualMod;
 import java.time.Instant;
 import java.util.Collections;
@@ -69,15 +67,17 @@ public class ArticleController {
       final Eli eli,
       @RequestParam final Optional<String> amendedBy,
       @RequestParam final Optional<String> amendedAt) {
-    final var optionalNorm = loadNormUseCase.loadNorm(new LoadNormUseCase.Query(eli.getValue()));
+    final var norm = loadNormUseCase.loadNorm(new LoadNormUseCase.Query(eli.getValue()));
 
     // amendedAt and amendedBy refer to passive modifications (i.e. amended at a specific
     // date or by a specific change law). So we collect a list of all passive modifications
     // matching both criteria.
 
     var passiveModificationsAmendedAtOrBy =
-        optionalNorm
-            .flatMap(n -> n.getMeta().getAnalysis().map(Analysis::getPassiveModifications))
+        norm
+            .getMeta()
+            .getAnalysis()
+            .map(Analysis::getPassiveModifications)
             .orElse(Collections.emptyList())
             .stream()
             .filter(
@@ -93,17 +93,14 @@ public class ArticleController {
                 passiveModification -> {
                   if (amendedAt.isEmpty()) return true;
                   else
-                    return optionalNorm
-                        .get()
-                        .getStartEventRefForTemporalGroup(
+                    return norm.getStartEventRefForTemporalGroup(
                             passiveModification.getForcePeriodEid().orElseThrow())
                         .equals(amendedAt);
                 })
             .toList();
 
     return ResponseEntity.ok(
-        optionalNorm.map(Norm::getArticles).stream()
-            .flatMap(List::stream)
+        norm.getArticles().stream()
             .filter(
                 article -> {
                   // If we don't filter by anything related to passive modifications,
@@ -130,18 +127,20 @@ public class ArticleController {
                 })
             .map(
                 article -> {
-                  var targetLawZf0 =
+                  final var targetLaw =
                       article
                           .getAffectedDocumentEli()
-                          .flatMap(
+                          .map(
                               eliTargetLaw ->
                                   loadNormUseCase.loadNorm(new LoadNormUseCase.Query(eliTargetLaw)))
-                          .map(
-                              targetLaw ->
-                                  loadZf0UseCase.loadOrCreateZf0(
-                                      new LoadZf0UseCase.Query(
-                                          optionalNorm.get(), targetLaw, true)))
                           .orElse(null);
+
+                  final var targetLawZf0 =
+                      targetLaw != null
+                          ? loadZf0UseCase.loadOrCreateZf0(
+                              new LoadZf0UseCase.Query(norm, targetLaw, true))
+                          : null;
+
                   return ArticleResponseMapper.fromNormArticle(article, targetLawZf0);
                 })
             .toList());
@@ -202,10 +201,7 @@ public class ArticleController {
       final Eli eli, @PathVariable final String eid) {
     final var eliValue = eli.getValue();
 
-    final var norm =
-        loadNormUseCase
-            .loadNorm(new LoadNormUseCase.Query(eliValue))
-            .orElseThrow(() -> new NormNotFoundException(eliValue));
+    final var norm = loadNormUseCase.loadNorm(new LoadNormUseCase.Query(eliValue));
 
     final var foundArticle =
         norm.getArticles().stream()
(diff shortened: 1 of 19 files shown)


backend/src/main/java/de/bund/digitalservice/ris/norms/application/port/input/LoadArticleHtmlUseCase.java
```diff
@@ -1,12 +1,21 @@
...
-  Optional<String> loadArticleHtml(Query query);
+  /**
+   * Loads the HTML of an article from a norm.
+   *
+   * @param query The article to load
+   * @return HTML string of the article
+   * @throws NormNotFoundException If the norm is not found
+   * @throws ArticleNotFoundException If the norm exists but does not contain the article
+   */
+  String loadArticleHtml(Query query) throws ArticleNotFoundException, NormNotFoundException;
```
andreasphil, line 18: @VictorDelCampo I noticed that not all of our use cases declare that they're throwing exceptions. I added the `throws` here because I figured it would be useful to know how loading the article HTML can go wrong but not sure what the typical approach here is. Same for the other use cases.

VictorDelCampo: In java you have two types of exceptions, 

1. checked exceptions that represent errors outside the control of the program and that a well-written application should anticipate and recover from (like for example file not found `FileNotFoundException`), meaning if you write code using a method from a library that may throw that checked exception you should handle that case in your code. These exceptions are checked at compile-time, meaning if you don't handle them you will get an error.
2. runtime exceptions that are used to indicate programming errors (like the famous `NullPointerException` or `ArrayIndexOutOfBoundsException`). Java does not verify these unchecked exceptions at compile-time, meaning you don't need to handle them and therefore are not needed in the `throws` declaration of a method.


The general praxis is that runtime exceptions (the ones we have) are not declared in the `throws` definition but can be added, specially for readability of the code and giving the caller some extra information. Also we would give in such a way the intention, telling the caller that those runtime exceptions are a intentional part of the method's behavior and not an unexpected bug. Also declaring the runtime exceptions could be used as defensive programming, to encourage the caller to consider certain edge cases and validate inputs before calling the method.

But on the other side, declaring all exceptions may clutter the method signature and reduce readability. And only if we expect the caller to handle those possible exceptions we should declare them in the signature. IMO we should handle those in the services and the caller (controllers) don't care about what happen, they just expect, for example a Norm object

It should be IMO a team decision and probably a topic for the dev weekly?

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/NormService.java
```diff
@@ -90,20 +92,28 @@ public Optional<String> updateNormXml(UpdateNormXmlUseCase.Query query)
   }
 
   @Override
-  public List<String> loadSpecificArticles(LoadSpecificArticleXmlFromNormUseCase.Query query) {
+  public List<String> loadSpecificArticlesXmlFromNorm(
```
VictorDelCampo, line 95: In this method you throw an `ArticleOfTypeNotFoundException` if no articles of that type are found, great but you don't throw any exception if no articles were present at all? I am not sure if we can have that case but if it feels inconsistent. For the type ones you do the mapping and check the resulting list if it is empty but why not do the same with articles of every type? It could be than right after line 101 and by already throwing an exception there the rest of the method is not used

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/NormService.java
```diff
@@ -90,20 +92,28 @@ public Optional<String> updateNormXml(UpdateNormXmlUseCase.Query query)
...
+            .getArticles();
 
     if (query.refersTo() == null) {
       return articles.stream().map(a -> XmlMapper.toString(a.getNode())).toList();
     } else {
-      return articles.stream()
-          .filter(a -> Objects.equals(a.getRefersTo().orElse(""), query.refersTo()))
-          .map(a -> XmlMapper.toString(a.getNode()))
-          .toList();
+      var articlesOfType =
```
VictorDelCampo, line 106: Good but the checking if the resulting list is empty could be done in the same chain right after it is collected and returning directly the result of the stream. Like this (only changed the else part):
```

    if (query.refersTo() == null) {
      return articles.stream().map(a -> XmlMapper.toString(a.getNode())).toList();
    } else {
      return articles.stream()
          .filter(a -> Objects.equals(a.getRefersTo().orElse(""), query.refersTo()))
          .map(a -> XmlMapper.toString(a.getNode()))
          .collect(
              Collectors.collectingAndThen(
                  Collectors.toList(),
                  result -> {
                    if (result.isEmpty())
                      throw new ArticleOfTypeNotFoundException(query.eli(), query.refersTo());
                    return result;
                  }));
    }
```
    
 But it is a question of taste :)

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/NormService.java
```diff
@@ -26,7 +26,7 @@ public class NormService
     implements LoadNormUseCase,
         LoadNormXmlUseCase,
         UpdateNormXmlUseCase,
-        LoadSpecificArticleXmlFromNormUseCase,
+        LoadSpecificArticlesXmlFromNormUseCase,
```
VictorDelCampo, line 29: Should probably be implemented in the `ArticleService` and not the `NormService` right?

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/controller/ArticleController.java
```diff
@@ -197,31 +147,26 @@ public ResponseEntity<String> getArticlesRender(
...
+        foundArticle
             .getAffectedDocumentEli()
-            .flatMap(
-                eliTargetLaw -> loadNormUseCase.loadNorm(new LoadNormUseCase.Query(eliTargetLaw)))
-            .map(
-                targetLaw ->
-                    loadZf0UseCase.loadOrCreateZf0(
-                        new LoadZf0UseCase.Query(optionalNorm.get(), targetLaw, true)))
+            .map(eliTargetLaw -> loadNormUseCase.loadNorm(new LoadNormUseCase.Query(eliTargetLaw)))
+            .map(targetLaw -> new LoadZf0UseCase.Query(norm, targetLaw, true))
```
VictorDelCampo, line 164: Nice stream map chaining creating the query in one step before and lambda-calling the `loadZf0UseCase`!

backend/src/main/java/de/bund/digitalservice/ris/norms/application/port/input/LoadNormUseCase.java
```diff
@@ -14,7 +15,7 @@ public interface LoadNormUseCase {
    * @param query The query containing the ELI (European Legislation Identifier) of the norm.
    * @return An {@link Optional} containing the loaded {@link Norm} if found, or empty if not found.
    */
-  Optional<Norm> loadNorm(Query query);
+  Norm loadNorm(Query query) throws NormNotFoundException;
```
VictorDelCampo, line 18: Topic if we want to add runtime exceptions in the method signature, specially when defining the interface

backend/src/main/java/de/bund/digitalservice/ris/norms/application/port/input/LoadArticlesFromNormUseCase.java
```diff
@@ -0,0 +1,37 @@
...
+   */
+  List<Article> loadArticlesFromNorm(Query query) throws NormNotFoundException;
+
+  /**
+   * Contains the parameters needed for loading articles from a norm.
+   *
+   * @param eli The ELI used to identify the norm
+   * @param amendedBy ELI of an amending law. When specified, only articles with passive
+   *     modifications from that amending law are included in the result.
+   * @param amendedAt eId of a lifecycle event. When specified, only articles with passice
```
VictorDelCampo, line 28: Type, passive not passice ;)

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/controller/ArticleController.java
```diff
@@ -67,105 +66,56 @@ public ResponseEntity<List<ArticleResponseSchema>> getArticles(
...
+    final var articlesWithZf0 =
+        loadArticlesFromNormUseCase.loadArticlesFromNorm(query).stream()
             .map(
                 article -> {
-                  var targetLawZf0 =
+                  final var targetLawZf0 =
                       article
                           .getAffectedDocumentEli()
-                          .flatMap(
+                          .map(
```
VictorDelCampo, line 83: It is already a great optimization but I was a bit confused at the beginning (when reading, I know we talked about it) of why the first line is getting the `Norm` object with the passed eli but then we call the getArticlesFromNorm (or however it is called) to retrieve the articles of that norm we have already retrieved. But sure we do more compiclated stuff.... anyway coming to my proposal: i would put this loadNorm, that we just need to calling the LoadOrCreateZf0 thing inside of the mapping when creating the Query. Like this:
```
                  final var targetLawZf0 =
                      article
                          .getAffectedDocumentEli()
                          .map(
                              eliTargetLaw ->
                                  loadNormUseCase.loadNorm(new LoadNormUseCase.Query(eliTargetLaw)))
                          .map(targetLaw -> {
                            final var norm = loadNormUseCase.loadNorm(new LoadNormUseCase.Query(eli.getValue()));
                            return new LoadZf0UseCase.Query(norm, targetLaw, true);
                          })
                          .map(loadZf0UseCase::loadOrCreateZf0)
                          .orElse(null);
```

Because it is only needed there and not before. I think we should refactor the LoadOrCreateZf0Service but it is not the focus of this PR

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/ArticleService.java
```diff
@@ -28,27 +33,101 @@ public ArticleService(
...
+    final var norm =
+        loadNormPort
+            .loadNorm(new LoadNormPort.Command(query.eli()))
+            .orElseThrow(() -> new NormNotFoundException(query.eli()));
+
+    List<Article> articles = norm.getArticles();
+
+    // Filter list of articles by passive mods if at least one filter is specified
+    if (amendedBy != null || amendedAt != null) {
+      var filterPassiveMods = getPassiveModsAmendingByOrAt(norm, amendedBy, amendedAt);
```
VictorDelCampo, line 74: I get your encapsulation to have smaller methods by for me it was more confusing because of the creation if a Predicate<> applied then in the filter...

I would just have 1 big complicated method but maybe a bit more structured, like:

```
    // Filter list of articles by passive mods if at least one filter is specified
    if (amendedBy != null || amendedAt != null) {
      return filterArticlesByAmendedByOrAt(norm, amendedBy, amendedAt);
    }
```

With the big method being:
    
```
private List<Article> filterArticlesByAmendedByOrAt(final Norm fromNorm, final String amendedBy, final String amendedAt) {
    if (amendedBy == null && amendedAt == null) return List.of();

    final List<TextualMod> passiveMods = fromNorm.getMeta()
            .getAnalysis()
            .map(Analysis::getPassiveModifications)
            .orElse(Collections.emptyList());

    final List<TextualMod> filteredPassiveMods = passiveMods.stream()
            .filter(passiveModification -> {
              if (amendedBy != null) {
                return passiveModification.getSourceHref()
                        .flatMap(Href::getEli)
                        .map(amendedBy::equals)
                        .orElse(false);
              }
              return true;
            })
            .filter(passiveModification -> {
              if (amendedAt != null) {
                return passiveModification.getForcePeriodEid()
                        .flatMap(fromNorm::getStartEventRefForTemporalGroup)
                        .map(amendedAt::equals)
                        .orElse(false);
              }
              return true;
            })
            .toList();

    return fromNorm.getArticles().stream().filter(article ->  // If we filter by amendedAt or amendedBy: Those properties are found
            // in the passive modifications we already collected above. What's left
            // now is to only return the articles that are going to be modified by
            // those passive modifications.
            filteredPassiveMods.stream()
                    .map(TextualMod::getDestinationHref)
                    .flatMap(Optional::stream)
                    .map(Href::getEId)
                    .flatMap(Optional::stream)
                    .anyMatch(
                            destinationEid ->
                                    // Modifications can be either on the article itself or anywhere
                                    // inside the article, hence the "contains" rather than exact
                                    // matching.
                                    destinationEid.contains(article.getEid().orElseThrow()))).toList();

  }
```

VictorDelCampo (commented):
Awesome work Andreas. Thanks for your backend-refactoring efforts. I have some change proposals. If you like them just use them. Otherwise in general is almost everything good

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/NormService.java
andreasphil: Didn't think about it, this has existed before and I only renamed it. I'll have a look!

andreasphil:
As the changes in the feedback were suggested but not required, I'm merging this already to avoid creating a huge PR. Will make some changes in a follow up.

backend/src/main/java/de/bund/digitalservice/ris/norms/application/port/input/LoadArticleHtmlUseCase.java
andreasphil: Thanks for the detailed explanation!

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/NormService.java
andreasphil: I kept it like this because that was the original behavior, but you're right it's not very consistent. Will update it to return a 404 when no articles are found

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/ArticleService.java
andreasphil: As discussed—will keep it like this

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/NormService.java
andreasphil: Interesting to see the approach with `collect` but if it comes down to taste I'll stick with the current implementation as I find that more readable 😇

PR #559, opened 2024-08-28
Add empty states to references page

VictorDelCampo:
Difference to the empty state on the textko page is that we do see the title displayed over the empty state message. We should align the designs with the textko page, but in my opinion keeping the title looks better. Therefore it is implemented like this here. I question that already on a slack thread but it got lost into nirvana.

Just double check the variables I am using for the v-if because I am not 100% sure if it is nice

- For displaying "please select a mod first" (middle panel) I am v-if-ing the `selectedModQuotedContentXmlString`, which is computed from `selectedModQuotedContent`, which is itself computed from the prop `selectedModEId` and the ref `amendingNormDocument` which is created out of the prop `normXml`. To be honest I got lost in the computed chain but that property seemed to be the right now. In order not to display the title "Verweise" of the right panel, I use the same v-if but for the whole section, in the same component.
- For displaying "no refs found" (right panel) I am checking if the computed `refs` array is empty. I think this is the right one. There is one side effect that when you delete one entry of the table, the empty state is displayed (since the in-memory XML is updated and does not contain any refs anymore). And you need to click on save so that the updated XML is persisted to the backend. If you forget to click on save, the ref will still be there. But IMO this was the behavior before already, you deleted the entry, disappeared from the table but you still needed to click on save. I think this needs further iteration by design.

Regarding tests, I added:
- Two new E2E tests, for each empty state. Even if the empty states are in components (not views) I think it is nice to test them from the user perspective.
- But of course also added one unit test per component containing the respective empty state

### DIFF
--- a/frontend/e2e/amending-law-references.spec.ts
+++ b/frontend/e2e/amending-law-references.spec.ts
@@ -1,6 +1,8 @@
 import { test, expect } from "@playwright/test"
 
-test("navigate to amending law overview", async ({ page }) => {
+test("navigate to amending law references page without selected mods", async ({
+  page,
+}) => {
   await page.goto("/amending-laws")
   await page.getByRole("link", { name: "BGBl. I 1002 Nr. 10" }).click()
   await page.getByRole("link", { name: "Betroffene Normenkomplexe" }).click()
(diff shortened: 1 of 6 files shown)


andreasphil (approved):

andreasphil:
@VictorDelCampo The variables make sense to me. And I agree that it is confusing that you need to press save, that has been the behavior before though. I think the idea is to eventually add confirmation prompts when trying to navigate away from a page with unsaved changes but haven't really planned that yet.

PR #561, opened 2024-08-29
Risdev-4710 move announcement controller error handling to controller advice

hannes-mk:
Next to making the service methods return non-Optionals, this PR introduces an `AnnouncementNotFound` exception and its handler.


backend/src/main/java/de/bund/digitalservice/ris/norms/application/exception/AnnouncementNotFoundException.java
```diff
@@ -0,0 +1,9 @@
+package de.bund.digitalservice.ris.norms.application.exception;
+
+/** Indicates that the requested norm does not exist. */
```
andreasphil: Typo: should say "announcement" instead of "norm"

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/exceptions/AnnouncementExceptionHandler.java
```diff
@@ -0,0 +1,38 @@
...
+
+  /**
+   * Exception handler method for handling {@link AnnouncementNotFoundException}.
+   *
+   * @param e The exception that occurred.
+   * @return A {@link ResponseEntity} with an HTTP 404 status and the exception message.
+   */
+  @ExceptionHandler(AnnouncementNotFoundException.class)
+  @ResponseStatus(HttpStatus.NOT_FOUND)
+  public ResponseEntity<String> handleException(final AnnouncementNotFoundException e) {
```
andreasphil: Todo: can you please move this to `NormExceptionHandler`? That one is meant to handle our custom exceptions, so we don't need to create a new ControllerAdvice for every single exception we have.

(as a side note, I'll admit that the name `NormExceptionHandler` is a bit confusing. "Norm" in this context doesn't mean our Norm domain entity but "the Norms application")

backend/src/main/java/de/bund/digitalservice/ris/norms/application/exception/AnnouncementNotFoundException.java
```diff
@@ -0,0 +1,9 @@
...
+public class AnnouncementNotFoundException extends RuntimeException {
+
+  public AnnouncementNotFoundException(final String eli) {
+    super("Announcement with eli %s does not exist".formatted(eli));
```
andreasphil: Suggestion: Announcements themselves don't have ELIs, but they refer to a norm with an ELI, right? In that case the message should read something like "Announcement for norm with eli XY does not exist".

backend/src/main/java/de/bund/digitalservice/ris/norms/application/exception/AnnouncementNotFoundException.java
hannes-mk: exactly. Good catch!

backend/src/main/java/de/bund/digitalservice/ris/norms/application/exception/AnnouncementNotFoundException.java
hannes-mk: Frankly, I did not know. Thanks!

backend/src/main/java/de/bund/digitalservice/ris/norms/application/exception/AnnouncementNotFoundException.java
hannes-mk: exactly. Good catch!

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/exceptions/AnnouncementExceptionHandler.java
hannes-mk: Right, it was the naming that made me create a new one. Will move.

PR #565, opened 2024-08-30
Rename the app's "global" exception handler

hannes-mk:
RISDEV-0000


hannes-mk:
Will be done in  branch https://github.com/digitalservicebund/ris-norms/tree/RISDEV-4635_implement-error-responses

PR #569, opened 2024-08-30
Switch java formatter to Prettier + format

andreasphil:
RISDEV-0000

---

As discussed in our dev channel + weekly, I had a look at ways to improve the automatic formatting of our Java code. The issues we want to fix are:

- Consistent indentation; currently our code has a mix of 2, 4 and 8 spaces
- When chaining method calls, each call should have its own line
- Parameters also should each have their own line if they don't fit into one

Our goal initially was to achieve this by configuring our current formatter, rather than switching formatters.

However, it turns out that our [Spotless](https://github.com/diffplug/spotless/tree/main/plugin-gradle) config was using [google-java-format](https://github.com/google/google-java-format) which intentionally cannot be configured:

> **Note:** There is no configurability as to the formatter's algorithm for formatting. This is a deliberate design decision to unify our code formatting on a single format.

Besides google-java-format, Spotless also has out of the box support for other formatters, among them [Prettier](https://prettier.io/) and [palantir-java-format](https://github.com/palantir/palantir-java-format).

**palantir-java-format** is a fork of google-java-format. Therefore it is close to what we're already using, but it has improvements especially around how lambdas are laid out. It also can't be configured. I tried it and while it is an improvement over the Google formatter, it does not solve any of our issues. In addition, it seems to be using spaces for indentation even more liberally, making the already verbose Java code take up even more space.

For this reason I decided to configure Prettier and format our code with that. I left the settings at default, apart from slightly increasing print width from 80 to 100 (again, because Java is so verbose). IMO this is vastly more readable and fixes all 3 issues mentioned above, while not really changing anything else (as far as I can tell).

Using Prettier has the additional benefit that we're using one formatter for basically our entire code base (FE + BE). This means that our code looks very consistent across the entire stack. It also doesn't require any new dependencies because it is already used in other places in our Spotless config, so it is there anyways.

### DIFF
--- a/backend/build.gradle.kts
+++ b/backend/build.gradle.kts
@@ -166,7 +166,19 @@ tasks {
 spotless {
     java {
         removeUnusedImports()
-        googleJavaFormat()
+
+        prettier(
+            mapOf(
+                "prettier" to "3.0.3",
+                "prettier-plugin-java" to "2.3.0",
+            ),
+        ).config(
+            mapOf(
+                "parser" to "java",
+                "printWidth" to 100,
+                "plugins" to listOf("prettier-plugin-java"),
+            ),
+        )
     }
 
     kotlinGradle {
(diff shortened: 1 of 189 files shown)


andreasphil:
Closing this for now as we're not sure how to proceed with this.

PR #582, opened 2024-09-05
Types and basic mapping for error handling frontend implementation

andreasphil:
This PR adds types and basic logic for mapping error responses (as described in the [ADR](https://github.com/digitalservicebund/ris-norms/blob/main/doc/adr/0012-error-responses.md)) to human-readable messages in the frontend. I played around with a few different options until I found one I was satisfied with. There might still be things to simplify but I would already be interested in feedback on the general approach.

My goals were:

- Support for plain messages as well as interpolations (i.e. "Norm with ELI {eli} was not found"). For interpolations I wanted to not "invent" our own format and also if possible avoid adding dependencies, so template strings seemed like a good option.

- Use TypeScript as much as possible to let us know if messages are missing or the template strings need to be updated because parameters have changed. With this solution, we declare the types of errors we expect as well as any extension fields of that error in the `types/errorResponse.ts`, and TypeScript will let us know if we need to add or update existing messages.

- Support in the UI for both simple messages as well as messages with a title and additional description, hence the mapping returning an object rather than a string.

If this looks good so far, the next steps would be to:

- Add unit tests for the mapping
- Integrate with our reactive API calls
- Update components to display in the UI
- And then of course at some point add the actual mappings and types, currently these are just examples for testing

### DIFF
--- a/frontend/src/lib/errorResponseMapper.ts
+++ b/frontend/src/lib/errorResponseMapper.ts
@@ -18,8 +18,7 @@ import {
  * @returns Whether the candidate is an error response
  */
 export function isErrorResponse(
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any -- Fetch errors are always any
-  e: any,
+  e: any, // eslint-disable-line @typescript-eslint/no-explicit-any -- Fetch errors are always any
 ): e is ErrorResponse {
   return typeof e === "object" && e.type in errorMessages
 }
@@ -35,7 +34,7 @@ export function isErrorResponse(
  */
 export function mapErrorResponse(e: ErrorResponse): MappedErrorResponse {
   const mapper = (errorMessages as ErrorResponseMapping)[e.type]
-  let message = mapper?.(e) ?? errorMessages.__fallback__
+  let message = mapper?.(e) ?? errorMessages.__fallback__()
   if (typeof message === "string") message = { title: message }
   return message
 }
--- a/frontend/src/lib/errorResponseMapper.ts
+++ b/frontend/src/lib/errorResponseMapper.ts
@@ -1,6 +1,7 @@
+import { errorMessages } from "@/lib/errorMessages"
 import {
   ErrorResponse,
-  ErrorResponseMessageMapping,
+  ErrorResponseMapping,
   MappedErrorResponse,
 } from "@/types/errorResponse"
 
@@ -17,11 +18,10 @@ import {
  * @returns Whether the candidate is an error response
  */
 export function isErrorResponse(
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any -- Has to be any because fetch errors are always any
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any -- Fetch errors are always any
   e: any,
-  mapping = errorMessages,
 ): e is ErrorResponse {
-  return typeof e === "object" && e.type in mapping
+  return typeof e === "object" && e.type in errorMessages
 }
 
 /**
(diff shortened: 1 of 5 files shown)


frontend/src/lib/errorResponseMapper.ts
```diff
@@ -0,0 +1,63 @@
...
+  let message = mapper?.(e) ?? mapping.__fallback__
+  if (typeof message === "string") message = { title: message }
+  return message
+}
+
+/* -------------------------------------------------- *
+ * All error messages                                 *
+ * -------------------------------------------------- */
+
+const errorMessages: ErrorResponseMessageMapping = {
```
andreasphil: TypeScript will complain if this does not include _all_ keys listed in `ErrorResponseDetail`, including the fallback 👌

frontend/src/types/errorResponse.ts
```diff
@@ -0,0 +1,95 @@
+/**
+ * An error response returned by the API. This implements the error responses
+ * ADR, please refer to doc/adr/0012-error-responses.md for more information.
+ */
+export type ErrorResponse<
+  T extends keyof ErrorResponseDetail = "__fallback__",
```
andreasphil: The default to `__fallback__` is needed so we can use `ErrorResponse` without a generic param in places where we don't care about the specific type of error (e.g. in the mapper).

frontend/src/types/errorResponse.ts
```diff
@@ -0,0 +1,95 @@
...
+  Children = never,
```
andreasphil: Currently not used, we will need to update this once we actually support/need nested errors. Not necessary in the first iteration IMO but wanted to already add some kind of marker.

frontend/src/types/errorResponse.ts
```diff
@@ -0,0 +1,95 @@
...
+   * locations in the UI (e.g. an invalid form field).
+   */
+  instance?: string
+
+  /**
+   * Nested errors. If an error response contains multiple messages (e.g.
+   * for listing all fields that failed validation of a form), they will
+   * appear here.
+   */
+  errors?: Array<Children>
```
VictorDelCampo: Shouldn't the childen errors be of type themselves `ErrorDetail`? Not sure if it is correct but I remember that we say that the errors themselves are of the same type as the upper parent?

frontend/src/lib/errorResponseMapper.ts
```diff
@@ -0,0 +1,63 @@
...
+ * @returns
+ */
+export function mapErrorResponse(
+  e: ErrorResponse,
+  mapping = errorMessages,
+): MappedErrorResponse {
+  const mapper = mapping[e.type]
+  let message = mapper?.(e) ?? mapping.__fallback__
+  if (typeof message === "string") message = { title: message }
+  return message
```
malte-laukoetter:
```suggestion
  const message = mapper?.(e) ?? mapping.__fallback__
  if (typeof message === "string") {
    return { title: message }
  }
  return message
```

frontend/src/types/errorResponse.ts
```diff
@@ -0,0 +1,95 @@
...
+ */
+export type ErrorResponseMessageMapping = {
+  [K in keyof ErrorResponseDetail]: (
+    e: ErrorResponse<K>,
+  ) => MappedErrorResponse | string
+}
+
+/**
+ * The human-readable error message that should be displayed in the UI for
+ * for each error type returned by the API.
```
malte-laukoetter:
```suggestion
 * each error type returned by the API.
```

frontend/src/lib/errorResponseMapper.ts
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * For any backend error response, maps the technical details of that response
+ * to a human-readable message in the UI.
+ *
+ * @param e Response received from the backend
+ * @param mapping Mapping to retrieve the messages from. The default mapping should
+ *  already include everything you need, this is only intended for testing.
+ * @returns
+ */
+export function mapErrorResponse(
```
malte-laukoetter: I'm a bit unsure how much this whole setup really provides us with type safety and how much is just checking that the type of `ErrorResponseDetail` and the error messages are aligned.

In `mapErrorResponse` TS for example assumes that only the `__fallback__` case ever happens:
Screenshot 2024-09-06 at 08 50 31

I think we could achieve the same using this and keep the type next to its use. This would also allow us to support the children errors in a simpler way and is probably at general a bit simpler:
```ts
export type ErrorResponse2<
  ExtensionFields extends object = object,
  Children = never,
> = {
  type: string
  title?: string
  details?: string
  instance?: string
  errors?: Array<Children>
} & {
  [K in keyof ExtensionFields]?: ExtensionFields[K]
}

const messages = {
  __fallback__: (e) =>
    "Ein unbekannter Fehler ist aufgetreten.",
  "/errors/article-not-found": (e: ErrorResponse2<{ eid: string }>) => ({
    title: `Article with eID ${e.eid} not found`,
  }),
  "/errors/article-of-type-not-found": (e: ErrorResponse2<{ eli: string }>) =>
    `Norm with eli ${e.eli} does not include articles of type ${e.type?.join(",")}`,
  "/errors/norm-not-found": (e) =>
    `Norm with ELI ${e.instance} not found`,
} satisfies {
  [key: string]: (e: ErrorResponse2) => string | MappedErrorResponse
}
```

We would then still need to adjust this function here to do some casts to `any` but that is something we will need anyway (unless we do typechecks at runtime using something like zod).

malte-laukoetter (changes requested):

PR #593, opened 2024-09-11
Treat eId as mandatory value

SebastianRossa:
In the context of introducing the custom standardized error responses in the backend we had to do some refactoring so that all exceptions are handled and converted to the ProblemDetail object.

We realized that we are using in many places the default `NoSuchElementException`, which is thrown, when on a Optional the `orElseThrow()` is used. We wanted to replace this default with a custom one, like maybe using the `MandatoryNodeNotFoundException`. But we realized that many of this default usage comes when trying to get the eId of an element.

We started therefore reformating the fromNode function of the Eid class so that not an optional is returned, since we know all/many elements MUST have an eId. We could then just filter out those elements passed to that function that do not belong to the list of nodes with mandatory eId.

But we realized that firstly with LDML.de 1.7 almost every XML node will get a mandatory eId (like for example akn:meta and many children there). Therefore we park this implementation for when:

1. XSD validation after upload is in place, because we can then be sure that our DB won't have amending laws with elements without eId attribute (if they must have it)
2. We convert to LDML.de 1.7 because almost every node will need to have an eId.

We added for now an exception handler for the `NoSuchElementException` as a fallback till we go on with this work.

### DIFF
--- a/backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/controller/ArticleController.java
+++ b/backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/controller/ArticleController.java
@@ -157,7 +157,7 @@ public class ArticleController {
 
     final var foundArticle =
         norm.getArticles().stream()
-            .filter(article -> article.getEid().isPresent() && article.getEid().get().equals(eid))
+            .filter(article -> article.getEid().equals(eid))
             .findFirst()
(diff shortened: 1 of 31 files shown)


andreasphil:
Closing this PR as it has become quite outdated and it's unclear how to proceed. Further work on this will be specified and tracked here: https://digitalservicebund.atlassian.net/browse/RISDEV-5323

PR #595, opened 2024-09-12
docs: add API docs for create annonucement endpoint

andreasphil:
RISDEV-0000

### DIFF
--- a/doc/backend/norms-api-v1.yaml
+++ b/doc/backend/norms-api-v1.yaml
@@ -28,10 +28,13 @@ paths:
       requestBody:
         required: true
         content:
-          application/xml:
+          multipart/form-data:
             schema:
-              type: string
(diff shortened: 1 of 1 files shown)


doc/backend/norms-api-v1.yaml
```diff
@@ -23,6 +23,34 @@ paths:
...
                 $ref: '#/components/schemas/ProblemDetail'
+    post:
+      summary: Create a new announcement by uploading the XML of a norm
+      requestBody:
+        required: true
+        content:
+          application/xml:
+            schema:
+              type: string
+              format: binary
```
malte-laukoetter:
```suggestion
          multipart/form-data:
            schema:
              type: object
              properties:
                file:
                  type: string
                  format: binary
```

PR #597, opened 2024-09-12
RISDEV-4827 run schematron validation

malte-laukoetter:


backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/LdmlDeValidator.java
```diff
@@ -85,4 +99,81 @@ public void fatalError(SAXParseException exception) {
...
+   * @throws IOException The schematron file couldn't be read
+   * @throws XmlProcessingException A problem occurred during the processing of the XML
+   * @throws LdmlDeSchematronException A Schematron rule is violated
+   */
+  public void validateSchematron(Norm norm) throws IOException {
+    Source xsltSource = new StreamSource(schematronXslt.getInputStream());
+    xsltSource.setSystemId(schematronXslt.getURL().toString());
+    Transformer transformer = null;
+    try {
+      transformer = new TransformerFactoryImpl().newTransformer(xsltSource);
```
VictorDelCampo, line 116: Creating a `Transformer` can be expensive. If the `schematronXslt` is a constant and doesn't change often, why not reusing a single Transformer object instead of creating one every time

You could create the object in the constructor of the `LdmlDeValidator` and ths way it will be created just once when the service is instantiated, something like this

```
    public LdmlDeValidator(@Value("classpath:schema/fixtures/legalDocML.de.xsl") Resource schematronXslt) {
        this.transformer = createTransformer(schematronXslt);
    }

    private Transformer createTransformer(Resource schematronXslt) {
        try {
            Source xsltSource = new StreamSource(schematronXslt.getInputStream());
            xsltSource.setSystemId(schematronXslt.getURL().toString());
            return new TransformerFactoryImpl().newTransformer(xsltSource);
        } catch (IOException | TransformerConfigurationException e) {
            throw new XmlProcessingException(e.getMessage(), e);
        }
    }
```

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/LdmlDeValidator.java
```diff
@@ -85,4 +99,81 @@ public void fatalError(SAXParseException exception) {
...
+    } catch (TransformerConfigurationException e) {
+      throw new XmlProcessingException(e.getMessage(), e);
+    }
+
+    String xmlText = XmlMapper.toString(norm.getDocument());
```
VictorDelCampo, line 121: Why not just directly converting the `Document` to a `Source` --> `DomSource` to pass it to the transformer?

`Source xmlSource = new DOMSource(norm.getDocument())`

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/LdmlDeValidator.java
```diff
@@ -85,4 +99,81 @@ public void fatalError(SAXParseException exception) {
...
+              .map(
+                  node -> {
+                    // The location includes an XPath with expanded QNames
+                    // (Q{namespace}<localPart>).
+                    var xPath = node.getAttributes().getNamedItem("location").getNodeValue();
+
+                    // Find the eId of the node responsible for this problem. Sometimes the location
+                    // points to an attribute, so we might need to move up in the element tree to
+                    // find the xPath.
+                    List<Node> eIdNodes =
```
VictorDelCampo, line 157: Should we maybe cache the xpath results? So that we don't need to compute it again if we are querying the same xpath, since xpath evaluation is relatively expensive and the catching can be easily achieved with:

```
                          List<Node> eIdNodes = xPathCache.computeIfAbsent(
                              xPath + "/ancestor-or-self::*/@eId",
                              key -> NodeParser.getNodesFromExpression(key, norm.getDocument())
                          );
```

PR #606, opened 2024-09-16
RISDEV-4845 add ldml.de schema files directly into our repo

malte-laukoetter:


backend/build.gradle.kts
```diff
@@ -155,11 +166,45 @@ tasks {
...
+                            val pipelineFile =
+                                // find the jar file for name.dmaus.schxslt:schxslt and look into it as if it was a zip archive
+                                zipTree(
+                                    classpath.filter {
+                                        it.name.startsWith("schxslt")
+                                    }.singleFile,
+                                ).filter {
+                                    // file the pipeline xsl file for converting sch to xsl in this jar
+                                    it.absolutePath.endsWith("/2.0/pipeline-for-svrl.xsl")
+                                }.singleFile
```
malte-laukoetter, line 200: I'm unhappy with how the loading of the xsl file from the dependency is done but couldn't figure out another way :/

PR #614, opened 2024-09-19
risdev-4873 Implement toast component instead of tooltip

hamo225:


frontend/e2e/amending-law-metadata-editor-rahmen.spec.ts
```diff
@@ -500,9 +500,7 @@ test.describe("metadata view", () => {
     await metadataPage.saveMetadata()
 
     // Then
-    await expect(
-      page.getByRole("tooltip", { name: "Gespeichert!" }),
-    ).toBeVisible()
+    await expect(page.getByText("Gespeichert!")).toBeVisible()
```
malte-laukoetter: has the toast component any a11y role we could check here instead of just checking that the test is on the page?

malte-laukoetter (approved):

PR #623, opened 2024-09-24
Ignore io.sentry:sentry in versionCatalogUpdate

VictorDelCampo:
For sentry we are using the sentry jvm gradle plugin, which is resolving the main sentry dependency of `io.sentry:sentry` in version `7.13.0`

But our bi-weekly versionCatalogUpdate is finding a missing update:

```
The following dependencies have later release versions:
 - io.sentry:sentry [7.13.0 -> 7.14.0]
     https://github.com/getsentry/sentry-java
```

Meaning that job will always fail because we are not directly using that dependency but letting the sentry plugin resolve it.

In order to let the version-catalog-task ignore that dependency I have added a custom config for the task in our `build.gradle.kts`. This way we won't get the "missing" update anymore but rather a warning in the output:

```
Failed to determine the latest version for the following dependencies (use --info for details):
 - io.sentry:sentry
     7.13.0
```
Which we can ignore. Our job in the pipeline is only checking for the property `outdated.count`  `build/dependencyUpdates/report.json` and will therefore no longer fail.

One side effect of adding this custom config is that the default behavior of the task not updating to unstable versions is not longer in place. Therefore I manually added the recommended function from the creators of the plugin ([here](https://github.com/ben-manes/gradle-versions-plugin?tab=readme-ov-file#rejectversionsif-and-componentselection)) using the example 2 to disallow release candidates as upgradable versions from stable versions

### DIFF
--- a/backend/build.gradle.kts
+++ b/backend/build.gradle.kts
@@ -1,4 +1,5 @@
 import com.adarshr.gradle.testlogger.theme.ThemeType
+import com.github.benmanes.gradle.versions.updates.DependencyUpdatesTask
 import com.github.jk1.license.filter.DependencyFilter
 import com.github.jk1.license.filter.LicenseBundleNormalizer
 import com.github.jk1.license.render.CsvReportRenderer
@@ -144,7 +145,8 @@ tasks {
         useJUnitPlatform { excludeTags("integration") }
     }
 
-    withType(com.github.benmanes.gradle.versions.updates.DependencyUpdatesTask::class) {
+    // Needed to ignore io.sentry:sentry in the task. Adding isStable because default config was being overriden (so that we only get stable versions)
+    withType(DependencyUpdatesTask::class) {
         fun isStable(version: String): Boolean {
             val stableKeyword = listOf("RELEASE", "FINAL", "GA").any { version.uppercase().contains(it) }
             val regex = "^[0-9,.v-]+(-r)?$".toRegex()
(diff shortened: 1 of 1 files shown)


backend/build.gradle.kts
```diff
@@ -276,6 +278,19 @@ tasks.named<Checkstyle>("checkstyleTest") {
     configFile = rootProject.file("checkstyle/config-test.xml")
 }
 
+tasks.named<DependencyUpdatesTask>("dependencyUpdates").configure {
+    rejectVersionIf {
+        candidate.group == "io.sentry" || (isNonStable(candidate.version) && !isNonStable(currentVersion))
+    }
```
malte-laukoetter: is this overwriting the config from line 157 (`rejectVersionIf { !isStable(candidate.version) }`? if yes I think we should remove one of the two configurations

backend/build.gradle.kts
```diff
@@ -276,6 +278,19 @@ tasks.named<Checkstyle>("checkstyleTest") {
...
+}
+
+fun isNonStable(version: String): Boolean {
+    val stableKeyword = listOf("RELEASE", "FINAL", "GA").any { version.uppercase().contains(it) }
+    val regex = "^[0-9,.v-]+(-r)?$".toRegex()
+    val isStable = stableKeyword || regex.matches(version)
+    return isStable.not()
+}
```
malte-laukoetter: this function is now basically twice in this file (line 150 `isStable`), can we reuse one of them?

backend/build.gradle.kts
malte-laukoetter: this is probably the reason that unstable versions where added again

malte-laukoetter (changes requested):
the idea of ignoring sentry updates in that task sounds good :)

backend/build.gradle.kts
VictorDelCampo: Oh true I completely oversaw that config. Moving the sentry ignore to that part

backend/build.gradle.kts
VictorDelCampo: Jep good catch!

PR #630, opened 2024-09-26
Migrate existing components to that of risui

hamo225:
RISDEV-4655


frontend/src/components/RisModForm.spec.ts
```diff
@@ -29,6 +30,9 @@ describe("RisModForm", () => {
 
   it("Should render the form with only mandatory fields", () => {
     render(RisModForm, {
+      global: {
+        plugins: [PrimeVue],
```
andreasphil, line 34: Todo: can we initialize this globally in `vitest-setup.ts`? I expect we'll need to have this in basically every test eventually. Not sure if easily possible though, could you have a look?

hamo225: Yes thought the same thing about making it global and tried initially but did not invest too much time into it. Will merge and have another look after.

PR #681, opened 2024-10-15
Unified error toast handling, show detail text

andreasphil:
RISDEV-4978

### DIFF
--- a/frontend/src/components/controls/RisExpandableText.vue
+++ b/frontend/src/components/controls/RisExpandableText.vue
@@ -1,0 +1,55 @@
+<script setup lang="ts">
+import { useElementId } from "@/composables/useElementId"
+import { ref, useTemplateRef, watchEffect } from "vue"
+
+const { length = 3 } = defineProps<{
+  /**
+   * Specifies the maximum number of visible lines.
+   * @default 3
+   */
+  length?: number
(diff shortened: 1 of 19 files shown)


frontend/src/components/controls/RisExpandableText.vue
```diff
@@ -0,0 +1,59 @@
...
+    </button>
+  </div>
+</template>
+
+<style module>
+.truncate {
+  overflow: hidden;
+  display: -webkit-box;
+  -webkit-box-orient: vertical;
+  -webkit-line-clamp: v-bind(length);
```
malte-laukoetter, line 57: is this the only way to do that styling? according to mdn atleast using `-webkit-box-orient` is not only not a standard property but also deprecated: https://developer.mozilla.org/en-US/docs/Web/CSS/box-orient

andreasphil: I took this from [this guide on MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-line-clamp#truncating_a_paragraph), which is also how Tailwind implements it. I agree the syntax for line clamping is very awkward but as far as I know this is how you do it and how it works in all browsers 🥴

PR #683, opened 2024-10-15
Use href and eli classes in more places

malte-laukoetter:


backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/StringToExpressionEliConverter.java
malte-laukoetter, line 1: I'm a bit unsure about placing this thing here, but i couldn't think of a better place...

PR #695, opened 2024-10-18
Risdev 5142 adjust time machine to update meta section expression eli manifestation eli related metadata

malte-laukoetter:


backend/src/main/resources/db/migration/V0.15__announcement_norm_one_to_many.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE announcements
+    DROP CONSTRAINT announcements_eli_fkey;
```
malte-laukoetter, line 2: todo: change the existing data in the eli column from manifestation to expression eli

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/EliService.java
```diff
@@ -0,0 +1,47 @@
...
+  public ExpressionEli findNextExpressionEli(
+    WorkEli workEli,
+    LocalDate pointInTime,
+    String language
+  ) {
+    var expressionEli = ExpressionEli.fromWorkEli(workEli, pointInTime, 1, language);
+
+    while (true) {
+      try {
+        loadNormUseCase.loadNorm(new LoadNormUseCase.Query(expressionEli));
```
VictorDelCampo, line 40: Why not just doing a sql query to look for all entries in the norms table with an eli_expression that start with the workEli and then in the java code filtering those out with the given pointInTime and language and then finding the latest version to finally determine the next available version? This way you would only have 1 DB call. Not sure how many would be necessary on prod mode, meaning how many versions we would normally have for 1 expression. And using an index on the eli_expression column on our table would be enough probably? Now sure how many data sets we will have but won't be that many. wdyt?

malte-laukoetter: I would almost never expect there to be an existing expression for a given pointInTime as this is only ever the case if multiple amending norms change the same target norm at the same date (or we have a transitive change to the amending norm). So I think it will not lead to any problems that we might need to do multiple db calls here in such a case

PR #709, opened 2024-10-23
Update publish endpoint to update the DB and generate all required files

malte-laukoetter:


backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/output/database/service/DBService.java
```diff
@@ -154,7 +156,19 @@ public void deleteAnnouncementByNormEli(DeleteAnnouncementByNormEliPort.Command
 
   @Override
   @Transactional
-  public void deleteNorm(DeleteNormPort.Command command) {
-    normRepository.deleteByEliManifestation(command.eli().toString());
+  public void deleteUnpublishedNorm(DeleteUnpublishedNormPort.Command command) {
```
VictorDelCampo, line 159: Why having two different output ports for the process of deleting norms by an eli (work or manifestation) and a publish state? Why not just haven 1 output port for both specific use cases. This port would take:
1. Eli (work or manifestation), which delete query to call can be decided by the eli right?
2. Publish state
Then the application service containing the one output port would decide by which eli to delete and on which publish state and the DB service using the only port to decide which sql query to run

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/ReleaseService.java
```diff
@@ -19,22 +31,149 @@ public class ReleaseService implements ReleaseAnnouncementUseCase {
...
+    this.normService = normService;
+    this.timeMachineService = timeMachineService;
+    this.createNewVersionOfNormService = createNewVersionOfNormService;
+    this.deleteQueuedNormsPort = deleteQueuedNormsPort;
+    this.deleteUnpublishedNormPort = deleteUnpublishedNormPort;
   }
 
+  /**
+   * Queue all norms that are affected by the given announcement for publishing.
+   * This includes generating a new manifestation of the amending norm, and new expressions for each time boundary of
```
VictorDelCampo, line 63: Generating a new manifestation of the amending norm even if nothing changed since it was imported?

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/ReleaseService.java
```diff
@@ -19,22 +31,149 @@ public class ReleaseService implements ReleaseAnnouncementUseCase {
...
+   * every target norm. It also creates a new manifestation for the current expression of the target norm so it contains
```
VictorDelCampo, line 64: Why new manifestation of the target norm (ZF0) if we did create this one already when uploading the amending law and we haven't published it yet?

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/ReleaseService.java
```diff
@@ -19,22 +31,149 @@ public class ReleaseService implements ReleaseAnnouncementUseCase {
...
   @Override
+  @Transactional
   public Announcement releaseAnnouncement(ReleaseAnnouncementUseCase.Query query) {
     var announcement = loadAnnouncementByNormEliUseCase.loadAnnouncementByNormEli(
       new LoadAnnouncementByNormEliUseCase.Query(query.eli())
     );
-    announcement.setReleasedByDocumentalistAt(Instant.now());
 
+    // find all norms to publish
+    Set<Norm> normsToPublish = new HashSet<>();
```
VictorDelCampo, line 86: Why do you need a set and then iterate over the set? Won't it always be 1 norm? Since we use the expression eli and the DBService loadNorm is retrieving when passing the expressin eli the latest manifestation right?

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/ReleaseService.java
```diff
@@ -19,22 +31,149 @@ public class ReleaseService implements ReleaseAnnouncementUseCase {
...
+    normsToPublish.add(normService.loadNorm(new LoadNormUseCase.Query(announcement.getEli())));
+
+    for (Norm norm : normsToPublish) {
+      var analysis = norm.getMeta().getAnalysis();
+      if (analysis.isEmpty()) {
```
VictorDelCampo, line 91: This should not happen, currently with delta approach, right? Should we even be allowing publishing an amending law if the analysis is empty? Or throw an error?

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/ReleaseService.java
```diff
@@ -19,22 +31,149 @@ public class ReleaseService implements ReleaseAnnouncementUseCase {
...
+    // different expressions.
+    normsToPublish.forEach(norm ->
+      deleteQueuedNormsPort.deleteQueuedForPublishNorms(
+        new DeleteQueuedNormsPort.Command(norm.getWorkEli())
+      )
+    );
+
+    // generate all future versions for all norms that will be published
+    Set<Norm> allVersionsOfAllNormsToPublish = new HashSet<>();
+    for (Norm norm : normsToPublish) {
```
VictorDelCampo, line 118: No future version needed for the amending law right? And it is also within the set normsToPublish right?

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/ReleaseService.java
```diff
@@ -19,22 +31,149 @@ public class ReleaseService implements ReleaseAnnouncementUseCase {
...
+      }
+
+      allVersionsOfAllNormsToPublish.add(latestNormExpression);
+    }
+
+    allVersionsOfAllNormsToPublish.forEach(norm ->
+      norm.setPublishState(NormPublishState.QUEUED_FOR_PUBLISH)
+    );
+    allVersionsOfAllNormsToPublish.forEach(norm ->
+      updateOrSaveNormPort.updateOrSave(new UpdateOrSaveNormPort.Command(norm))
```
VictorDelCampo: You can also call the port in the same for each from before

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/ReleaseService.java
```diff
@@ -19,22 +31,149 @@ public class ReleaseService implements ReleaseAnnouncementUseCase {
...
+    );
+
+    // Create new manifestations for the expressions that are directly targeted by this announcement (so not the once
+    // only created by the time-machine, but all the other once) so the newest manifestation is still unpublished and
+    // can be edited further
+    normsToPublish.forEach(norm -> {
+      // We need the new manifestations to use tomorrow's date so they do have a different eli to the once just
```
VictorDelCampo, line 160: Is this really needed? If on the same day, after publishing, more changes are done to the same annoucement, won't be working on the same manifestations we had created already (amending law and ZF0) and then just override the QUEUED manifestations of the future versions?

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/ReleaseService.java
malte-laukoetter: its needed if the amending norm was already published before, and in the case of the first publish there is no harm done by doing it anyway. It's probably actually better as then the pointInTimeManifestation is the date of the publishing.

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/ReleaseService.java
malte-laukoetter: Primarily to set the pointInTimeManifestation date to the current date.

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/ReleaseService.java
malte-laukoetter: We also store the target norms in this set so the publishing process can 1. already handle transitive changes and we 2. do not care anymore after this step if a norm is an amending or target norm as we need to do the same steps anyway. (in line 104 the target norms get added to the set)

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/ReleaseService.java
malte-laukoetter: its written in a general way so it can also handle publishing norms that are not amending norms. But yes in theory this should always exist for the amending norm

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/ReleaseService.java
malte-laukoetter: yes there wouldn't be any future versions created for the amending norm. So for amending norms just the `latestNormExpression` will be added to the set.

The normsToPublish include just one version of each norm that is published (so in the normal case the amending norm and the zf0 of the target norm). While `allVersionsOfAllNormsToPublish` contains all the different expressions and manifestations of those norms that will be published. So at the end `allVersionsOfAllNormsToPublish` contains the xmls that will be published.

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/ReleaseService.java
malte-laukoetter: oh true will simplify that :)

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/ReleaseService.java
malte-laukoetter: our application assumes that the zf0 is the newest manifestation of the mentioned expression. But the publishing process created new manifestations for those expressions with dates that are newer than the date of the manifestation of the zf0 and amending norm previously used for editing. So we would try to overwrite the queued files. Therefore we need new manifestations with a newer date so the newest manifestations are still unpublished.

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/output/database/service/DBService.java
malte-laukoetter: oh yes they now use the same logic. I'll combine them :)

malte-laukoetter: nevermind one of those ports works on work level elis and the other one on manifestation elis so we can't combine them that trivially. So I'll keep it as is for now. (Otherwise we should probably be able to handle all combinations of elis and publish-states and I'm not sure we should do that)

VictorDelCampo (approved):

PR #718, opened 2024-10-25
Implementation bucket service (which mock)

VictorDelCampo:
RISDEV-5187


backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/BucketService.java
```diff
@@ -0,0 +1,87 @@
+package de.bund.digitalservice.ris.norms.application.service;
```
malte-laukoetter: Shouldn't this be an output adapter? So similar to the database? I don't think we should care in the application layer about what kind of storage solution we use for the published norms and just call a port.

VictorDelCampo: Jep you are so right! I added two output ports (private and public data)

malte-laukoetter (approved):

PR #727, opened 2024-10-28
Validate resulting versions before publishing

andreasphil:
RISDEV-3201


backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/ReleaseService.java
```diff
@@ -148,6 +143,21 @@ public Announcement releaseAnnouncement(ReleaseAnnouncementUseCase.Query query)
...
+    });
+
+    // Delete the files from a previous release of the same announcement if they are still queued for publishing.
+    // This is to prevent residual files from remaining if some time boundaries changed and this release will create
+    // different expressions.
+    normsToPublish.forEach(norm ->
+      deleteQueuedNormsPort.deleteQueuedForPublishNorms(
+        new DeleteQueuedNormsPort.Command(norm.getWorkEli())
+      )
+    );
```
malte-laukoetter: this needs to happen in the beginning as otherwise we could use wrong expression elis (see the comment).

It should not be a problem that this is modifying the database state as the whole method is annotated with `@Transactional` and will therefore only be executed if everything works out. Atleast thats my understanding, I haven't tested it.

malte-laukoetter (approved):

PR #732, opened 2024-10-29
RISDEV-5120 - Publish service

VictorDelCampo:
### EDIT on 01.11.2024
- Introduced a `Changelog` class for when interacting with the content of the change log (meaning after loading from S3 and before updating it in S3)

### EDIT on 31.10.2024
- Introduced a `Scheduler` component as an input adapter, that will use the input port implemented by the `PublishService`, better aligning with hexagonal architecture

### ORIGINAL TEXT on 30.10.2024

Key aspects:

**BucketService**
- Added two new output ports that are implemented here, so that we can also delete from both public and private buckets (in case one of them failed, meaning if public succeeded but private fails, then we delete from the public)
- I integrated into this very same service the managing of the changelog. I was thinking about creating a new service but would be duplicating much code of injecting the two s3-clients and the bucket names (or passing them around). For testability yes we would be restricted to use the whole service and the public methods.

**PublishService**
- Configurable cron expression (set through the application properties)
- To be scheduled (and run) on staging/uat/production
- Removing of the private metadata moved here from the `ReleaseService`
- Added rollback strategy using the delete output ports, for the reason stated above under the bucket service.
- I wrote a unit test with a different approach, using @MockBean and @SpyBean (more info in the java docs)
- Also added a small integration test for this application service, to test the changing of the publish state of the norm

**Exceptions**
- Added two new exceptions:
  - `StorageException`: abstract one to be used by those concrete exceptions (depending on the storage solution), in our current case the `BucketException`
  - `BucketException`: concrete exception for our BucketService. It will be thrown by the service if any error occurs (including the mapping from the `Document` of the norm into a `String`)

**S3 Mock Client**
- Extended so that it can deals with public/private folders representing two different "buckets" in the same location
- Delete function upgraded to clean the directory of a norm up to the bucket name.
- Added a BaseIntegration tests for those tests interacting with the mock client (so that we can clean after each test and also remove the dir after all test of the class)

**Change not relevant to this PR (sorry)**
- Renaming exceptions package to exception in the `adapter.input.restapi` package


backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/PublishService.java
```diff
@@ -0,0 +1,95 @@
...
+  ) {
+    this.loadNormsByPublishStatePort = loadNormsByPublishStatePort;
+    this.publishPublicNormPort = publishPublicNormPort;
+    this.publishPrivateNormPort = publishPrivateNormPort;
+    this.updateOrSaveNormPort = updateOrSaveNormPort;
+    this.deletePublicNormPort = deletePublicNormPort;
+    this.deletePrivateNormPort = deletePrivateNormPort;
+  }
+
+  @Scheduled(cron = "${publish.cron}")
```
malte-laukoetter: I'm still of the opinion that this should be part of an input adapter that calls this usecase. Also if the usecase is only called from the `@Scheduled` annotation right here there is no need for it to be a usecase at all

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/output/s3/BucketService.java
```diff
@@ -39,15 +50,22 @@
...
 @Service
 @Slf4j
-public class BucketService implements PublishPublicNormPort, PublishPrivateNormPort {
+public class BucketService
+  implements
+    PublishPublicNormPort, PublishPrivateNormPort, DeletePublicNormPort, DeletePrivateNormPort {
+
+  private static final String PUT = "PUT";
+  private static final String DELETE = "DELETE";
+  private static final String CHANGELOG_KEY = "changelog.json";
```
malte-laukoetter: are we using the same changelog file all the time? The agreement with the portal was that we write a new changelog file whenever we change some files. So eg. `2023-12-01.json` on one day and then `2023-12-02.json` at the next day and so on (the actual names of the files do not matter as long as they are unique).

PR #736, opened 2024-10-30
Risdev 5228 adjust return of release endpoint to contain the correct published elis

malte-laukoetter:
* create release entity (to store which norm manifestations are released when)
* change the previous /release endpoint to /releases as there can now be mulitiple releases of the same announcement
  * also change the response schema for a release to include "just" a list of elis instead of two fields for amending and target norms
* update the ReleaseService to create such a release and use them to delete the files created for a previous release that is still QUEUED


malte-laukoetter:
There are two todo comments in the e2e tests about two more things i would like to test in them that i didn't find the time to write tests for today :)

malte-laukoetter:
@VictorDelCampo @andreasphil feel free to merge this, i think everything in here is done (except for the comment about missing e2e test cases) :)

PR #745, opened 2024-11-01
Validate resulting versions before publishing (continued)

andreasphil:
Follow up to

- #727 
- #741 

RISDEV-3201


VictorDelCampo (commented):
LGMT

PR #753, opened 2024-11-04
Use ris-ui Breadcrumb component RISDEV-5098

hamo225:


andreasphil (commented):
The text of links and non-link items is not aligned. Not sure why, maybe they're using different line heights? Or an `items-center` missing somewhere.

image

The SonarCloud are fine, feel free to mark them as accepted. It's good that it warns us about that but since we're using the `href`s from the RouterLink we know they'll be safe to use.

Otherwise LGTM, feel free to merge without another review after making the change.

PR #756, opened 2024-11-05
Use generic error message for unhandled exceptions

andreasphil:
This commit removes the detailed error output from the fallback exception handler that is used as controller advice for all exceptions that aren't handled explicitly. The message is replaced with a generic note that an error has occurred. If we use the exception message we might leak internal information such as DB constraints in the API response.

RISDEV-4871


andreasphil:
@VictorDelCampo No test because as far as I'm aware we don't really test the other controller advice either. Let me know if I should add one. Not sure how and where that would be though.

VictorDelCampo (approved):
LGTM

PR #761, opened 2024-11-06
Refactor usage of Optionals for mandatory eIds

andreasphil:
Follow up to #593 

This PR refactors (almost) all eIds to not use Optionals anymore, in places where they are required in XSD-valid documents.


SebastianRossa (commented):
Looks good

PR #767, opened 2024-11-08
Add lock for scheduler

SebastianRossa:


backend/src/main/java/de/bund/digitalservice/ris/norms/config/SchedulerLockConfig.java
```diff
@@ -0,0 +1,38 @@
...
+import org.springframework.jdbc.core.JdbcTemplate;
+import org.springframework.scheduling.annotation.EnableScheduling;
+
+/**
+ * Configuration class for defining scheduler lock settings in the application. This class is
+ * annotated with {@link Configuration}, {@link EnableSchedulerLock} and {@link EnableScheduling} to
+ * activate the scheduler lock
+ */
+@Configuration
+@EnableScheduling
```
VictorDelCampo, line 18: We do have already this annotation in the app class, applied globably. Shoudn't we also add the @EnableSchedulerLock there and not in the config class?

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/scheduler/Scheduler.java
```diff
@@ -33,8 +35,11 @@ public Scheduler(PublishNormUseCase publishNormUseCase) {
    * The cron schedule is controlled by the property {@code publish.cron}.</p>
    */
   @Scheduled(cron = "${publish.cron}")
+  @SchedulerLock(name = "scheduledPublishToBucket", lockAtMostFor = "240m", lockAtLeastFor = "120m")
```
VictorDelCampo, line 38: Didn't read through the documentation but shouldn't we be setting the lockAtMostFor to a conservative value just above the longest tie we expect the task to run? Since we don't know yet how long will the job be running not sure yet if we should even say a couple of hours... The lockAtLeastFor can be a few minutes I think.
Not sure about the releasing of the lock if the other pod will try even later to run the publication job?

backend/src/main/java/de/bund/digitalservice/ris/norms/config/SchedulerLockConfig.java
SebastianRossa: Good point! I tend to rename the config to SchedulerConfig and do all the configuration there. This way, we separate our concerns, and it is analogous to SecurityConfig and SessionConfig.

backend/src/test/java/de/bund/digitalservice/ris/norms/adapter/input/scheduler/SchedulerTest.java
```diff
@@ -1,38 +0,0 @@
...
-import org.springframework.boot.test.mock.mockito.MockBean;
-import org.springframework.boot.test.mock.mockito.SpyBean;
-import org.springframework.scheduling.annotation.EnableScheduling;
-import org.springframework.test.context.TestPropertySource;
-import org.springframework.test.context.junit.jupiter.SpringExtension;
-
-@ExtendWith(SpringExtension.class)
-@EnableScheduling
-@TestPropertySource(properties = "publish.cron=*/1 * * * * *") // Runs every 5 seconds
-class SchedulerTest {
```
VictorDelCampo, line 20: We dont need this test anymore? It was testing the setting of the properties and check if the job is started at all

VictorDelCampo (changes requested):
See comments

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/scheduler/Scheduler.java
SebastianRossa: Here's a link to the documentation: https://github.com/lukas-krecan/ShedLock 

lockAtLeastFor:
It states: "Lastly, you can set lockAtLeastFor attribute which specifies minimum amount of time for which the lock should be kept. Its main purpose is to prevent execution from multiple nodes in case of really short tasks and clock difference between the nodes."
I think the value can be as high as 24 hours  without any damage since we only run the job once a day. Since our pods are run on one cluster we shouldn't have clock issues and the value could also be as low as one minute.
To be on the save side I decided for 2h but I think it doesn't really matter.

lockAtMostFor:
I only guessed here how long it could take at most plus some buffer time. Maybe we can adjust the value once we know how long it will run?

backend/src/test/java/de/bund/digitalservice/ris/norms/adapter/input/scheduler/SchedulerTest.java
SebastianRossa: The test was breaking due to the AssertLock. I had the impression this test is mostly testing if the scheduler works. Under the assumption: "Don’t Test The Framework" I threw it out.

backend/src/main/java/de/bund/digitalservice/ris/norms/config/SchedulerLockConfig.java
VictorDelCampo: I see... and I see your point. Should we then also move the EnableScheduling to the config?

backend/src/test/java/de/bund/digitalservice/ris/norms/adapter/input/scheduler/SchedulerTest.java
VictorDelCampo: Fine :)

backend/src/main/java/de/bund/digitalservice/ris/norms/config/SchedulerLockConfig.java
SebastianRossa: Yes I will do that.

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/scheduler/Scheduler.java
VictorDelCampo: Alright, fine :)

VictorDelCampo (approved):
LGMT

PR #886, opened 2025-01-07
Move domain specific code from service to Norm object

SebastianRossa:


backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/Norm.java
```diff
@@ -137,6 +137,23 @@ public List<Article> getArticles() {
       .toList();
   }
 
+  /**
+   * Returns a {@link List} of all target norms {@link ExpressionEli}s of an amending {@link Norm}.
+   *
+   * @return The list of target norm elis
+   */
+  public List<ExpressionEli> targetLawElis() {
```
malte-laukoetter:
```suggestion
  public List<ExpressionEli> getTargetLawElis() {
```

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/Norm.java
```diff
@@ -137,6 +137,23 @@ public List<Article> getArticles() {
...
+    return getNodesFromExpression("//body//article[not(ancestor-or-self::mod)]", document)
+      .stream()
+      .map(Article::new)
```
malte-laukoetter:
```suggestion
    return getArticles()
```

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/Norm.java
```diff
@@ -137,6 +137,23 @@ public List<Article> getArticles() {
...
+      .filter(article ->
+        article.getRefersTo().isPresent() &&
+        !article.getRefersTo().get().equals("geltungszeitregel")
```
malte-laukoetter: we could also have this logic in the `Article` class eg. `Article::isAmendingArticle` or `not(Article::isGeltungszeitregel)` or something similar

malte-laukoetter (commented):
I really like moving this kind of logic into the domain layer :) Just added some comments so I don't forget my thoughts on how we could improve this even more ^^

VictorDelCampo (commented):
LGTM :)

PR #1016, opened 2025-02-19
A11Y Setup

hamo225:
RISDEV-6515


.github/workflows/end-to-end-tests.yml
```diff
@@ -58,7 +58,11 @@ jobs:
           npm run build
           npm run preview -- --port 5173 &
           DEV_PID=$!
-          npm run test:e2e -- --project ${{ inputs.project }} --repeat-each 1
+          if [[ "${{ inputs.project }}" == "a11y" ]]; then
```
andreasphil: Question: why is this necessary? This should work:

```sh
npm run test:e2e -- --project a11y --repeat-each 1
```

As far as I can tell both branches of the `if` to the same thing.

frontend/README.md
```diff
@@ -66,6 +66,18 @@ npm run test:e2e -- --project msedge --repeat-each 1
 
 Alternatively the [DEVELOPING.md](../DEVELOPING.md#how-to-run-locally) also explains how to run the e2e-tests inside a docker container.
 
+## Accessibility Testing (a11y)
+
+The project includes automated accessibility (a11y) testing using [axe-core](https://github.com/dequelabs/axe-core).
+
+### **Running Accessibility Tests**
```
andreasphil: nitpick: remove `**`

frontend/a11y/a11y.utils.ts
```diff
@@ -0,0 +1,20 @@
+import { AxeBuilder } from "@axe-core/playwright"
```
andreasphil: todo: rename this file to `a11y-utils.ts` or something like that. `.utils.ts` (with a dot) suggests a specific convention for filenames or purpose of the file which we don't have for utils.

andreasphil (changes requested):
@hamo225 what about tests for the other views? I think we discussed this but don't remember what the agreement was.

hamo225:
@andreasphil We agreed one view for now, since we were not sure what views will stay

PR #1035, opened 2025-02-24
Eli classes in frontend

malte-laukoetter:


frontend/src/types/eli/DokumentExpressionEli.ts
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  static fromString(eli: string): DokumentExpressionEli {
+    const match =
+      /eli\/bund\/(?<agent>[^/]+)\/(?<year>[^/]+)\/(?<naturalIdentifier>[^/]+)\/(?<pointInTime>[^/]+)\/(?<version>[^/]+)\/(?<language>[^/]+)\/(?<subtype>[^/.]+)/.exec(
+        eli,
+      )
+
+    if (match == null) {
+      throw new Error("Invalid Dokument Expression Eli")
```
andreasphil: Question: this means we can potentially have an exception thrown anywhere we use an ELI, right? When in the past the ELIs used to be simple strings. I guess in those cases the exception used to be raised in the backend and then handled by our regular API error handling. Now in theory we can have a lot of uncaught exceptions in the frontend. Do you think that would be an issue in practice?

frontend/src/types/eli/DokumentExpressionEli.ts
```diff
@@ -0,0 +1,65 @@
+import { NormExpressionEli } from "@/types/eli/NormExpressionEli"
```
andreasphil: Nitpick: not sure if `types` is the correct place for this; at least I assumed so far that anything inside `types` is really just types. Should we move the classes to `lib/eli`?

frontend/src/types/eli/DokumentExpressionEli.ts
malte-laukoetter: At the moment the `fromString` methods are only used in tests. In future it might be necessary to turn some stuff from responses from api-endpoints into elis, but in those cases it is already validated by the backend. Or it could be created from some user input, but in that case we would need some validation anyway. So no I don't think it is a problem in practice as this method is in practice not used in a way that could create lots of uncaught exceptions

frontend/src/types/eli/DokumentExpressionEli.ts
malte-laukoetter: I'm not always sure where to draw the line between those two folders. But yes i agree it probably fits better into `lib` so I'll move it there

PR #1039, opened 2025-02-25
E-Verkündung API specification

andreasphil:
RISDEV-6747

### DIFF
--- a/doc/backend/norms-everkuendung-api-v1.yaml
+++ b/doc/backend/norms-everkuendung-api-v1.yaml
@@ -42,6 +42,7 @@ paths:
         - Das Verfahren zum manuellen Übertragen der Daten, wenn die Verkündung wiederholt nicht angenommen wurde, muss noch definiert werden.
         - Sollte die Antwort auf eine erfolgreiche Anfrage weitere Informationen enthalten oder reicht der Statuscode?
         - Das Vorgehen und ggf. die Schnittstelle für die Übergangsphase, d.h. bevor die E-Gesetzgebung Daten aus dem LEA-Editor liefert, muss noch definiert werden.
+        - Eventuell wird es in Zukunft notwendig sein, statt einer synchronen Anfrage eine asynchrone Anfrage zu verwenden, damit das Backend größere Datenmengen besser verarbeiten kann. Es ist zu klären, ob wir die Schnittstelle direkt so gestalten wollen, oder ob das eine künftige Ausbaustufe sein wird.
       security:
         - ClientCredentials: [upload_verkuendung]
       requestBody:
--- a/doc/backend/norms-everkuendung-api-v1.yaml
+++ b/doc/backend/norms-everkuendung-api-v1.yaml
@@ -9,44 +9,50 @@ info:
 
     ## Authentifizierung
 
-    Für die Nutzung aller Endpunkte muss ein gültiger Access Token als `Authorization: Bearer <token>` übermittelt werden. Der Token kann über einen [OAuth2 Client Credentials Grant](https://datatracker.ietf.org/doc/html/rfc6749#section-4.4) abgerufen werden. Die notwendigen Informationen (URL, Client Secret) werden vom DigitalService bereitgestellt.
+    Für die Nutzung aller Endpunkte muss ein gültiger Access Token als `Authorization: Bearer <token>` übermittelt werden. Der Token kann über einen [OAuth2 Client Credentials Grant](https://datatracker.ietf.org/doc/html/rfc6749#section-4.4) abgerufen werden. Die notwendigen Informationen werden vom DigitalService bereitgestellt.
 
 servers:
   - url: /api/v1
     description: Backend
+
+security:
+  - ClientCredentials: [upload_verkuendung]
 
 paths:
   /verkuendungen:
     post:
       summary: Annahme von Verkündungen.
       description: |
-        Die Schnittstelle erlaubt das Hochladen eines Normendokumentationspaketes bis zu einer Größe von 200mb. Das Dokumentationspaket liegt in Form eines ZIP-Archivs vor (`bgbl{1|2}_{jahr}_{nr}.zip`). Es enthält folgende Daten:
+        Die Schnittstelle erlaubt das Hochladen eines Normendokumentationspakets.
 
-        - Rechtsetzungsdokument (XML, PDF)
-        - Regelungstext (XML)
-        - soweit vorhanden: Strukturierte Anlage(n) (XML)
-        - soweit vorhanden: Nicht-strukturierte Anlagen, z.B. eingebundene Bilder, PDF-Formulare (etwa: PDF, JPG, PNG)
-        - optional: konsolidierte Fassung aus der E-Gesetzgebung (PDF)
+        Es kann immer nur eine Anfrage gleichzeitig verarbeitet werden.
 
         Das Archiv wird während der Anfrage innerhalb von 120 Sekunden validiert. Eine erfolgreiche Übermittelung wird mit dem Status-Code `200 OK` bestätigt.
 
-        Wenn die Validierung fehlschlägt oder nicht innerhalb dieser Zeit abgeschlossen werden kann, wird die Anfrage mit einem Fehler beendet. Falls der Fehler aufseiten von NeuRIS liegt, sollte die Übertragung nach einem gewissen Zeitraum ernet versucht werden. Dieser Zeitraum kann dem `Retry-After`-Header der Antwort entnommen werden. Wurde die Verkündung nach 5 Versuchen nicht angenommen, muss ein manueller Austausch mit der Normendokumentation stattfinden.
+        Wenn die Validierung fehlschlägt oder nicht innerhalb dieser Zeit abgeschlossen werden kann, wird die Anfrage mit einem Fehler beendet. Falls der Fehler aufseiten von NeuRIS liegt, sollte die Übertragung nach einem gewissen Zeitraum ernet versucht werden. Dieser Zeitraum kann dem [`Retry-After`-Header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After) der Antwort entnommen werden. Wurde die Verkündung nach 5 Versuchen nicht angenommen, muss ein manueller Austausch mit der Normendokumentation stattfinden.
 
         ## Offene Fragen
 
-        - Die Dateiformate für nicht-strukturierte Anlagen sind noch nicht abschließend geklärt.
-        - Es ist noch offen, welchen Mehrwert die konsolidierte Fassung bringt und ob sich die Übermittelung lohnt.
-        - Art und Format der Signatur müssen noch definiert werden.
-        - Es muss entschieden werden, ob nur das ZIP-Archiv und/oder die Dateien im Archiv signiert werden.
-        - Das Verfahren zum Austausch des öffentlichen Schlüssels für die Verifizierung der Signatur muss noch definiert werden.
+        - Die erlaubten Dateiformate (MIME-Types) für nicht-strukturierte Anlagen sind noch nicht abschließend geklärt.
         - Das Verfahren zum manuellen Übertragen der Daten, wenn die Verkündung wiederholt nicht angenommen wurde, muss noch definiert werden.
-        - Sollte die Antwort auf eine erfolgreiche Anfrage weitere Informationen enthalten oder reicht der Statuscode?
         - Das Vorgehen und ggf. die Schnittstelle für die Übergangsphase, d.h. bevor die E-Gesetzgebung Daten aus dem LEA-Editor liefert, muss noch definiert werden.
-        - Eventuell wird es in Zukunft notwendig sein, statt einer synchronen Anfrage eine asynchrone Anfrage zu verwenden, damit das Backend größere Datenmengen besser verarbeiten kann. Es ist zu klären, ob wir die Schnittstelle direkt so gestalten wollen, oder ob das eine künftige Ausbaustufe sein wird.
-      security:
-        - ClientCredentials: [upload_verkuendung]
+        - Signierung:
+          - Art und Format der Signatur müssen noch definiert werden.
+          - Es muss entschieden werden, ob nur das ZIP-Archiv und/oder die Dateien im Archiv signiert werden.
+          - Das Verfahren zum Austausch des öffentlichen Schlüssels für die Verifizierung der Signatur muss noch definiert werden.
+        - Eventuell wird es in Zukunft notwendig sein, statt einer synchronen Anfrage eine asynchrone Anfrage zu verwenden, damit das Backend größere Datenmengen besser verarbeiten kann.
       requestBody:
         required: true
+        description: |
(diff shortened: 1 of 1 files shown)


doc/backend/norms-everkuendung-api-v1.yaml
```diff
@@ -0,0 +1,153 @@
...
+    description: Backend
+
+paths:
+  /verkuendungen:
+    post:
+      summary: Annahme von Verkündungen.
+      description: |
+        Die Schnittstelle erlaubt das Hochladen eines Normendokumentationspaketes bis zu einer Größe von 200mb. Das Dokumentationspaket liegt in Form eines ZIP-Archivs vor (`bgbl{1|2}_{jahr}_{nr}.zip`). Es enthält folgende Daten:
+
+        - Rechtsetzungsdokument (XML, PDF)
```
VictorDelCampo: Not sure if before the eGG is in place, if the e-Verkündung (or better said the structuring provider - Strukturierungsdienstleister) will also send the Rechtsetzungsdokument

doc/backend/norms-everkuendung-api-v1.yaml
```diff
@@ -0,0 +1,153 @@
...
+        ## Offene Fragen
+
+        - Die Dateiformate für nicht-strukturierte Anlagen sind noch nicht abschließend geklärt.
+        - Es ist noch offen, welchen Mehrwert die konsolidierte Fassung bringt und ob sich die Übermittelung lohnt.
+        - Art und Format der Signatur müssen noch definiert werden.
+        - Es muss entschieden werden, ob nur das ZIP-Archiv und/oder die Dateien im Archiv signiert werden.
+        - Das Verfahren zum Austausch des öffentlichen Schlüssels für die Verifizierung der Signatur muss noch definiert werden.
+        - Das Verfahren zum manuellen Übertragen der Daten, wenn die Verkündung wiederholt nicht angenommen wurde, muss noch definiert werden.
+        - Sollte die Antwort auf eine erfolgreiche Anfrage weitere Informationen enthalten oder reicht der Statuscode?
+        - Das Vorgehen und ggf. die Schnittstelle für die Übergangsphase, d.h. bevor die E-Gesetzgebung Daten aus dem LEA-Editor liefert, muss noch definiert werden.
```
VictorDelCampo, line 38: Oh now i see here this open question/topic. But if we offer the openAPI in the format of when eGG is in place (a.k.a. Rechtsetzungsdokument) we should then also deliver an openAPI with the content that we will be getting before the eGG is in place right?

doc/backend/norms-everkuendung-api-v1.yaml
```diff
@@ -0,0 +1,153 @@
...
+  title: RIS Norms API für die E-Verkündung
+  version: 1.0.0
+  description: |
+    API für die Datenübertragung von der E-Verkündung zu NeuRIS.
+
+    Über die Schnittstelle zwischen E-Verkündung und NeuRIS werden die Daten aus dem Bundesgesetzblatt I und II ausgetauscht.
+
+    ## Authentifizierung
+
+    Für die Nutzung aller Endpunkte muss ein gültiger Access Token als `Authorization: Bearer <token>` übermittelt werden. Der Token kann über einen [OAuth2 Client Credentials Grant](https://datatracker.ietf.org/doc/html/rfc6749#section-4.4) abgerufen werden. Die notwendigen Informationen (URL, Client Secret) werden vom DigitalService bereitgestellt.
```
VictorDelCampo: Missing also that client-ID will also be provided, together with the url and the client-secret?

doc/backend/norms-everkuendung-api-v1.yaml
```diff
@@ -0,0 +1,153 @@
...
+        - Es ist noch offen, welchen Mehrwert die konsolidierte Fassung bringt und ob sich die Übermittelung lohnt.
+        - Art und Format der Signatur müssen noch definiert werden.
+        - Es muss entschieden werden, ob nur das ZIP-Archiv und/oder die Dateien im Archiv signiert werden.
+        - Das Verfahren zum Austausch des öffentlichen Schlüssels für die Verifizierung der Signatur muss noch definiert werden.
+        - Das Verfahren zum manuellen Übertragen der Daten, wenn die Verkündung wiederholt nicht angenommen wurde, muss noch definiert werden.
+        - Sollte die Antwort auf eine erfolgreiche Anfrage weitere Informationen enthalten oder reicht der Statuscode?
+        - Das Vorgehen und ggf. die Schnittstelle für die Übergangsphase, d.h. bevor die E-Gesetzgebung Daten aus dem LEA-Editor liefert, muss noch definiert werden.
+        - Eventuell wird es in Zukunft notwendig sein, statt einer synchronen Anfrage eine asynchrone Anfrage zu verwenden, damit das Backend größere Datenmengen besser verarbeiten kann. Es ist zu klären, ob wir die Schnittstelle direkt so gestalten wollen, oder ob das eine künftige Ausbaustufe sein wird.
+      security:
```
VictorDelCampo: Should we apply the security at the root level (for all endpoints to come, if some new come) instead of at the endpoint level?

doc/backend/norms-everkuendung-api-v1.yaml
```diff
@@ -0,0 +1,153 @@
...
+        "200":
+          description: Die Verkündung wurde erfolgreich an NeuRIS übermittelt.
+        "400":
+          description: |
+            Die Eingangsdaten sind fehlerhaft. Es kann sich um verschiedene Fehler handeln:
+
+            - Das ZIP-Archiv ist ungültig
+            - Die Signatur des Archivs fehlt oder ist ungültig
+            - Die Signatur der Dateien im Archiv fehlt oder ist ungültig
+        "401":
```
VictorDelCampo, line 84: Do we also need a 403 in case the needed scope/role is missing?

doc/backend/norms-everkuendung-api-v1.yaml
```diff
@@ -0,0 +1,153 @@
...
+                  format: binary
+                  description: ZIP-Archiv mit den Dokumenten
+                signature:
+                  type: string
+                  format: binary
+                  description: Digitale Signatur für das Archiv im PKCS#7-Format (.p7s)
+      responses:
+        "200":
+          description: Die Verkündung wurde erfolgreich an NeuRIS übermittelt.
+        "400":
```
VictorDelCampo, line 72: For 400 is the response content missing, meaning no ProblemDetail?

doc/backend/norms-everkuendung-api-v1.yaml
```diff
@@ -0,0 +1,153 @@
...
+          description: |
+            Ungültige Eingangsdaten:
+
+            - Die XML-Dokumente weisen entweder XSD- oder Schematron-Fehler auf. Schematron-Warnungen führen nicht zu einem Fehler.
+            - Die XML-Dokumente liegen in einer nicht unterstützten Version von LegalDocML.de vor.
+          content:
+            application/problem+json:
+              schema:
+                $ref: "#/components/schemas/ProblemDetail"
+        "500":
```
VictorDelCampo, line 117: Should we also add the `Retry-After` header for the response of the 500 error? Including info written in prosa before?

doc/backend/norms-everkuendung-api-v1.yaml
```diff
@@ -0,0 +1,153 @@
...
+          example: 413
+        detail:
+          type: string
+          description: Menschenlesbare Details zur Ursache des Problems.  Diese Information ist zum Debugging gedacht und sollte nicht in einer Benutzeroberfläche verwendet werden.
+          example: Die Eingangsdaten haben eine Größe von 220mb, maximal erlaubt sind 200mb.
+        instance:
+          type: string
+          format: uri
+          description: Der API-Aufruf, der den Fehler ausgelöst hat.
+          example: /api/v1/announcements
```
VictorDelCampo: copy&pasta :) should be /api/v1/verkuendungen

doc/backend/norms-everkuendung-api-v1.yaml
```diff
@@ -0,0 +1,153 @@
...
+        "500":
+          description: Interner Server-Fehler.
+          content:
+            application/problem+json:
+              schema:
+                $ref: "#/components/schemas/ProblemDetail"
+
+components:
+  securitySchemes:
+    BearerAuth:
```
VictorDelCampo: Why do we need this? The flow used is the client credentials which implies that a token is used when communicating with the API right? Because right now two different schemas of security are defined but only the client-credentials is "used" directly in the API endpoint of /verkuendungen

doc/backend/norms-everkuendung-api-v1.yaml
```diff
@@ -0,0 +1,153 @@
...
+                file:
+                  type: string
+                  format: binary
+                  description: ZIP-Archiv mit den Dokumenten
+                signature:
+                  type: string
+                  format: binary
+                  description: Digitale Signatur für das Archiv im PKCS#7-Format (.p7s)
+      responses:
+        "200":
```
malte-laukoetter: I would have expected this to be 201 - Created

doc/backend/norms-everkuendung-api-v1.yaml
```diff
@@ -0,0 +1,153 @@
...
+            application/problem+json:
+              schema:
+                $ref: "#/components/schemas/ProblemDetail"
+        "413":
+          description: Die Eingangsdaten überschreiten die maximale Dateigröße von 200mb.
+          content:
+            application/problem+json:
+              schema:
+                $ref: "#/components/schemas/ProblemDetail"
+        "422":
```
malte-laukoetter: I'm not sure how we differentiate which errors should return 400 and which 422. Should we just merge those two errors to simplify it for us? Thats also what the zalando api guidelines recommend: https://opensource.zalando.com/restful-api-guidelines/#status-code-422

doc/backend/norms-everkuendung-api-v1.yaml
```diff
@@ -0,0 +1,153 @@
+openapi: 3.0.0
+info:
+  title: RIS Norms API für die E-Verkündung
+  version: 1.0.0
+  description: |
```
malte-laukoetter, line 5: Should we mention something about rate-limiting or how many requests they are allowed to send at once?

doc/backend/norms-everkuendung-api-v1.yaml
```diff
@@ -0,0 +1,153 @@
...
+    post:
+      summary: Annahme von Verkündungen.
+      description: |
+        Die Schnittstelle erlaubt das Hochladen eines Normendokumentationspaketes bis zu einer Größe von 200mb. Das Dokumentationspaket liegt in Form eines ZIP-Archivs vor (`bgbl{1|2}_{jahr}_{nr}.zip`). Es enthält folgende Daten:
+
+        - Rechtsetzungsdokument (XML, PDF)
+        - Regelungstext (XML)
+        - soweit vorhanden: Strukturierte Anlage(n) (XML)
+        - soweit vorhanden: Nicht-strukturierte Anlagen, z.B. eingebundene Bilder, PDF-Formulare (etwa: PDF, JPG, PNG)
+        - optional: konsolidierte Fassung aus der E-Gesetzgebung (PDF)
```
malte-laukoetter: Shouldn't this be part of the description of the request body?

doc/backend/norms-everkuendung-api-v1.yaml
```diff
@@ -0,0 +1,153 @@
...
+
+paths:
+  /verkuendungen:
+    post:
+      summary: Annahme von Verkündungen.
+      description: |
+        Die Schnittstelle erlaubt das Hochladen eines Normendokumentationspaketes bis zu einer Größe von 200mb. Das Dokumentationspaket liegt in Form eines ZIP-Archivs vor (`bgbl{1|2}_{jahr}_{nr}.zip`). Es enthält folgende Daten:
+
+        - Rechtsetzungsdokument (XML, PDF)
+        - Regelungstext (XML)
```
malte-laukoetter: also do we want to enforce that the naming of the files is consistent with what we expect? eg the regelungstext files are always named `regelungstext-{n}.xml` etc.?

doc/backend/norms-everkuendung-api-v1.yaml
```diff
@@ -0,0 +1,153 @@
...
+        - Es ist noch offen, welchen Mehrwert die konsolidierte Fassung bringt und ob sich die Übermittelung lohnt.
+        - Art und Format der Signatur müssen noch definiert werden.
+        - Es muss entschieden werden, ob nur das ZIP-Archiv und/oder die Dateien im Archiv signiert werden.
+        - Das Verfahren zum Austausch des öffentlichen Schlüssels für die Verifizierung der Signatur muss noch definiert werden.
+        - Das Verfahren zum manuellen Übertragen der Daten, wenn die Verkündung wiederholt nicht angenommen wurde, muss noch definiert werden.
+        - Sollte die Antwort auf eine erfolgreiche Anfrage weitere Informationen enthalten oder reicht der Statuscode?
+        - Das Vorgehen und ggf. die Schnittstelle für die Übergangsphase, d.h. bevor die E-Gesetzgebung Daten aus dem LEA-Editor liefert, muss noch definiert werden.
+        - Eventuell wird es in Zukunft notwendig sein, statt einer synchronen Anfrage eine asynchrone Anfrage zu verwenden, damit das Backend größere Datenmengen besser verarbeiten kann. Es ist zu klären, ob wir die Schnittstelle direkt so gestalten wollen, oder ob das eine künftige Ausbaustufe sein wird.
```
malte-laukoetter: do you already have ideas on what changes that would include? E.g. is it a lot more work for us to directly design the api to support that without needing to do more coordination with the e-verkündung. (Not implementing the async processing immediately but that we could just do it without changing the api)

doc/backend/norms-everkuendung-api-v1.yaml
andreasphil: Had that before the oauth scheme and forgot to delete it, thanks!

doc/backend/norms-everkuendung-api-v1.yaml
andreasphil: In the original proposal, both were in 400. I split them in two; in my head the difference is

- 400: File is corrupted or determined not to be trustworthy based on the signature
- 422: File and signature are correct, but the documents are not valid

Does that distinction make sense for you or would you consider both 422? I don't have strong feelings about this either way.

doc/backend/norms-everkuendung-api-v1.yaml
andreasphil: I was wondering to what extent we even require a specific filename format. Why do we care? Isn't all the relevant information inside the bundle anyways?

doc/backend/norms-everkuendung-api-v1.yaml
andreasphil: Yes, but let's discuss separately. I will send you and Victor an invite to discuss some of the more complicated questions.

doc/backend/norms-everkuendung-api-v1.yaml
andreasphil: I added a 429 indicating rate limiting, but left the specific conditions a bit vague for now. Let's discuss—depends also on whether the API will be sync or async.

PR #1040, opened 2025-02-25
Remove link between Announcements and Releases

malte-laukoetter:


backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/AnnouncementService.java
```diff
@@ -186,20 +185,7 @@ private void deleteAnnouncement(NormExpressionEli expressionEli) {
...
-        .flatMap(Collection::stream)
-        .forEach(norm ->
-          deleteNormPort.deleteNorm(
-            new DeleteNormPort.Command(
-              norm.getManifestationEli(),
-              NormPublishState.QUEUED_FOR_PUBLISH
-            )
-          )
-        );
+      // TODO: (Malte Laukötter, 2025-02-25) we originally deleted all releases and published norms from the announcement here as well should we do this again?
```
malte-laukoetter, line 188: this is still here as it will be touched again when removing the delta stuff from the upload

backend/src/main/resources/db/migration/V0.31__link_releases_to_norm_expressions.sql
```diff
@@ -0,0 +1 @@
+DROP TABLE announcement_releases;
```
malte-laukoetter: - [x] improve name of migration

PR #1048, opened 2025-02-28
Don't delete migrated documents that have already been published

andreasphil:


backend/src/main/resources/db/migration/V0.32__add_completed_column_to_migration_log_table.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE migration_log
+  ADD COLUMN IF NOT EXISTS completed boolean NOT NULL DEFAULT FALSE;
```
malte-laukoetter, line 2: should this column be set to true for the existing migration logs?

backend/src/main/java/de/bund/digitalservice/ris/norms/application/port/output/UpdateMigrationLogCompletedPort.java
```diff
@@ -0,0 +1,25 @@
+package de.bund.digitalservice.ris.norms.application.port.output;
+
+import de.bund.digitalservice.ris.norms.domain.entity.MigrationLog;
+import java.util.UUID;
+
+/**
+ * Port for updating the completion state of a migration log entry.
+ */
+public interface UpdateMigrationLogCompletedPort {
```
malte-laukoetter: `CompleteMigrationLogPort`? We only use it to set it to completed so we could keep the surface of the port smaller

backend/src/main/java/de/bund/digitalservice/ris/norms/application/port/output/UpdateMigrationLogCompletedPort.java
```diff
@@ -0,0 +1,25 @@
...
+  /**
+   * Updates the completion state of an existing {@link MigrationLog} entry.
+   *
+   * @param command Specifies the entry to update and the new value
+   * @return Number of affected entries
+   */
+  int updateMigrationLogCompleted(final Command command);
```
malte-laukoetter: could this return `void`? Or is the number of affected entries used anywhere? That seems to me like a implementation detail the consumer of the port doesn't need to know about

backend/src/main/java/de/bund/digitalservice/ris/norms/application/port/output/UpdateMigrationLogCompletedPort.java
```diff
@@ -0,0 +1,25 @@
...
+
+  /**
+   * Command for updating the completion state of a migration log entry.
+   *
+   * @param id The ID of the migration log entry to be updated
+   * @param completed The new completion state of the migration log entry
+   */
+  record Command(UUID id, boolean completed) {}
```
malte-laukoetter: If we change the port to only set it to completed we don't need the command anymore

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/output/database/repository/MigrationLogRepository.java
```diff
@@ -13,4 +17,9 @@
 @Repository
 public interface MigrationLogRepository extends JpaRepository<MigrationLogDto, UUID> {
   Optional<MigrationLogDto> findFirstByOrderByCreatedAtDesc();
+
+  @Modifying
+  @Transactional
+  @Query("UPDATE MigrationLogDto m SET m.completed = :completed WHERE m.id = :id")
+  int updateCompletedById(@Param("id") final UUID id, @Param("completed") final boolean completed);
```
malte-laukoetter, line 24: in all other cases we loaded the dto, modified it and then used `Repository::save` to save it again. This might actually be more performant to do it this way, but I'm not sure if that is worth introducing a different pattern

backend/src/main/resources/db/migration/V0.32__add_completed_column_to_migration_log_table.sql
VictorDelCampo: Should yes but not really important since those "old" migration logs won't be ever be picked up again it is fine.... but yeah we should do it "right" and update in the same flyway script also all older entries so that the new column is set to TRUE

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/output/database/repository/MigrationLogRepository.java
VictorDelCampo: Andreas implemented it in such a way at the beginning but told him why not doing it like this. I understand it is another "pattern" but in my opinion, for this very small entity and use case, it is better. Furthermore we had already load the entry from the db, transformed it into an object of the domain model and would load the same entry again from the db to just be able to called the "save" method? Sounds to me not nice to be honest. Also for the migration log we won't ever have the case that we update the whole object meaning other properties (size specially) since our app should only have "read" access and "write" access for the "completed" column. This way it aligns with the output port we have where we only pass the boolean value with the id for the entry

backend/src/main/resources/db/migration/V0.32__add_completed_column_to_migration_log_table.sql
andreasphil: @VictorDelCampo didn't we talk about this and decided it should be `FALSE` so new entries would have the correct value automatically? If we changed it to `TRUE` here then we would need to also update the migration importer script to explicitly set it `FALSE`, correct?

backend/src/main/java/de/bund/digitalservice/ris/norms/application/port/output/UpdateMigrationLogCompletedPort.java
andreasphil: It's currently only used in the tests. There is no other way to know whether the update was successful or not because Spring doesn't throw if the ID doesn't exist. However we don't use that information either way so I could remove it. What do you think?

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/output/database/repository/MigrationLogRepository.java
andreasphil: Agree with Victor here, I don't think it's an issue that it's slightly different but its a lot simpler.

backend/src/main/resources/db/migration/V0.32__add_completed_column_to_migration_log_table.sql
VictorDelCampo: @andreasphil yes it should be default to FALSE for new entries, but all old entries should be set to TRUE. Meaning we need a 2nd part of the same SQL, a 2nd statement like:
`UPDATE migration_log SET completed = true;`

backend/src/main/java/de/bund/digitalservice/ris/norms/application/port/output/UpdateMigrationLogCompletedPort.java
malte-laukoetter: what i did in the past to check that an update is successful is to load the entity again and check that it has the update data

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/output/database/repository/MigrationLogRepository.java
malte-laukoetter: fine with me :)

backend/src/main/java/de/bund/digitalservice/ris/norms/application/port/output/UpdateMigrationLogCompletedPort.java
VictorDelCampo: i agree with malte, this is an output port to set as "completed" a migration log, very specific case, meaning the implementation of the output port should do directly the setting to TRUE of that column, we dont need to pass the boolean value and we just need the UUID of the entry

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/PublishService.java
```diff
@@ -85,38 +88,59 @@ public void processQueuedFilesForPublish() {
       }
     });
 
-    loadLastMigrationLogPort
+    var lastMigrationLog = loadLastMigrationLogPort
```
VictorDelCampo, line 91: This filtering looks too complex, why not just:
`var lastMigrationLog = loadLastMigrationLogPort.loadLastMigrationLog().filter(f -> !f.isCompleted());`
Is it just for the logging? I think we dont need the info when a migration log was found that was already completed right?

backend/src/main/java/de/bund/digitalservice/ris/norms/application/port/output/UpdateMigrationLogCompletedPort.java
VictorDelCampo: Agree

backend/src/main/java/de/bund/digitalservice/ris/norms/application/port/output/UpdateMigrationLogCompletedPort.java
VictorDelCampo: For tests I also see that the good way, to really check if the entry on DB was successfully updated. I would also change the return type to void (even if in sql, which it is a implementation detail of db returns the number of updated rows). For us in the application we don't really need that value in my opinion

PR #1054, opened 2025-03-03
RISDEV-5306-Component-for-automatically-ordering-XML-elements-in-the-way-given-by-the-XSD

malte-laukoetter:


malte-laukoetter:
Sonar complains about duplicated code, but there is no real duplicated code, so i will just ignore that warning :)

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/LdmlDeElementSorter.java
```diff
@@ -0,0 +1,303 @@
...
+    var complexType = findDefinitionForSchemaType(typeName);
+
+    if (complexType.isEmpty()) {
+      elementOrderByType.put(typeName, List.of());
+      return List.of();
+    }
+
+    // Skip sorting for mixed content as there can be text nodes in between the sequence which we therefore cannot
+    // reliably reorder
+    if (complexType.get().getAttribute("mixed").equals("true")) {
```
SebastianRossa: Should we leave a javadocs comment in the public function that those types will not be sorted?

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/LdmlDeElementSorter.java
```diff
@@ -0,0 +1,303 @@
...
+      elementOrderByType.put(typeName, List.of());
+      return List.of();
+    }
+
+    var elements = NodeParser.getElementsFromExpression("./sequence/*", complexType.get());
+
+    var elementOrder = elements
+      .stream()
+      .flatMap(element -> {
+        if (Objects.equals(element.getTagName(), "xs:element")) {
```
SebastianRossa: Might this become a switch statement as well?

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/LdmlDeElementSorter.java
```diff
@@ -0,0 +1,303 @@
...
+            return Integer.MAX_VALUE;
+          }
+
+          return index;
+        })
+      )
+      .forEach(element::appendChild);
+  }
+
+  private List<String> findChildElementOrder(Element element) {
```
SebastianRossa, line 95: I think we could go more OOP here. While I think this class is relatively small and therefore could stay as is, it might be easier to understand when we move everything related to finding the ordering for a specific element to an Element class. The Element class could have a public getOrder() method. Inside this class all Element specific functionality could be encapsulated. This would mean all NodeParser call would be in this class similar to our other domain models.

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/LdmlDeElementSorter.java
```diff
@@ -0,0 +1,303 @@
...
+import org.w3c.dom.Document;
+import org.w3c.dom.Element;
+import org.w3c.dom.Node;
+
+/**
+ * Service for fixing the order of nodes within a LegalDocML.de element based on the xsd-Schema.
+ */
+public class LdmlDeElementSorter {
+
+  private final List<Document> schemas;
```
SebastianRossa, line 28: I think everything related to loading the xsd files could go into a repository. Thus making it interchangeable like loading from the internet, the database or some other place in the future.

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/LdmlDeElementSorter.java
malte-laukoetter: I tried around a bit but couldn't find a way to add some classes in here that still supported the multiple schema files and caching without making the code (at least in my eyes) more confusing. It would have split logic into many places and greatly reduced the locality of the logic. 

If we ever get more places that parse parts of the XSD-Schema for something we should probably reconsider this but for now I don't really see a way that adding classes for things in here increases the readability.

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/LdmlDeElementSorter.java
malte-laukoetter: i added a service for providing the schema files and also used it for loading the schemas in the LdmlDeValidator

PR #1055, opened 2025-03-04
Repo cleanup

andreasphil:
This PR performs various housekeeping in the repo with the goals of **reducing clutter, removing outdated information and files, and increasing clarity:**

- Updated the environment variable examples to include all relevant variables
- Removed the redundant top level `node-version` file: if I remember correctly this was needed due to a quirk in the setup of one developer, who is not in the team anymore. Please let me know if this causes issues for anyone.
- Merged German and English Codes of Conduct, to they can both be referenced at the same time and are consistent with our Contribution Guidelines.
- Removed the E2E test Docker setup, which was broken anyways and apparently not used by anyone in a long time
- Rename the `DockerfileApp` to a more standard `Dockerfile`

It also overhauls the READMEs, most importantly at the repo root, as well as the frontend and backend:

- Goals:
  - Quick reference for experienced Norms devies, who just want to look something up
  - Give people not familiar with the project an idea about how it works, without trying to be exhaustive
- Consider the system as a whole (frontend + backend + infra) instead of isolated modules
- Each piece of information should exist once and be referenced rather than duplicated (except for the quickstart)
- All outdated information removed


README.md
```diff
@@ -1,53 +1,170 @@
...
-Please refer to [`DEVELOPING.md`](./DEVELOPING.md) for further details.
+If you would like to contribute, check out [`CONTRIBUTING.md`](./CONTRIBUTING.md). Please also consider our [Code of Conduct](./CODE_OF_CONDUCT.md).
 
-# Contributions Welcome!
+## Additional Resources
 
-See [`CONTRIBUTING.md`](./CONTRIBUTING.md) for details.
+- 🔒 [RIS Reports](https://github.com/digitalservicebund/ris-reports) (additional documentation, including architecture diagrams and JavaDocs)
+- 🔒 [Infrastructure](https://github.com/digitalservicebund/ris-norms-infra)
+- 🔒 [Importer for migrated data](https://github.com/digitalservicebund/ris-norms-migration-import/)
```
malte-laukoetter, line 166: I'm not sure if we want to link them here as they are not public (or we should atleast add a comment about that so no one is confused)

malte-laukoetter (approved):
Thanks a lot for cleaning up the readme files! It's a lot cleaner and simpler to understand now :)

README.md
malte-laukoetter: oh that is what the lock is supposed to indicate, than don't mind my comment :)

VictorDelCampo:
@andreasphil since we don't have the nginx setup for a long time and you removed the docker file using it, you could also remove the `nginx-playwright.conf`

frontend/README.md
```diff
@@ -1,104 +1,128 @@
 # Frontend
 
-The frontend is the main entry point for users of _RIS norms_.
+The Norms frontend is a Single Page Application built with [TypeScript](https://www.typescriptlang.org/), [Vue](https://vuejs.org/), and [Tailwind](https://tailwindcss.com/). Most components come from [RIS UI](https://github.com/digitalservicebund/ris-ui), a component library shared with other NeuRIS projects. RIS UI is a theme for [PrimeVue](https://primevue.org/).
 
-# Development
+The frontend also includes our E2E tests.
 
-## Prerequisites
+Before diving into the code, please get familiar with our [code conventions](https://digitalservicebund.atlassian.net/wiki/x/BIC1N).
```
VictorDelCampo: This should also have the "lock" symbol because not publicly available?

monitoring/README.md
```diff
@@ -1,3 +1,5 @@
+# Monitoring
```
VictorDelCampo, line 1: Not really sure what do we have the `monitoring` module anymore ...

backend/README.md
```diff
@@ -1,123 +1,91 @@
-# Backend module
+# Backend
 
-Backend service built with Spring Boot.
+The Norms backend is a [Spring Boot](https://docs.spring.io/spring-boot/index.html) application build with Java. We use [Postgres](https://www.postgresql.org/) as our database and [Keycloak](https://www.keycloak.org/) as our identity provider. Both run locally using [Docker](https://www.docker.com/). In a live system (staging + production), the backend also serves the [frontend](../frontend/).
```
VictorDelCampo: missing UAT? :) when you say staging + production, though don't think relevant at all to say staging and production

backend/README.md
```diff
@@ -1,123 +1,91 @@
...
 
-License scanning is performed as part of the pipeline's `build` job. Whenever a production dependency
-is being added with a yet unknown license the build is going to fail.
+We aim to write JavaDoc comments to at least all public classes and methods. To fulfill this goal we use checkstyle, which is integrated not only into gradle but into lefthook (pre-commit).
 
-**To run a scan locally:**
+To run a check locally:
 
-```bash
-./gradlew checkLicense
```
VictorDelCampo, line 102: You remove the part about the backend license? Why?

backend/README.md
```diff
@@ -1,123 +1,91 @@
...
-
-**To run a check locally:**
-
-```bash
+```sh
 ./gradlew check
 ```
 
-## Check for outdated libraries
+## Check for Outdated Libraries
```
VictorDelCampo: I think i saw this "format" several times, why Outdated Libraries with capital letters at the beginning?

README.md
```diff
@@ -1,53 +1,170 @@
...
 
-  - the domain model
-  - the API specification and
-  - our Architecture Decision Records (ADR)
+## Slack Notifications
 
-- Additionally, there is the general [RIS Documentation](https://digitalservicebund.github.io/ris-reports/), which shows the RIS Norms
-  - [architecture diagrams](https://digitalservicebund.github.io/ris-reports/docs/architecture/diagrams_list.html)
-  - backend [JavaDocs](https://digitalservicebund.github.io/ris-reports/docs/backend-code-documentation/norms-java.html)
+Opt in to CI posting notifications for failing jobs to a particular Slack channel by setting a repository secret with the name `SLACK_WEBHOOK_URL`, containing a url for [incoming webhooks](https://api.slack.com/messaging/webhooks).
```
VictorDelCampo: Not sure about this part of slack in the global README and if at all in any readme :S

VictorDelCampo (commented):
Amazing clean up! Minor comments

README.md
andreasphil: Exactly! It's a bit weird because we're partly public and partly not, but since most people working with this are DS people who do have access to the resources I wanted to include them

backend/README.md
andreasphil: Good question, it's just a habit that I don't know when I picked it up but am 100% sure I have not invented. Had to look it up, it's [Associated Press style title case](https://en.wikipedia.org/wiki/Title_case#AP_Stylebook) (even though I don't know if I always use it correctly).

README.md
andreasphil: It was there before (in a different place) so I kept it, but wouldn't mind deleting it either. What do you prefer?

backend/README.md
andreasphil: I generally removed the docs about the license checker (for the frontend too) because I don't consider it a thing that people actively use but more of an automated check that runs in the pipeline. And I tried to not list every single possible command, only the most relevant for developing. Open to adding it back in if you disagree

backend/README.md
andreasphil: Was just intended as examples, will update it

monitoring/README.md
andreasphil: Do we still need/use it? Otherwise I can remove it. I assumed that is how our Grafana dashboards are created?

backend/README.md
andreasphil: Made the casing more consistent for all headings

backend/README.md
VictorDelCampo: Maybe just leave it there to create awareness about licensing? includes also the command that you would need to run locally if we face a problem in the pipeline. But don't have such a big strong opinion. you decide

README.md
VictorDelCampo: I would remove it cause i dont really understand why it should be there. It we had the info about what channel to follow i could maybe understand it (for new devs) but not this info about setting it up, cause it is anyway done in the settings of the GH repo

backend/README.md
andreasphil: There is a note about the licenses in the main README, it just doesn't include the specific command. Leaving it as it is

monitoring/README.md
VictorDelCampo: Will discuss this topic with Sebastian. You can close your PR if this was the only open topic

PR #1060, opened 2025-03-05
Migrate to Tailwind V4

hamo225:
RISDEV-6434


frontend/package.json
```diff
@@ -82,7 +82,7 @@
     "postcss": "~8.5.3",
     "prettier": "~3.5.3",
     "prettier-plugin-tailwindcss": "~0.6.11",
-    "tailwindcss": "~3.4.17",
+    "tailwindcss": "^4.0.7",
```
andreasphil: todo: should be "~" instead of "^"

frontend/src/components/RisLawPreview.vue
```diff
@@ -413,103 +413,122 @@ const slots: SetupContext["slots"] = useSlots()
 
 <style scoped>
 .preview-container:focus :deep(.focused) {
```
andreasphil, line 415: question: can't we use `@reference` for this? Same for the styles in the other components

frontend/src/components/RisLawPreview.vue
```diff
@@ -413,103 +413,122 @@ const slots: SetupContext["slots"] = useSlots()
...
 }
 
 .default-norm-style :deep(:is(table, thead, td)) {
-  @apply border border-blue-400;
+  border: 1px solid var(--color-blue-400);
 }
 
 .default-norm-style :deep(:is(thead, td:first-child)) {
-  @apply font-bold;
+  font-weight: 700;
```
andreasphil: todo: use the [variable for the font weight](https://tailwindcss.com/docs/font-weight#customizing-your-theme) instead. Same for the other usage of `font-weight` with different weights

frontend/src/views/amending-law/affected-documents/references/RisRefEditor.vue
```diff
@@ -55,12 +56,15 @@ const updateXmlSnippet = useDebounceFn((newXmlSnippet) => {
 
 const { refersTo, href } = useRef(xmlSnippet, updateXmlSnippet)
 
-const inputEl = ref<HTMLInputElement | null>(null)
+const inputEl = ref<ComponentPublicInstance | null>(null)
```
andreasphil, line 59: question: you sure this is correct? I thought it was supposed to be something like `InstanceType<typeof InputText>` but that might be outdated

frontend/src/views/amending-law/affected-documents/references/RisRefSelectionPanel.vue
```diff
@@ -181,14 +181,20 @@ function eidToSlotName(eid: string) {
 }
 
 :deep(.akn-ref) {
-  @apply bg-highlight-1-default px-2 outline outline-dotted outline-1 outline-blue-800;
+  background-color: var(--highlight-1-default);
+  padding-inline: var(--spacing-2);
+  outline: 1px dotted var(--color-blue-800);
```
andreasphil, line 186: question: is there a variable for the outline as a whole? If so please also update the outlines in the other components

frontend/src/views/amending-law/articles/editor/AmendingLawArticleEditor.view.vue
```diff
@@ -225,7 +225,7 @@ watch(loadXmlError, (err) => {
                 </div>
                 <RisLawPreview
                   v-else
-                  class="amendingLawPreview ds-textarea flex-grow p-2"
+                  class="amendingLawPreview ds-textarea grow p-2"
```
andreasphil: todo: remove `ds-textarea`, that class no longer exists with the removal of Angie. Also in the other cases where it's currenty used. Can you do a general check for classes prefixed with `ds-`? I think all of them can be removed

frontend/e2e/application/expressionMetadataElement.spec.ts
```diff
@@ -168,15 +168,9 @@ test.describe("metadata view", { tag: ["@RISDEV-6266"] }, () => {
...
-      artSnRadio = sharedPage.getByRole("radio", {
-        name: "SN - Stammnorm",
-      })
-      artAnRadio = sharedPage.getByRole("radio", {
-        name: "ÄN - Änderungsnorm",
-      })
-      artUnRadio = sharedPage.getByRole("radio", {
-        name: "ÜN - Übergangsnorm",
-      })
+      artSnRadio = sharedPage.getByTestId("radio-sn").getByRole("radio")
```
andreasphil: question: why are the Test IDs needed?

andreasphil (changes requested):

frontend/src/components/RisLawPreview.vue
hamo225: @ apply can be kept for tailwind classes but for custom ones not. Its no longer the recommended way. For better performance use CSS variables. No need any @ reference directive. no need to process these @ apply.

frontend/e2e/application/expressionMetadataElement.spec.ts
hamo225: solved by using input-id which is the case for prime vue.

PR #1068, opened 2025-03-07
RISDEV-6988-Cleanup-documents-for-prototype

SebastianRossa:


backend/src/test/java/de/bund/digitalservice/ris/norms/application/service/PrototypeCleanupServiceTest.java
```diff
@@ -0,0 +1,178 @@
...
+      .nodeListToList(
+        norm.getRegelungstext1().getMeta().getProprietary().get().getElement().getChildNodes()
+      )
+      .stream()
+      .filter(node -> node.getNodeType() == Node.ELEMENT_NODE)
+      .toList();
+
+    assertThat(proprietyChildNodes).hasSize(1);
+    assertThat(proprietyChildNodes.get(0).getNamespaceURI())
+      .isEqualTo("http://MetadatenRIS.LegalDocML.de/1.7.2/");
```
malte-laukoetter: i would use `Namespace.METADATEN_RIS.getNamespaceUri()` here as well so we don't need to touch this when updating the ldml version

backend/src/test/java/de/bund/digitalservice/ris/norms/application/service/PrototypeCleanupServiceTest.java
```diff
@@ -0,0 +1,178 @@
...
+import de.bund.digitalservice.ris.norms.domain.entity.Norm;
+import de.bund.digitalservice.ris.norms.utils.NodeParser;
+import java.util.List;
+import org.junit.jupiter.api.Test;
+import org.w3c.dom.Element;
+import org.w3c.dom.Node;
+
+class PrototypeCleanupServiceTest {
+
+  private final PrototypeCleanupService underTest = new PrototypeCleanupService();
```
malte-laukoetter: nitpick: in the other tests we have always named this by the name of the service (so eg. `prototypeCleanupService`)

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/PrototypeCleanupService.java
```diff
@@ -0,0 +1,112 @@
...
+import org.w3c.dom.Node;
+
+/**
+ * Service responsible for cleaning up all {@link Dokument}s contained in a {@link Norm} since in the prototype
+ * not all available data can be published due to legal obligations
+ */
+@Service
+public class PrototypeCleanupService {
+
+  private static final Set<String> SHOULD_STAY = Set.of(
```
malte-laukoetter: i was confused for a short moment what exactly `SHOULD_STAY` contains and what it will be used for. Maybe move this down to the one method that is using it? or give it another name?

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/PrototypeCleanupService.java
```diff
@@ -0,0 +1,112 @@
...
+
+    return norm;
+  }
+
+  private void cleanRisMetadata(Dokument dokument) {
+    Optional<Element> dsProprietary = NodeParser.getElementFromExpression(
+      "//Q{" + Namespace.METADATEN_RIS.getNamespaceUri() + "}legalDocML.de_metadaten",
+      dokument.getDocument()
+    );
+    if (dsProprietary.isPresent()) {
```
malte-laukoetter: nitpick: there is also `Optional::ifPresent` that could have been used here, or maybe an early return?

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/PortalPrototypePublishService.java
```diff
@@ -65,7 +68,7 @@ public void publishNormsToPortalPrototype() {
 
       // TODO: (Malte Laukötter, 2025-03-07) check if norm is allowed to be published
 
-      // TODO: (Malte Laukötter, 2025-03-07) clean up norm
+      prototypeCleanupService.clean(norm.get());
       // TODO: (Malte Laukötter, 2025-03-07) also run the cleanup from the normal PublishService
```
malte-laukoetter: we don't have a ticket for this todo item yet, as i wasn't sure if you would just do it in here as well, should i create one?

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/Metadata.java
```diff
@@ -43,7 +43,10 @@ public enum Metadata {
     "qualifizierteMehrheit",
     true
   ),
-  ORGANISATIONS_EINHEIT(Namespace.METADATEN_RIS, "./organisationsEinheit", "organisationsEinheit");
+  ORGANISATIONS_EINHEIT(Namespace.METADATEN_RIS, "./organisationsEinheit", "organisationsEinheit"),
+  ENTRY_INTO_FORCE_DATE(Namespace.METADATEN_RIS, "./entryIntoForce/@date", "entryIntoForce", true),
+  EXPIRY_DATE(Namespace.METADATEN_RIS, "./expiry/@date", "expiry", true),
```
malte-laukoetter: we will probably have a merge conflict here as we both added these lines :) Also I think the 3 parameter should be date for both of these as that is the attribute that should be read and `isAttribute` is true

malte-laukoetter (approved):
Looks good to me, i left some (mostly nitpicky) comments :)

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/PortalPrototypePublishService.java
VictorDelCampo: Should we integrate the clean-up service into `PortalPrototypePublishService` to have just 1 service with this "temporary" code? or do we wanna keep the clean-up separated for testing purposes?

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/Metadata.java
VictorDelCampo: Jep it should be date, as these metadata are attributes. Good catch!

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/PrototypeCleanupService.java
VictorDelCampo: I prefer personally to have them at the top of the class (like in the table of content service) but you can decide. 
About the naming what about "ALLOWED_METADATA" ?

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/PrototypeCleanupService.java
```diff
@@ -0,0 +1,112 @@
...
+   *   <li>Removes specific notes related to amendments.</li>
+   *   <li>Modifies lifecycle events so the real date of the event is hidden.</li>
+   * </ul>
+   *
+   * @param norm The {@link Norm} containing the documents to be cleaned.
+   * @return The cleaned {@link Norm} with updated metadata and lifecycle events.
+   */
+  public Norm clean(Norm norm) {
+    Set<Dokument> dokumente = norm.getDokumente();
+    for (Dokument dokument : dokumente) {
```
VictorDelCampo, line 47: Bit nitpicking.... what about using a forEach on the Set directly?

```
    norm.getDokumente().forEach(dokument -> {
        cleanRisMetadata(dokument);
        ...
    });
```

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/PrototypeCleanupService.java
VictorDelCampo: +1 for if present

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/PrototypeCleanupService.java
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  private void cleanRisMetadata(Dokument dokument) {
+    Optional<Element> dsProprietary = NodeParser.getElementFromExpression(
+      "//Q{" + Namespace.METADATEN_RIS.getNamespaceUri() + "}legalDocML.de_metadaten",
+      dokument.getDocument()
+    );
+    if (dsProprietary.isPresent()) {
+      String query = createXpathAllDsProprietaryExceptShouldStay();
+      var nodesToDelete = NodeParser.getNodesFromExpression(query, dsProprietary.get());
```
VictorDelCampo: Nitpicking again, i would make another foreach here and 1-liner:

```
 NodeParser.getNodesFromExpression(createXpathAllDsProprietaryExceptShouldStay(), dsProprietary)
            .forEach(node -> node.getParentNode().removeChild(node))
```

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/PrototypeCleanupService.java
```diff
@@ -0,0 +1,112 @@
...
+      element.setAttribute("date", "1001-01-01");
+    }
+  }
+
+  private void deleteMetaData(Dokument dokument, String namespaceUri) {
+    Optional<Element> proprietaryElement = NodeParser.getElementFromExpression(
+      "//Q{" + namespaceUri + "}legalDocML.de_metadaten",
+      dokument.getDocument()
+    );
+    proprietaryElement.ifPresent(element -> element.getParentNode().removeChild(element));
```
VictorDelCampo: would combine these two statements:
```
    NodeParser.getElementFromExpression(
        "//Q{" + namespaceUri + "}legalDocML.de_metadaten",
        dokument.getDocument()
    ).ifPresent(element -> element.getParentNode().removeChild(element));
```

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/PrototypeCleanupService.java
```diff
@@ -0,0 +1,112 @@
...
+    var nodesToDelete = NodeParser.getNodesFromExpression(query, metadataElement);
+    for (Node node : nodesToDelete) {
+      node.getParentNode().removeChild(node);
+    }
+  }
+
+  private void cleanLifecycleEvents(Dokument dokument) {
+    Element metadataElement = dokument.getMeta().getElement();
+    String query =
+      "//eventRef[not(@type='generation' and @refersTo='ausfertigung') and not(@type='generation' and @refersTo='inkrafttreten')]";
```
VictorDelCampo: Didn't you say something about only the first `ausfertigung` and the first `inkraftreten` can stay? Or is it exactly the combination of generation+ausfertigung and generation+inkraftreten ? These combinations can only happen once each?

VictorDelCampo (approved):
LGTM. Also like Malte just being a nitpicker. And maybe also a comment: you went to xpath querying instead of using our domain classes (eventRef for example). Was this a deliberated decision? I guess you wanted to avoid relying on the domain since this service will be "temporary", if so, GOOD

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/PortalPrototypePublishService.java
SebastianRossa: @VictorDelCampo I prefer small concise classes. It's just easier for me to grasp what is happening there. Having all the cleanup methods in this class would distract me from the higher abstraction of this publishing service. And yes: testing was a breeze :-)

SebastianRossa: @malte-laukoetter i wasn't sure why we would have the cleanup step in the"normal" PublishService. Can you please clarify?

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/Metadata.java
SebastianRossa: Thanks a lot for clarifying this. I thought the actual tag would go in here and wasn't aware this changes when isAttribute is true. 
Looking into how getTag is used I get why we have it anyways. Should we add a small comment how this works?
E.g. in the Metadata docs:  *   <li>A tag name that points to the metadata value. This can be the actual tag name or an attribute depending on where the value is stored</li>

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/PrototypeCleanupService.java
SebastianRossa: What advantage do you see in using forEach? 
From my point of view the readability is the same while a for loop is faster (performance)

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/PortalPrototypePublishService.java
malte-laukoetter: It is also filtering out data from the xmls that we are not allowed to publish (eg. metadata that should only be available to the private portal). I would want to make sure that that definitely happens and not rely on the data in the database being already cleaned up (which might be the case i'm not 100% sure). And as that cleanup might remove different things than this new cleanup service we should call both.

malte-laukoetter: i created a subtask for it :)

VictorDelCampo: Not sure if we need to also call the cleanup of the publish service... since that clean up is bound to that job. And we have a separated clean up for the prototyp job. I think they are not related

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/PrototypeCleanupService.java
VictorDelCampo: To be honest, just readability in this case. But it is just my taste. I agree with the for loop being faster and if we don't have any need of additional functionality provided by the `forEach` (like parallel execution) we should stick to the for loop, specially if we dont have big collections, like the case for our set of `Dokument`. Let's keep then your for loop :)

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/PortalPrototypePublishService.java
malte-laukoetter: Both add restrictions on what data can be available to the public, the stuff from the existing publish service may never be public and the cleanup here is just the data that can not be available for the prototype. So to ensure we do not publish data that is not supposed to be public we need to run both. My idea is to move the cleanup from the publish service into it's own service and then call it here as well.

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/PrototypeCleanupService.java
SebastianRossa: As I understand generation/ausfertigung can only happen once. Then amendment/ausfertigung can occur. A Neufassung would be generation/neufassung. 
Not sure with the generation/inkrafttreten though. That might happen twice as i see it. Once at the former version and than on Neufassung. But i can't find any example.

malte-laukoetter:
One more thing i just remembered, I add some integration tests for the whole prototype publishing service in https://github.com/digitalservicebund/ris-norms/pull/1076/ it would be cool if you could add in a check that the published norm is cleaned up after we have merged the two PRs :)

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/PortalPrototypePublishService.java
VictorDelCampo: I see your reasoning and you are logically right.. I don't think it is good though to intertwine both clean-ups, since the jobs (and services including the clean-ups steps) should be independent. Also the clean-up for the prototype should take into consideration what should not be public anyway (organisationsEinheit), meaning the definition of what does not belong in the XML should already contain that. We are doing now work a devs that is not expected/specified

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/PrototypeCleanupService.java
```diff
@@ -0,0 +1,112 @@
...
+    Set<Dokument> dokumente = norm.getDokumente();
+    for (Dokument dokument : dokumente) {
+      cleanRisMetadata(dokument);
+      cleanRegularMetaData(dokument);
+      cleanBundesRegierungMetaData(dokument);
+      cleanNotes(dokument);
+      cleanLifecycleEvents(dokument);
+    }
+
+    return norm;
```
malte-laukoetter: one last thought: as this method is modifying the norm in place I think it would be a bit simpler to understand if we do not return it here

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/PrototypeCleanupService.java
VictorDelCampo: So what about laws with several inkraftrete when they are announced/generated? they would have several eventRef with the combination generation/inkraftreten right?

VictorDelCampo: Page 56 of LDML.de 1.7.2:

_inkrafttreten – Inkrafttreten eines Dokuments bzw. eines Teils davon oder eines auf das
Dokument anzuwendenden Änderungsbefehls._

VictorDelCampo: In page 114 you can see examples of several eventRef with generation/inkraftreten, this is the case when you have split time boundaries (gespaltene Geltungszeit), meaning one law that when announced already specified different inktraftreten

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/PortalPrototypePublishService.java
malte-laukoetter: I already see us adjusting the cleanup for publishing and then forgetting to also adjust the prototype-publishing. As yes in the moment all the non-public fields are in the metadata sections we clean up, but this might change and we would now always need to keep in mind both places.
My understanding of the prototype publishing is that all requirements of the "normal" publishing also apply here, as it is "just" normal publishing with some **additional** restrictions on which norms to publish and what data is allowed to be included.

malte-laukoetter: https://github.com/digitalservicebund/ris-norms/pull/1082

VictorDelCampo: But Malte.. this is once again your understanding and not a explicit requirement. We are doing now again things only on the dev side, without syncing with PM. If it is needed, let make it visible and gather first input from PM...

PR #1076, opened 2025-03-10
RISDEV-6990-Create-filter-step-only-those-in-the-GII-list-and-those-that-are-valid

malte-laukoetter:


backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/scheduler/PortalPrototypePublishingScheduler.java
```diff
@@ -16,7 +16,7 @@
  */
 @Component
 @Slf4j
-@Profile({ "production" })
+@ConditionalOnProperty("publish.portal-prototype.enabled")
```
VictorDelCampo, line 19: What are your arguments to change this? Two thoughts:
1. Do we know that we may need to disable this job on-demand? This would be a good argument for using such a property. Nevertheless I think we are deviating of the usage we had till now in regards of activating something depending on environments
2. Should we had a default value here? I think spring set "missing" property for conditionals to `false` but not 100% sure. Should we for visibility just use the `matchIfMissing = false` ?

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/output/resource/ResourceService.java
```diff
@@ -0,0 +1,44 @@
...
+  public ResourceService(
+    @Value(
+      "classpath:/portal-prototype-publishing-allowlist.txt"
+    ) Resource portalPrototypePublishingAllowlist
+  ) {
+    try {
+      this.publishingAllowlist.addAll(
+          Files.readAllLines(portalPrototypePublishingAllowlist.getFile().toPath())
+        );
+    } catch (IOException e) {
```
VictorDelCampo: I think we should raise an exception if we were not able to load the resource which will happen on application start-up. Otherwise we wouldn't realize it and the next prototype job run would not filter by the abbreviations and we would be possibly publishing not-allowed-norms.
What about not catching the `IOException` and just propagating it?

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/output/resource/ResourceService.java
```diff
@@ -0,0 +1,44 @@
...
+import org.springframework.stereotype.Service;
+
+/**
+ * Service for loading data from classpath resources.
+ */
+@Service
+@Log4j2
+public class ResourceService implements LoadPortalPublishingAllowListPort {
+
+  private final List<String> publishingAllowlist = new ArrayList<>();
```
VictorDelCampo: If this list should not be modified after it has been created on start-up in the constructor a.k.a be immutable we should use something else, specially not and `ArrayList`. 

And what `Files.readAllLines()` is returning is also a mutable list.

What about wrapping the result of `Files.readAllLines()` into an immutable list like `List.copyOf() `?

backend/src/main/java/de/bund/digitalservice/ris/norms/config/PortalPrototypeBucketServiceConfiguration.java
```diff
@@ -22,7 +22,7 @@
  * </ul>
  */
 @Configuration
-@Profile({ "production", "local", "local-docker" })
+@ConditionalOnProperty("publish.portal-prototype.enabled")
```
VictorDelCampo, line 25: Oh wait now we also don't need this in local and local-docker? I forgot what was the reason..

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/Proprietary.java
```diff
@@ -195,6 +195,22 @@ public void setRessort(final String newValue, final LocalDate date) {
       );
   }
 
+  /**
+   * Gets the date the expression enters into force
+   * @return the date of the entry into force
+   */
+  public Optional<LocalDate> getInkrafttreteDatum() {
```
VictorDelCampo, line 202: Not sure if these two methods belong to out domain model. I know reusing code for getting the metadata value but does this logic belong to our domain which represents the LDML.de standard? I understood that we filter this way because migration is putting that data there but later on if we need such a logic of filtering for another reason, will be also accessing the RIS metadata? Or rather using the lifecycle?

VictorDelCampo (changes requested):
Very nice structure code Malte, specially I love the introduction of the `ResourceService`, following the hexagonal architecture. Top!

I just made some comments, some of them should be tackled and other are just thoughts. Let me know if you need a discussion

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/scheduler/PortalPrototypePublishingScheduler.java
malte-laukoetter: With using a profile it was hard to be able to create an integration test as that test would then need to use that profile, but that created other problems. In this way I was able to just enable this feature in the integration test without also taking eg. the prod database config.

The default for a missing property is false, and that is also what i would expect when i would be reading that code without any knowledge about that annotation so I think it's clear enough this way?

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/output/resource/ResourceService.java
malte-laukoetter: The fallback is an empty list and in that case we will not publish anything.
But yes that exception handling is not really needed here, ill remove it (and just propagate the exception) :)

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/Proprietary.java
malte-laukoetter: In my understanding the custom metadata fields are also part of our domain. Also if I remember correctly the plan is to move the metadata from the neuris-specific block to the standard fields once they are in the spec.

So once the lifecycle data is garanteed to be correct we should be able to "just" change these 2 methods to load the dates from the lifecycle and no longer need those metadata fields to get the correct dates. Now that I'm thinking about it these methods might actually be better placed in either the metadata class or even the norm class 🤔

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/output/resource/ResourceService.java
VictorDelCampo: Oh you are right, having an empty list would lead to publishing nothing.. but yeah we should know when this happened, otherwise we won't be updating the data on the buckets ever

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/Proprietary.java
VictorDelCampo: I am just asking myself why are we using those migration metadata to filter out data and not just using the lifecycle events? Didn't fully check it but isn't the lifecycle containing also that information and is exactly for that? But if it is not the case which would surprise me, yes lets keep this logic in our domain model

malte-laukoetter: it seems that the data is available in the lifecycle data as well (atleast for the two norms I checked)

`refersTo="ausserkrafttreten" source="attributsemantik-noch-undefiniert" type="generation"`
and
`refersTo="inkrafttreten" source="attributsemantik-noch-undefiniert" type="generation"`

As the ticket mentions the two metadata fields we used here I'll check with Christian if it is ok if we use the lifecycle data instead :)

malte-laukoetter: we will for now use the `ris:` metadata fields as they might handle some edgecases.

PR #1085, opened 2025-03-13
Add import timestamp to announcements

andreasphil:
Implements RISDEV-7047. For now I decided to set the timestamp purely on the DB level, i.e. it is set automatically to the current time when the announcement is created. The application can read it, but it can not set the value on insertion or change it afterwards. This should be all we need and minimizes code/API surface.

Coincidentally also implements RISDEV-7048, but I added an integration test to verify that the timestamp is set correctly.

### DIFF
--- a/backend/src/test/java/de/bund/digitalservice/ris/norms/integration/adapter/input/restapi/AnnouncementControllerIntegrationTest.java
+++ b/backend/src/test/java/de/bund/digitalservice/ris/norms/integration/adapter/input/restapi/AnnouncementControllerIntegrationTest.java
@@ -154,7 +154,7 @@ class AnnouncementControllerIntegrationTest extends BaseIntegrationTest {
         "eli/bund/bgbl-1/2017/s419/2017-03-15/1/deu"
       );
       assertThat(announcement.get().getImportTimestamp())
-        .isCloseTo(Instant.now(), new TemporalUnitWithinOffset(1, ChronoUnit.HOURS));
+        .isCloseTo(Instant.now(), new TemporalUnitWithinOffset(5, ChronoUnit.MINUTES));
     }
 
     @Test
--- a/backend/src/test/java/de/bund/digitalservice/ris/norms/integration/adapter/input/restapi/AnnouncementControllerIntegrationTest.java
+++ b/backend/src/test/java/de/bund/digitalservice/ris/norms/integration/adapter/input/restapi/AnnouncementControllerIntegrationTest.java
@@ -15,7 +15,10 @@ import de.bund.digitalservice.ris.norms.domain.entity.eli.NormExpressionEli;
 import de.bund.digitalservice.ris.norms.integration.BaseIntegrationTest;
 import de.bund.digitalservice.ris.norms.utils.XmlMapper;
 import java.io.ByteArrayInputStream;
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
 import java.util.UUID;
+import org.assertj.core.data.TemporalUnitWithinOffset;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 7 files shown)


backend/src/main/resources/db/migration/V0.33__add_timestemp_to_announcements_table.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE announcements
```
VictorDelCampo: type in filename, timestemp --> timestamp

backend/src/test/java/de/bund/digitalservice/ris/norms/integration/adapter/output/database/DBServiceIntegrationTest.java
```diff
@@ -274,7 +274,11 @@ void itFindsAnnouncementOnDB() {
     );
 
     // Then
-    assertThat(announcementOptional).isPresent().contains(announcement);
+    assertThat(announcementOptional)
+      .get()
+      .usingRecursiveComparison()
+      .ignoringFields("importTimestamp")
```
VictorDelCampo, line 282: Even if we are not setting the timestamp in the application and letting the db create that value I would include a "checking" to some level in an integration test, like maybe just checking that the value is not empty? Or even checking if the value is within the range we would expact (minutes wise?)

backend/src/test/java/de/bund/digitalservice/ris/norms/integration/adapter/output/database/DBServiceIntegrationTest.java
```diff
@@ -302,7 +306,9 @@ void itLoadsAllAnnouncementsFromDB() {
     final List<Announcement> announcements = dbService.loadAllAnnouncements();
 
     // Then
-    assertThat(announcements).containsExactly(announcement2, announcement1);
+    assertThat(announcements)
+      .usingRecursiveFieldByFieldElementComparatorIgnoringFields("importTimestamp")
```
VictorDelCampo, line 315: Same here as the comment above, though if you check it in the other test, maybe is not needed here

backend/src/test/java/de/bund/digitalservice/ris/norms/integration/adapter/input/restapi/AnnouncementControllerIntegrationTest.java
```diff
@@ -127,6 +130,33 @@ void itCreatesANewAnnouncement() throws Exception {
         .isPresent();
     }
 
+    @Test
+    void itStoresTheImportTimestamp() throws Exception {
```
VictorDelCampo, line 134: Not sure why a new test (even integration) is needed in the adapter layer without having touched yet the adapter layer... meaning we should the controller is not returning the timestamp yet because the mapper between domain and controller hasn't been touched yet.

VictorDelCampo (commented):
Good work, just some comments and possibly the one about the controller test leads to more work on the same branch? I though the ticket would be to develop the whole flow including the API. If not, new ticket needed and i wouldn't include the controller test in this work.

backend/src/test/java/de/bund/digitalservice/ris/norms/integration/adapter/input/restapi/AnnouncementControllerIntegrationTest.java
andreasphil: I added it because of the way we sliced the ticket. "RISDEV-7048 Extend use case for creating announcements to store the timestamp" is automatically solved by setting the timestamp on the DB level, so nothing to do here. But it's also still expected behavior of the import that the timestamp exists, hence the test.

Actually doing something with the value and returning it in the controller is part of another ticket. Will leave it in for now, especially since otherwise it would be added in the next step anyway.

VictorDelCampo: Yeah but the test would be then something else, since for now testing the timestamp you are retrieving the object directly through the repository. When you adapt the mapper from domain to api response schema you will just test the response schema of the API call in the test

PR #1086, opened 2025-03-13
Implement the Announcement details page 

hamo225:
RISDEV-7045
RISDEV-7046


frontend/src/components/controls/RisHeader.vue
```diff
@@ -219,7 +219,7 @@ export function useHeaderContext() {
     v-bind="$attrs"
   >
     <section class="flex items-center">
-      <Breadcrumb :model="allBreadcrumbs">
+      <Breadcrumb :model="allBreadcrumbs" aria-label="Breadcrumb-Navigation">
```
andreasphil: todo: as discussed, this should not be needed and can be removed

frontend/src/components/RisLawPreview.vue
```diff
@@ -41,13 +41,19 @@ const props = withDefaults(
      * Should the default styling for norms be applied? Default: true
      */
     styled?: boolean
+
```
andreasphil: todo: as discussed, this should not be needed and can be removed. Use `:arrow-focus="false"` instead

andreasphil (changes requested):

PR #1087, opened 2025-03-13
API for announcement detail screen

andreasphil:
RISDEV-7049

### DIFF
--- a/backend/src/test/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/mapper/VerkuendungResponseMapperTest.java
+++ b/backend/src/test/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/mapper/VerkuendungResponseMapperTest.java
@@ -3,9 +3,11 @@ package de.bund.digitalservice.ris.norms.adapter.input.restapi.mapper;
 import static org.assertj.core.api.Assertions.assertThat;
 
 import de.bund.digitalservice.ris.norms.adapter.input.restapi.schema.VerkuendungResponseSchema;
+import de.bund.digitalservice.ris.norms.domain.entity.Announcement;
 import de.bund.digitalservice.ris.norms.domain.entity.Norm;
 import de.bund.digitalservice.ris.norms.domain.entity.Regelungstext;
 import de.bund.digitalservice.ris.norms.utils.XmlMapper;
+import java.time.Instant;
 import java.util.Set;
 import org.junit.jupiter.api.Test;
 
@@ -86,8 +88,17 @@ class VerkuendungResponseMapperTest {
       )
       .build();
 
+    var announcement = Announcement
+      .builder()
+      .eli(norm.getExpressionEli())
+      .importTimestamp(Instant.parse("2025-03-13T16:00:00Z"))
+      .build();
+
     // When
-    final VerkuendungResponseSchema result = VerkuendungResponseMapper.fromUseCaseData(norm);
+    final VerkuendungResponseSchema result = VerkuendungResponseMapper.fromAnnouncedNorm(
+      announcement,
+      norm
+    );
 
     // Then
     assertThat(result.getEli())
@@ -98,6 +109,7 @@ class VerkuendungResponseMapperTest {
     assertThat(result.getFrbrNumber()).isEqualTo("s593");
     assertThat(result.getFrbrDateVerkuendung()).isEqualTo("1964-08-05");
     assertThat(result.getFna()).isEqualTo("754-28-1");
+    assertThat(result.getImportedAt()).isEqualTo(announcement.getImportTimestamp());
   }
 
   @Test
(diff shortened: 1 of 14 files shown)


backend/src/test/java/de/bund/digitalservice/ris/norms/integration/adapter/input/restapi/AnnouncementControllerIntegrationTest.java
```diff
@@ -119,7 +119,8 @@ void itCreatesANewAnnouncement() throws Exception {
         .andExpect(status().isOk())
         .andExpect(
           jsonPath("eli", equalTo("eli/bund/bgbl-1/2017/s419/2017-03-15/1/deu/regelungstext-1"))
-        );
+        )
+        .andExpect(jsonPath("importedAt").exists());
```
andreasphil, line 123: To whoever reviews this: I am not sure if this is sufficient for testing, don't we also want to check that it's close to the current date? (Close because the date is set by the DB and we can't know the exact result down to the second). So currently the functionality is tested like this:

- This test checks that the date is returned at all
- The test below (`itStoresTheImportTimestamp`) checks that the timestamp is created correctly, although @VictorDelCampo wasn't happy with this test in an earlier review :D

backend/src/test/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/mapper/VerkuendungResponseMapperTest.java
```diff
@@ -0,0 +1,195 @@
...
+                             <akn:docTitle eId="einleitung-1_doktitel-1_text-1_doctitel-1" GUID="e08874b2-05a8-4d6e-9d78-7be24380c54b">Gesetz zum ersten Teil der Reform des Nachrichtendienstrechts</akn:docTitle>
+                          </akn:p>
+                       </akn:longTitle>
+                       <akn:block eId="einleitung-1_block-1" GUID="a0973d49-d628-42f7-a1da-b004bc980a44" name="attributsemantik-noch-undefiniert">
+                          <akn:date eId="einleitung-1_block-1_datum-1" GUID="f20d437a-3058-4747-8b8b-9b1e06c17273" refersTo="ausfertigung-datum" date="2023-12-29">Vom
+                             29.12.2023</akn:date>
+                       </akn:block>
+                    </akn:preface>
+                 </akn:act>
+              </akn:akomaNtoso>
```
malte-laukoetter, line 166: same here

backend/src/test/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/mapper/VerkuendungResponseMapperTest.java
```diff
@@ -0,0 +1,195 @@
...
+                           <akn:shortTitle eId="einleitung-1_doktitel-1_text-1_kurztitel-1" GUID="fdb8ed28-2e1f-4d81-b780-846fd9ecb716">( <akn:inline
+                                 eId="einleitung-1_doktitel-1_text-1_kurztitel-1_inline-1" GUID="bdff7240-266e-4ff3-b311-60342bd1afa2" refersTo="amtliche-abkuerzung" name="attributsemantik-noch-undefiniert">Vereinsgesetz</akn:inline>)</akn:shortTitle>
+                        </akn:p>
+                     </akn:longTitle>
+                     <akn:block eId="einleitung-1_block-1" GUID="010d9df0-817a-49b6-a121-d0a1d412a3e3" name="attributsemantik-noch-undefiniert">
+                        <akn:date eId="einleitung-1_block-1_datum-1" GUID="28fafbe4-403d-4436-8d0d-7241cbbdade0" refersTo="ausfertigung-datum" date="1964-08-05">Vom 5. August 1964 </akn:date>
+                     </akn:block>
+                  </akn:preface>
+                 </akn:act>
+              </akn:akomaNtoso>
```
malte-laukoetter, line 83: Could you move this to a fixture file? That makes updating it for changes in future ldml_de versions a lot simpler

VictorDelCampo: @malte-laukoetter could you explain why would it be simpler? It would be anyway a find+replace for both scenarios right? Whether it is a fixture file (XML format) or just simple template string within a test

malte-laukoetter: for the update from 1.6 to 1.7 we used a xslt file to do the transformation of the more complex changes that could not be achieved using a simple search-and-replace. And for doing that we need a xml file as an input.
For the eId changes for the 1.8 update I also imagine that we will need a similar solution.

PR #1119, opened 2025-03-20
Risdev 7204 create concept adr on how test data is organized used

malte-laukoetter:


doc/adr/0017-norm-test-files.md
```diff
@@ -0,0 +1,57 @@
...
+1. Xml-Strings may only be used for short snippets. Whole norms should always be loaded from a xml-file.
+2. We organize the xml files in two categories of test-norms:
+   1. General test norms
+      1. can be used by any test, and should thus be "just" some norms that could happen to be in our system during normal use
+      2. are stored in a folder structure according to their manifestation elis
+      3. are schema valid
+      4. should be documented in a README.md file explaining what norm it is (e.g. "Änderungsgesetz for the Vereinsgesetz")
+   2. Test specific test norms
+      1. may only be used in tests for one class
+      2. stored in the `test-specific` subfolder
```
VictorDelCampo: You mean in the corresponding test package? Like for example adapter/input/restapi/controller package for all unit controller tests?

doc/adr/0017-norm-test-files.md
```diff
@@ -0,0 +1,57 @@
...
+      3. do not need to be schema valid
+      4. are given a name that explains what they are used for
```
VictorDelCampo: Why these get a "custom" name and do not use the eli? I thought the reason for giving them the eli as file name (or structure) was specifically for having the eli in the method for loading the file and see what the eli is so that it can be used in the test more easily?

doc/adr/0017-norm-test-files.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Consequences
+
+- Any of the general test norms can be used when needing "just a norm"
+- If a specific kind of norm (e.g. a Änderungsgesetz) is needed the README.md shows if such a norm exists as a general test norm
+- If none of the general test norms are useful for the specific test a new test specific norm should to be created
+- All the different files related to one norm are in the same folder
+  - we can add a method to the Fixtures class to create a norm based on all files in a folder
```
VictorDelCampo: A "we can add" does not feel as a consequence, rather than just a possibility. But if we wanna load a complete norm and we have the structure with the eli and the folder containing all XMLs, then we MUST add such a method

doc/adr/0017-norm-test-files.md
```diff
@@ -0,0 +1,57 @@
...
+- The eli of a general test norm is always clearly visible as it is used within `Fixtures::loadNormFromDisk`
+- All general test norms have unique elis
+- All general test norms are schema valid
+- For updating the LegalDocMl.de version all norms are available as xml files
+
+## Open Questions
+
+* Should we place the norm fixtures in a more top-level folder within the resource folder?
```
VictorDelCampo: If we wanna have this general norms that could also be used uploading manually or automatically (e2e) tests then i would move this norms to the LegalDocML.de folder

doc/adr/0017-norm-test-files.md
```diff
@@ -0,0 +1,57 @@
...
+
+Additionally, some tests files only include very minimal data that makes them schema-invalid, which creates challenges with using them in certain scenarios.
+
+**We can not use real norms imported from the migration but need to use our own test norms**
+
+## Decision
+
+1. Xml-Strings may only be used for short snippets. Whole norms should always be loaded from a xml-file.
+2. We organize the xml files in two categories of test-norms:
+   1. General test norms
```
andreasphil, line 30: Can we have an agreement that these norms usually shouldn't be touched? I would like to avoid the situation that we keep adding edge cases for specific tests to them. Doesn't mean they're never allowed to change but there should be communication about when and how that happens, just like we do for our samples.

doc/adr/0017-norm-test-files.md
```diff
@@ -0,0 +1,57 @@
...
+  - we can add a method to the Fixtures class to create a norm based on all files in a folder
+- The eli of a general test norm is always clearly visible as it is used within `Fixtures::loadNormFromDisk`
+- All general test norms have unique elis
+- All general test norms are schema valid
+- For updating the LegalDocMl.de version all norms are available as xml files
+
+## Open Questions
+
+* Should we place the norm fixtures in a more top-level folder within the resource folder?
+* Or even keep them in the `/LegalDocML.de` folder?
```
andreasphil: 👍 for keeping it in the LDML.de folder

doc/adr/0017-norm-test-files.md
```diff
@@ -0,0 +1,57 @@
...
+* Better idea on how to organize the test specific norms?
```
andreasphil: I would like to have the test specific norms in folders with the same name as the test class (to keep it simple: just the test class, not nested classes). This makes it easy to see which norms are related to which test class, and obvious when a test class is wrongly using a fixture that belongs to a different test.

doc/adr/0017-norm-test-files.md
```diff
@@ -0,0 +1,57 @@
...
+2. We organize the xml files in two categories of test-norms:
+   1. General test norms
+      1. can be used by any test, and should thus be "just" some norms that could happen to be in our system during normal use
+      2. are stored in a folder structure according to their manifestation elis
+      3. are schema valid
+      4. should be documented in a README.md file explaining what norm it is (e.g. "Änderungsgesetz for the Vereinsgesetz")
+   2. Test specific test norms
+      1. may only be used in tests for one class
+      2. stored in the `test-specific` subfolder
+      3. do not need to be schema valid
```
andreasphil: What do you mean by "do not need to be schema valid"? That there is generally no requirement at all for any of these to be valid? Or that they should generally be valid, but some can intentionally be invalid to test those cases?

Maybe it would be helpful if all our test data was always valid, except the ones that are intentionally invalid. Wouldn't that increase confidence in the correctness of our implementation? But not sure if it is worth the effort ...

doc/adr/0017-norm-test-files.md
malte-laukoetter: for the test specific files i don't think it is helpful to store them based on the eli as that creates very deeply nested folder structures. And I also see usecases where we explicitly need two norms with the same eli for e.g. checking that fixing eids works

doc/adr/0017-norm-test-files.md
malte-laukoetter: i think we should add such a method at some point in time, but i didn't want to force us to do that now. The consequence is just that we are now able to do so, which we where not before.

doc/adr/0017-norm-test-files.md
malte-laukoetter: moved the general once to that folder

doc/adr/0017-norm-test-files.md
malte-laukoetter: i added a sentence about that, specifying that they 
> should not be changed for adding edge-cases. But can be extended if some data is fitting for the norm and "just" missing (e.g. a new metadata field is added).

doc/adr/0017-norm-test-files.md
malte-laukoetter: after trying it out i agree and changed the adr acordingly

doc/adr/0017-norm-test-files.md
malte-laukoetter: i tried to specify it more clearly, can you check it again? :)

doc/adr/0017-norm-test-files.md
VictorDelCampo: You are totally right!

PR #1124, opened 2025-03-21
E-Verkündung API iteration

andreasphil:
RISDEV-7111

### DIFF
--- a/doc/backend/norms-everkuendung-api-v1.yaml
+++ b/doc/backend/norms-everkuendung-api-v1.yaml
@@ -11,7 +11,7 @@ info:
 
     Für die Nutzung aller Endpunkte muss ein gültiger Access Token als `Authorization: Bearer <token>` übermittelt werden. Der Token kann über einen [OAuth2 Client Credentials Grant](https://datatracker.ietf.org/doc/html/rfc6749#section-4.4) abgerufen werden. Die notwendigen Informationen werden vom DigitalService bereitgestellt.
 servers:
-  - url: /api/v1
+  - url: /api/v1/external
     description: Backend
 
 security:
--- a/doc/backend/norms-everkuendung-api-v1.yaml
+++ b/doc/backend/norms-everkuendung-api-v1.yaml
@@ -54,7 +54,7 @@ paths:
                 signature:
                   type: string
                   format: binary
-                  description: "Digitale Signatur für das Archiv. Hinweis: Bis Art und Format der Signatur definiert sind (siehe offene Fragen) ist die Übersendung der Signatur möglich, die Korrektheit wird aber nicht überprüft."
+                  description: "Digitale Signatur für das Archiv. Hinweis: Bis Art und Format der Signatur definiert sind ist die Übersendung der Signatur möglich, die Korrektheit wird aber nicht überprüft."
       responses:
         "202":
           description: Die Verkündung wurde erfolgreich an NeuRIS übermittelt. Sie wird jetzt im Hintergrund verarbeitet. Die Antwort enthält eine Prozess-ID, mit deren Hilfe der Status am Endpunkt `/verkuendungen/status/{processId}` abgefragt werden kann.
@@ -70,12 +70,6 @@ paths:
                 $ref: "#/components/schemas/ProblemDetail"
         "403":
           description: Die Anfrage hat eine gültige Authentifizierung, jedoch fehlt die erforderliche Berechtigung.
-          content:
-            application/problem+json:
-              schema:
-                $ref: "#/components/schemas/ProblemDetail"
-        "409":
-          description: Duplikat erkannt. Eine Verkündung mit dieser ELI wurde bereits übermittelt.
           content:
             application/problem+json:
               schema:
@@ -120,6 +114,7 @@ paths:
 
         - Das ZIP-Archiv ist ungültig
         - Die Signatur des Archivs fehlt oder ist ungültig
+        - Duplikat erkannt: Eine Verkündung mit dieser ELI existiert bereits
         - Die XML-Dokumente weisen entweder XSD- oder Schematron-Fehler auf. Schematron-Warnungen führen nicht zu einem Fehler.
         - Die XML-Dokumente liegen in einer nicht unterstützten Version von LegalDocML.de vor
       parameters:
--- a/doc/backend/norms-everkuendung-api-v1.yaml
+++ b/doc/backend/norms-everkuendung-api-v1.yaml
@@ -33,7 +33,6 @@ paths:
 
         ## Offene Fragen
 
-        - Die erlaubten Dateiformate (MIME-Types) für nicht-strukturierte Anlagen sind noch nicht abschließend geklärt.
         - Das Verfahren zum manuellen Übertragen der Daten, wenn die Verkündung wiederholt nicht angenommen wurde, muss noch definiert werden.
         - Signierung:
           - Art und Format der Signatur müssen noch definiert werden.
@@ -49,7 +48,7 @@ paths:
           | Rechtsetzungsdokument                                   | XML                | `rechtsetzungsdokument.xml`             |                  |
           | Regelungstext(e)                                        | XML                | `regelungstext-{n}.xml`                 |                  |
           | Strukturierte Anlage(n)                                 | XML                | `offenestruktur-{n}.xml`                | soweit vorhanden |
-          | Nicht-strukturierte Anlage(n), z.B. eingebundene Bilder | PDF, JPG, PNG, ... | entsprechend Referenzierung im Dokument | soweit vorhanden |
+          | Nicht-strukturierte Anlage(n), z.B. eingebundene Bilder | PDF, JPG, PNG, GIF | entsprechend Referenzierung im Dokument | soweit vorhanden |
           | konsolidierte Fassung(en) aus der E-Gesetzgebung        | PDF                | `konsolidierte-fassungen/{name}.pdf`    |                  |
         content:
           multipart/form-data:
--- a/doc/backend/norms-everkuendung-api-v1.yaml
+++ b/doc/backend/norms-everkuendung-api-v1.yaml
@@ -276,7 +276,3 @@ components:
         detail:
           type: string
           description: Menschenlesbare Details zur Ursache des Problems. Diese Information ist zum Debugging gedacht und sollte nicht in einer Benutzeroberfläche verwendet werden.
-        extensions:
-          type: object
-          additionalProperties: true
-          description: Weitere Informationen, die die Ursache des Fehlers präzisieren. Ob und in welcher Form dieses Feld vorliegt, unterscheidet sich je nach Art des Fehlers.
--- a/doc/backend/norms-everkuendung-api-v1.yaml
+++ b/doc/backend/norms-everkuendung-api-v1.yaml
@@ -30,13 +30,6 @@ paths:
         Wurde das Normendokumentationspaket vollständig übertragen, wird die Übermittelung mit dem Status-Code `202 ACCEPTED` bestätigt. Die weitere Verarbeitung (z.B. Kontrolle der Signatur, Schema-Validierung) findet in einem Hintergrundprozess statt. Fortschritt und Erfolg bzw. Fehler dieses Prozesses kann über den Endpunkt `/verkuendungen/status/{processId}` abgefragt werden.
 
         Wenn das Normendokumentationspaket nicht übertragen werden konnte, wird die Anfrage mit einem Fehler beendet. Falls der Fehler aufseiten von NeuRIS liegt, sollte die Übertragung nach einem gewissen Zeitraum erneut versucht werden. Dieser Zeitraum kann dem [`Retry-After`-Header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After) der Antwort entnommen werden. Wurde die Verkündung nach 5 Versuchen nicht angenommen, muss ein manueller Austausch mit der Normendokumentation stattfinden.
-
-        ## Offene Fragen
-
-        - Signierung:
-          - Art und Format der Signatur müssen noch definiert werden.
-          - Es muss entschieden werden, ob nur das ZIP-Archiv und/oder die Dateien im Archiv signiert werden.
-          - Das Verfahren zum Austausch des öffentlichen Schlüssels für die Verifizierung der Signatur muss noch definiert werden.
       requestBody:
         required: true
         description: |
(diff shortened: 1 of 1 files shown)


doc/backend/norms-everkuendung-api-v1.yaml
```diff
@@ -27,19 +27,9 @@ paths:
 
         Es kann immer nur eine Anfrage gleichzeitig verarbeitet werden.
 
-        Das Archiv wird während der Anfrage innerhalb von 120 Sekunden validiert. Eine erfolgreiche Übermittlung wird mit dem Status-Code `201 CREATED` bestätigt.
+        Wurde das Normendokumentationspaket vollständig übertragen, wird die Übermittelung mit dem Status-Code `202 ACCEPTED` bestätigt. Die weitere Verarbeitung (z.B. Kontrolle der Signatur, Schema-Validierung) findet in einem Hintergrundprozess statt. Fortschritt und Erfolg bzw. Fehler dieses Prozesses kann über den Endpunkt `/verkuendungen/status/{processId}` abgefragt werden.
```
andreasphil, line 29: Are we going to make any statements or guarantees about how long this information will be available? E.g. would I be able to request the status of a process a week after the upload? Suggestion: obviously the information will be available for a reasonable amount of time (e.g. 1 day), but if and when we clean that up will just be internal to our system and we don't promise that we will keep it indefinitely.

doc/backend/norms-everkuendung-api-v1.yaml
```diff
@@ -27,19 +27,9 @@ paths:
 
         Es kann immer nur eine Anfrage gleichzeitig verarbeitet werden.
```
andreasphil, line 27: Are we going to keep the rate limiting even if we now only upload and save the file and do all the processing async in the background? Is rate limiting vs. not rate limiting making anything about this easier or harder for us?

doc/backend/norms-everkuendung-api-v1.yaml
```diff
@@ -123,6 +105,65 @@ paths:
...
+
+        - `accepted`: Die Verkündung wurde erfolgreich übermittelt, die Verarbeitung hat jedoch noch nicht begonnen.
+        - `processing`: Die Verkündung wird gerade verarbeitet.
+        - `success`: Die Verkündung wurde erfolgreich verarbeitet.
+        - `error`: Beim Verarbeiten der Verkündung ist ein Fehler aufgetreten. In diesem Fall enthält die Antwort weitere Informationen über die Art des Fehlers (siehe `VerkuendungStatusError`).
+
+        Fehler, die während der Verarbeitung auftreten können, sind beispielsweise:
+
+        - Das ZIP-Archiv ist ungültig
+        - Die Signatur des Archivs fehlt oder ist ungültig
```
andreasphil, line 115: Are we going to validate the signature sync during the upload or async? Not sure:

- Sync: because we should not accept potentially dangerous/manipulated data into our system in the first place
- Async: because we might need to still finish the upload and store the file in order to validate the signature? Not sure how that works in Spring

doc/backend/norms-everkuendung-api-v1.yaml
```diff
@@ -176,3 +219,48 @@ components:
...
+        - $ref: "#/components/schemas/VerkuendungStatusError"
+    VerkuendungStatusProcessingOrSuccess:
+      type: object
+      required:
+        - status
+      properties:
+        status:
+          type: string
+          enum: ["accepted", "processing", "success"]
+    VerkuendungStatusError:
```
andreasphil, line 246: This is basically a simplified `ProblemDetail`. Not reusing `ProblemDetail` because it includes properties that are useless or confusing in this context such as the HTTP status code. Ok?

andreasphil:
**Edit:** Solved, see e6765c7428c4aa258aa63c92a3aaf9fb1a936f78

---

I was also wondering about the `409` when posting an announcement:

```yaml
"409":
  description: Duplikat erkannt. Eine Verkündung mit dieser ELI wurde bereits übermittelt.
  content:
    application/problem+json:
      schema:
        $ref: "#/components/schemas/ProblemDetail"
```

How do we detect duplicates? Is it actually necessary to check the ELI? If yes, then this would need to be checked during the async processing and this status could be removed. But wouldn't it also be possible to verify based on the filename alone, because that is based on the BGBL Nr. which should also be unique. Not sure if we store that somewhere though ...

doc/backend/norms-everkuendung-api-v1.yaml
VictorDelCampo: I do think rate limiting is still meaningful, surely not in the way of "X requests per Y time" but rather like amount of not-yet-processed ZIPs?  And since the e-V will be the only external caller of the API we can easily have this "counting" by just checking how many request do we still have in process. But surely it is also depending on the implementation we do, like where to save the files (i read somewhere S3 buckets?) or if we are going to use a queue or something like that. Also it would be important to know how big the load will be coming from the e-V, how many ZIPs a day?

doc/backend/norms-everkuendung-api-v1.yaml
```diff
@@ -123,6 +105,65 @@ paths:
               schema:
                 $ref: "#/components/schemas/ProblemDetail"
 
+  /verkuendungen/status/{processId}:
```
VictorDelCampo, line 101: Do we also need to think about rate-limiting for polling the status?

doc/backend/norms-everkuendung-api-v1.yaml
VictorDelCampo: That is a very valid question.... to which I wouldn't have a good enough answer to pour it into the current state of the API specification.... what comes to my mind is:

- Who needs what? Norms app should keep track (for longer time) or this info in case something like "but we sent it" happens? Maybe also statistics about it like average duration time for the processing?
- What is the legal side?

doc/backend/norms-everkuendung-api-v1.yaml
```diff
@@ -123,6 +105,65 @@ paths:
...
+            application/json:
+              schema:
+                $ref: "#/components/schemas/VerkuendungStatus"
+        "401":
+          description: Fehlende Berechtigung, z.B. kann die Authentifizierung der Anfrage fehlen oder ungültig sein.
+          content:
+            application/problem+json:
+              schema:
+                $ref: "#/components/schemas/ProblemDetail"
+        "403":
```
VictorDelCampo, line 139: I would separate 401 and 403, giving too much information? Also not sure if we are also going to be working with specific roles

doc/backend/norms-everkuendung-api-v1.yaml
```diff
@@ -66,20 +56,12 @@ paths:
                   format: binary
                   description: "Digitale Signatur für das Archiv. Hinweis: Bis Art und Format der Signatur definiert sind (siehe offene Fragen) ist die Übersendung der Signatur möglich, die Korrektheit wird aber nicht überprüft."
       responses:
```
VictorDelCampo, line 57: Missing here the default 500?

doc/backend/norms-everkuendung-api-v1.yaml
```diff
@@ -27,19 +27,9 @@ paths:
...
-        - Das Verfahren zum manuellen Übertragen der Daten, wenn die Verkündung wiederholt nicht angenommen wurde, muss noch definiert werden.
-        - Signierung:
-          - Art und Format der Signatur müssen noch definiert werden.
-          - Es muss entschieden werden, ob nur das ZIP-Archiv und/oder die Dateien im Archiv signiert werden.
-          - Das Verfahren zum Austausch des öffentlichen Schlüssels für die Verifizierung der Signatur muss noch definiert werden.
-        - Eventuell wird es in Zukunft notwendig sein, statt einer synchronen Anfrage eine asynchrone Anfrage zu verwenden, damit das Backend größere Datenmengen besser verarbeiten kann.
+        Wenn das Normendokumentationspaket nicht übertragen werden konnte, wird die Anfrage mit einem Fehler beendet. Falls der Fehler aufseiten von NeuRIS liegt, sollte die Übertragung nach einem gewissen Zeitraum erneut versucht werden. Dieser Zeitraum kann dem [`Retry-After`-Header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After) der Antwort entnommen werden. Wurde die Verkündung nach 5 Versuchen nicht angenommen, muss ein manueller Austausch mit der Normendokumentation stattfinden.
```
VictorDelCampo, line 31: Above it is explicitly said the HTTP 202 but here no HTTP code? or possible codes?

doc/backend/norms-everkuendung-api-v1.yaml
```diff
@@ -27,19 +27,9 @@ paths:
...
 
-        Wenn die Validierung fehlschlägt oder nicht innerhalb dieser Zeit abgeschlossen werden kann, wird die Anfrage mit einem Fehler beendet. Falls der Fehler aufseiten von NeuRIS liegt, sollte die Übertragung nach einem gewissen Zeitraum erneut versucht werden. Dieser Zeitraum kann dem [`Retry-After`-Header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After) der Antwort entnommen werden. Wurde die Verkündung nach 5 Versuchen nicht angenommen, muss ein manueller Austausch mit der Normendokumentation stattfinden.
-
-        ## Offene Fragen
-
-        - Die erlaubten Dateiformate (MIME-Types) für nicht-strukturierte Anlagen sind noch nicht abschließend geklärt.
-        - Das Verfahren zum manuellen Übertragen der Daten, wenn die Verkündung wiederholt nicht angenommen wurde, muss noch definiert werden.
-        - Signierung:
-          - Art und Format der Signatur müssen noch definiert werden.
```
VictorDelCampo, line 39: Are there still some open questions we want to keep as such in the next version of the API specification? like the topic of the signature or if ZIP and files inside should be signed? Though the latter would be a decision to be made by the e-V right?

doc/backend/norms-everkuendung-api-v1.yaml
```diff
@@ -123,6 +105,65 @@ paths:
               schema:
                 $ref: "#/components/schemas/ProblemDetail"
 
+  /verkuendungen/status/{processId}:
+    get:
+      summary: Auskunft über den Fortschritt der Verarbeitung einer übertragenen Verkündung.
```
VictorDelCampo, line 103: nit-picking as non-native German, maybe better "Verarbeitungsstatus" anstatt "Fortschritt der Verarbeitung"?

doc/backend/norms-everkuendung-api-v1.yaml
VictorDelCampo: This is a very good question... actually it would make sense to reject the call already if the signature is not correct? Before we even use up space (S3 bucket or wherever we decide) ? Also depends on how the signature will be sent, detached or embedded? Because if it is embedded we would need to extract the ZIP...

doc/backend/norms-everkuendung-api-v1.yaml
andreasphil: Ok, let's keep it at one for now then. If we do any rate limiting at all they will need to implement a mechanism for that anyway. If we relax the requirements later it's not going to change much.

Regarding the amount of data: there will be one ZIP per announcement. I would guess on most days that's a handful or a small 2-digit number at most.

doc/backend/norms-everkuendung-api-v1.yaml
andreasphil: In my opinion: no, just like we don't currently rate limit any other endpoint. And since this endpoint is specifically for one user (the E-V) and authenticated, I don't see the need.

doc/backend/norms-everkuendung-api-v1.yaml
andreasphil: I separated them based on [your feedback](https://github.com/digitalservicebund/ris-norms/pull/1039#discussion_r1971063321) in the first version 🫢

doc/backend/norms-everkuendung-api-v1.yaml
andreasphil: 500 is [there](https://github.com/digitalservicebund/ris-norms/blob/4a41529651ef14d1ac61368f897c3986a61b8d33/doc/backend/norms-everkuendung-api-v1.yaml#L98) 🤔

doc/backend/norms-everkuendung-api-v1.yaml
andreasphil: Yup, the success code is only mentioned because someone might otherwise expect a 200. But I don't find it necessary to also list all the other codes in the description when they're already in the spec and there's nothing unusual about them

doc/backend/norms-everkuendung-api-v1.yaml
andreasphil: I removed the part about the signature because that's tracked in a separate issue and will be done before the implementation. That should be enough for us not to forget it. The question about signing individual files inside the archive has already been made, only the archive will be signed, not individual files.

doc/backend/norms-everkuendung-api-v1.yaml
andreasphil: Both are fine I guess, I'll stick with the current phrasing because I prefer more smaller words 😄

doc/backend/norms-everkuendung-api-v1.yaml
andreasphil: Ok, let's clarify that some other time. For the purpose of the API specification let's just not mention or commit to anything. It's highly unlikely that the E-V will care at all beyond maybe a few hours until the background job finished. And if they do we can still figure something out then.

doc/backend/norms-everkuendung-api-v1.yaml
andreasphil: After some research I think it should work like this:

The signature will be verified _asynchronously_, i.e. as part of the background job. Why?

The assumption is that it will almost always be correct anyways, since you still need a valid session (and client secret to obtain the session) to upload the file in the first place. Doesn't mean we don't need to be check it to be sure, but I wouldn't optimize for the "invalid signature" case in terms of space used etc. unless we have a specific reason to.

It also keeps the upload endpoint simple because it's only job will be to receive a file, store it somewhere and trigger background processing. Makes it easier for us but also easier for the E-V because it limits error cases on the endpoint.

(The "because we should not accept potentially dangerous/manipulated data into our system in the first place" from above isn't an argument because we need to fully upload the file before verifying the signature, so it will be in the system anyway.)

Regarding your question: the signature will be uploaded as a separate file with the same request as the data, so we can verify it without extracting the files.

PR #1132, opened 2025-03-25
Add database + domain entities for tracking background process status information

andreasphil:


backend/src/main/resources/db/migration/V0.34__create_tables_for_upload_processing_status.sql
```diff
@@ -0,0 +1,17 @@
...
+    id UUID NOT NULL PRIMARY KEY,
+    status verkuendung_import_processing_status NOT NULL DEFAULT 'CREATED',
+    created_at TIMESTAMPTZ NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    started_at TIMESTAMPTZ,
+    finished_at TIMESTAMPTZ
+);
+
+CREATE TABLE IF NOT EXISTS verkuendung_import_process_detail (
+    id UUID NOT NULL PRIMARY KEY,
+    process_id UUID REFERENCES verkuendung_import_processes(id),
```
VictorDelCampo, line 13: Since there is no detail without a process, i would set the `process_id` to NOT NULL

`process_id UUID NOT NULL REFERENCES verkuendung_import_processes(id)`

VictorDelCampo, line 13: We could also think about creating an index on the foreign key because even if we won't be doing JOINS (since with lazy loading hibernate will do a separate SELECT on demand) the filtering using the process_id will benefit from the index.

```
CREATE INDEX idx_verkuendung_process_detail_process_id 
ON verkuendung_import_process_detail(process_id);
```

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/output/database/dto/VerkuendungImportProcessDto.java
```diff
@@ -0,0 +1,73 @@
...
+
+  @OneToMany(fetch = FetchType.LAZY, cascade = CascadeType.ALL)
+  @JoinColumn(name = "process_id")
+  @Default
+  private List<VerkuendungImportProcessDetailDto> detail = new ArrayList<>();
+
+  /**
+   * The different statuses that a background process can have.
+   */
+  public enum Status {
```
VictorDelCampo, line 67: Not happy at all with duplicating an enum class --> error prone. And since we have an hexagonal architecture we can not share an enum class among layers. I would vote for just having the enum in our domain model and saving a simple string into the db. Meaning:
- no need for the enum type on db layer, but rather simple varchar of small size: `status VARCHAR(50) NOT NULL DEFAULT 'CREATED'`
- no need to map between enum in the mapper class, but rather map between domain enum to string using two simple functions:
```
  private static VerkuendungImportProcess.Status mapStatusToDomain(final String status) {
    try {
      return  VerkuendungImportProcess.Status.valueOf(status);
    } catch (IllegalArgumentException e) {
      throw new IllegalArgumentException("Invalid status: " + status, e);
    }
  }
private static String mapStatusToDto(final VerkuendungImportProcess.Status status) {
  return status.name(); 
}
```
Be sure to check if null values, if they are possible

backend/src/main/java/de/bund/digitalservice/ris/norms/application/port/output/LoadVerkuendungImportProcessPort.java
```diff
@@ -0,0 +1,26 @@
...
+   * @return The requested progress information
+   */
+  Optional<VerkuendungImportProcess> loadVerkuendungImportProcess(final Command command);
+
+  /**
+   * Parameters for loading the progress information.
+   *
+   * @param processId ID of the process for which information should be loaded
+   */
+  record Command(UUID processId) {}
```
VictorDelCampo: In the domain entity it is called `id` and not `processId` (this is the naming for the foreign key right?)

backend/src/test/java/de/bund/digitalservice/ris/norms/integration/adapter/output/database/DBServiceIntegrationTest.java
```diff
@@ -60,6 +63,9 @@ class DBServiceIntegrationTest extends BaseIntegrationTest {
   @Autowired
   private MigrationLogRepository migrationLogRepository;
 
+  @Autowired
+  private VerkuendungImportProcessesRepository verkuendungImportProcessesRepository;
```
VictorDelCampo, line 67: Repo autowired to integration test but not to the cleanUp function

VictorDelCampo (changes requested):
Hey @andreasphil good work but I do think there are some changes needed, left as comments. Also not sure why the output port was added and its implementation in the db service including spring repository. I though the ticket was just about extending domain and (db - dto) model. Meaning the only functionality needed would have been the mapper class and the mapper test.  But i guess you wanted to test if it works and for that I would say let's keep the code in this PR

backend/src/main/resources/db/migration/V0.34__create_tables_for_upload_processing_status.sql
VictorDelCampo: Not adding the NOT NULL because the way hibernating is doing the cascade with the OneToMany + JoinColumn. It is persisting first the child, then the parent and then setting the foreign key, leading to a not-null-constraint-violation if we set the process_id to NOT NULL

PR #1143, opened 2025-03-28
Implement endpoint for importing the Normendokumentationspaket

SebastianRossa:


backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/output/s3/NormenDokumentationsPaketBucketService.java
```diff
@@ -0,0 +1,63 @@
...
+  ) {
+    this.s3Client = s3Client;
+    this.bucketName = bucketName;
+  }
+
+  @Override
+  public void saveNormendokumentationspaket(Command command) throws IOException {
+    uploadToBucket(
+      command.processId() + "/" + command.file().getFilename(),
+      RequestBody.fromBytes(command.file().getContentAsByteArray())
```
VictorDelCampo: Do we want to directly stream the file to the S3 bucket?

backend/src/main/resources/application.yaml
```diff
@@ -11,8 +11,8 @@ spring:
     locations: ${flyway.locations}
   servlet:
     multipart:
-      max-file-size: 10MB
-      max-request-size: 10MB
+      max-file-size: 200MB
```
VictorDelCampo, line 14: By doing this how is spring checking the size of the file? Specially if we want to move to streaming directly to S3 bucket

backend/src/main/java/de/bund/digitalservice/ris/norms/config/OtcObsConfig.java
```diff
@@ -66,6 +73,20 @@ public S3Client publicS3Client() throws URISyntaxException {
...
+  /**
+   * Creates a {@link S3Client} for accessing the eVerkündungs S3 bucket.
+   *
+   * <p>This bean is only available in the "staging", "uat", and "production" profiles.</p>
+   *
+   * @return an instance of {@link S3Client} configured for the eVerkündungs bucket.
+   * @throws URISyntaxException if the endpoint URI is malformed.
+   */
+  @Bean(name = "eVerkuendungS3Client")
+  @Profile({ "staging", "uat", "production" })
```
VictorDelCampo, line 85: S3 client should only be configured for uat right?

backend/src/main/resources/application.yaml
SebastianRossa: Spring guarantees to check the file size. https://docs.spring.io/spring-boot/api/java/org/springframework/boot/autoconfigure/web/servlet/MultipartProperties.html We guess Spring checks the content-size header but haven't been going into details.

backend/src/main/java/de/bund/digitalservice/ris/norms/config/OtcObsConfig.java
malte-laukoetter: We don't think so:
- it is a general feature of our application that will (atleast at some point in time) also be needed on production
- and thus we might also have a need to check something with that feature on staging (eg. test a case that isn't working

I can also imagine us changing our current upload to also use this bucket once we changed it to support zip files, as the further processing should be the same.

Also in general I dont think our environments should differ too much - and that the portal-prototype thing is only enabled in one environment is the big exception and only okay as it is really only a temporary thing

PR #1164, opened 2025-04-03
Zielnormen display continued

andreasphil:
Follow up to #1133 

RISDEV-6941


frontend/src/views/amending-law/useGroupedZielnormen.ts
```diff
@@ -25,8 +25,10 @@ export function useGroupedZielnormen(
         return dateA.localeCompare(dateB)
       })
       .reduce((all, current) => {
-        const parsed = DokumentExpressionEli.fromString(current.eli)
-        const eli = `eli/bund/${parsed.agent}/${parsed.year}/${parsed.naturalIdentifier}`
+        const eli = DokumentExpressionEli.fromString(current.eli)
+          .asNormEli()
```
andreasphil: @hamo225 I think the `asNormEli` step is not necessary as `DokumentExpressionEli` can be converted into a norm work ELI directly. I changed it, just FYI

PR #1168, opened 2025-04-04
RISDEV-7114-Validate-incoming-announcement-archive

malte-laukoetter:


malte-laukoetter:
the duplicated code is just necessary boiler plate for the new NormsAppException & similarily the "low" coverage is as we don't test the details of all the new exceptions

malte-laukoetter:
it somehow touched more files than i hoped for :/ But the interesting changes are in
backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/VerkuendungsImportService.java and backend/src/main/java/de/bund/digitalservice/ris/norms/utils/ZipUtils.java

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/output/database/dto/NormManifestationDto.java
```diff
@@ -42,6 +42,10 @@ public class NormManifestationDto {
   @Column(name = "eli_norm_expression")
   private String expressionEli;
 
+  @Immutable
```
VictorDelCampo, line 45: I know it is not related to this PR but I just realized that you introduced the @Immutable annotation to the generated columns and I was curious to know how it behaves and read in the documentation of hibernate:

_@Immutable marks an entity or collection as immutable. An immutable entity may not be updated by the application. This allows Hibernate to make some minor performance optimizations. Updates to an immutable entity will be ignored, but no exception is thrown. @Immutable must be used on root entities only. @Immutable placed on a collection makes the collection immutable, meaning additions and deletions to and from the collection are not allowed. A HibernateException is thrown in this case._

Not sure if you wrote an integration test for this but... doesn't it say to be used on an entity (class) or collection? It doesn't say anything about simple fields like a `String` right?

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/Dokument.java
```diff
@@ -21,11 +21,12 @@
 import org.w3c.dom.NodeList;
 
 /**
- * Represents an abstract "Dokument" in LDML.de (which can be a "Rechtsetzungsdokument", a "Regelungstext" and a "Offene Struktur")
+ * Represents an abstract "Dokument" in LDML.de
  */
 @Getter
 @AllArgsConstructor
-public abstract class Dokument {
+public abstract sealed class Dokument
```
VictorDelCampo, line 25: Love the sealed class :)

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/Dokument.java
```diff
@@ -287,6 +288,24 @@ public void deleteTimeBoundary(TimeBoundaryChangeData timeBoundaryToDelete) {
     temporalDataNode.removeChild(temporalGroupNode);
   }
 
+  /**
+   * Load the filenames of all Dokumente references by this one.
+   * @return a List of filenames
+   */
+  public List<String> getReferencedDokumenteNames() {
```
VictorDelCampo, line 147: This is not something that would only belong to the `Rechtsetzungsdokument`?

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/OffeneStruktur.java
```diff
@@ -7,7 +7,7 @@
  * Represents the "OffeneStruktur" of a norm in LDML.de.
  */
 @Getter
-public class OffeneStruktur extends Dokument {
+public non-sealed class OffeneStruktur extends Dokument {
```
VictorDelCampo, line 10: Do we have any use case for possibly extending this class? And the other two also extending the `sealed` Dokument?

backend/src/main/java/de/bund/digitalservice/ris/norms/utils/ZipUtils.java
```diff
@@ -0,0 +1,88 @@
...
+        );
+      }
+
+      if (entry.getSize() > MAX_SINGLE_FILE_UNCOMPRESSED_SIZE) {
+        throw new IllegalArgumentException(
+          "File %s is too large to be extracted".formatted(entry.getName())
+        );
+      }
+
+      byte[] entryBytes = new byte[(int) entry.getSize()];
```
VictorDelCampo: This class looks already amazing Malte. You implemented already many of the things we would have done for avoiding threats (ZIP bomb). Just some comments on this class (sorry using the first comment line):

Doing `entry.getSize()`would return `-1` if the ZIP entry does not specify its size uncompressed size upfront and would lead to a `NegativeArraySizeException`. Depends how to e-V will be creating the archive, it could be that the size is not specified.

We would need to read the data in chunks and track the size manually, what do you think?

backend/src/main/java/de/bund/digitalservice/ris/norms/utils/ZipUtils.java
```diff
@@ -0,0 +1,88 @@
...
+      int readBytes = zipInputStream.readNBytes(entryBytes, 0, (int) entry.getSize());
+
+      // We expect the stream to have read the whole file and only have one single byte left that is indicating the end of the stream.
+      // Otherwise, the zip file was modified and the entry size is not correct so it could be a zip-bomb.
+      if (zipInputStream.available() != 1) {
```
VictorDelCampo: Not really sure about this.... the documentation says that the `available()` method always returns 1 before EOF has reached for the current entry, meaning the other possible value would only be 0, what would mean the EOF, so why then through an exception?

Meaning it doesn't tell us how many bytes are left or if we have "over-read", right?

backend/src/main/java/de/bund/digitalservice/ris/norms/utils/ZipUtils.java
```diff
@@ -0,0 +1,88 @@
...
+      if (zipArchiveSize > MAX_UNCOMPRESSED_SIZE) {
+        throw new IllegalArgumentException("Archive too large to be extracted");
+      }
+
+      if (zipEntryCount > MAX_FILES_TO_EXTRACT) {
+        throw new IllegalArgumentException("Too many files to be extracted");
+      }
+    }
+
+    zipInputStream.close();
```
VictorDelCampo: This should be done in a `finally` block or try-with-resources, to ensure clean-up on unexpected exceptions.

backend/src/main/java/de/bund/digitalservice/ris/norms/utils/ZipUtils.java
```diff
@@ -0,0 +1,88 @@
...
+          "Expected uncompressed filesize (%s) differs from actual file size for %s".formatted(
+              entry.getSize(),
+              entry.getName()
+            )
+        );
+      }
+
+      zipArchiveSize += readBytes;
+
+      log.debug("Extracted {} (Size: {} bytes)", entry.getName(), readBytes);
```
VictorDelCampo: For now just using the name for logging, but we should also make a check (does not need to be in this ticket) for checking path traversals, meaning entries with things like `../../../../etc/passwd`. But this topic is one of the threats we discussed in the threat modeling last week, so it will be something to tackled as a result of it. So now just FYI

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/VerkuendungsImportService.java
```diff
@@ -47,4 +88,200 @@ public UUID storeNormendokumentationspaket(Query query) throws IOException {
...
+      "Verified new norm from import: {} with {} Dokumenten and {} Binary files",
+      norm.getManifestationEli(),
+      norm.getDokumente().size(),
+      norm.getBinaryFiles().size()
+    );
+
+    return norm;
+  }
+
+  private void validateFileIsZipArchive(Resource file) {
```
VictorDelCampo, line 212: This is only relying on the MIME type detection which is inferring the type from the file extension. Though I think the Spring `MediaTypeFactory` is also sniffing the content based on file byte patterns, it is not 100% foolproof. Should we also add some magic number validation, meaning reading some first bytes?

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/VerkuendungsImportService.java
```diff
@@ -47,4 +88,200 @@ public UUID storeNormendokumentationspaket(Query query) throws IOException {
...
+      .orElseThrow(() -> new NotAZipFileException("Invalid filetype: " + file.getFilename()));
+
+    if (!mediaType.includes(MimeType.valueOf("application/zip"))) {
+      throw new NotAZipFileException("Invalid filetype: " + file.getFilename());
+    }
+  }
+
+  private Set<Dokument> findParseAndValidateDokumente(Map<String, byte[]> files) {
+    if (!files.containsKey(RECHTSETZUNGSDOKUMENT_FILENAME)) {
+      throw new MissingRechtsetzungsdokumentException();
```
VictorDelCampo, line 224: Are you also checking if there is only a single Rechtsetzungsdokument?

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/VerkuendungsImportService.java
```diff
@@ -47,4 +88,200 @@ public UUID storeNormendokumentationspaket(Query query) throws IOException {
...
+    dokumenteToProcess.add(RECHTSETZUNGSDOKUMENT_FILENAME);
+
+    while (!dokumenteToProcess.isEmpty()) {
+      String dokumentName = dokumenteToProcess.poll();
+      log.debug("Processing Dokument {}", dokumentName);
+      var dokument = parseAndValidateDokument(dokumentName, files.get(dokumentName));
+      dokumente.put(dokumentName, dokument);
+
+      if (!dokument.getManifestationEli().getSubtype().equals(dokumentName.split("\\.")[0])) {
+        throw new MismatchBetweenFilenameAndSubtypeException(
```
VictorDelCampo, line 240: I think this was not in the ticket as AC, but is good

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/VerkuendungsImportService.java
```diff
@@ -47,4 +88,200 @@ public UUID storeNormendokumentationspaket(Query query) throws IOException {
...
+          dokument.getManifestationEli()
+        );
+      }
+
+      if (normManifestationEli == null) {
+        normManifestationEli = dokument.getManifestationEli().asNormEli();
+      }
+
+      if (!dokument.getManifestationEli().asNormEli().equals(normManifestationEli)) {
+        throw new InconsistentEliException(
```
VictorDelCampo, line 251: Same here, not in AC but is good!

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/VerkuendungsImportService.java
```diff
@@ -47,4 +88,200 @@ public UUID storeNormendokumentationspaket(Query query) throws IOException {
...
+        "Parsed & validated Dokument {} now looking for referenced dokuments in it",
+        dokument.getManifestationEli()
+      );
+
+      dokumenteToProcess.addAll(
+        dokument
+          .getReferencedDokumenteNames()
+          .stream()
+          // check if we already know about the referenced dokument
+          .filter(Predicate.not(dokumente::containsKey))
```
VictorDelCampo, line 268: Is this something we should raise an exception about? A document that was already processed, meaning it was referenced already and it is referenced again? is it "fachlich" possible?

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/VerkuendungsImportService.java
```diff
@@ -47,4 +88,200 @@ public UUID storeNormendokumentationspaket(Query query) throws IOException {
...
+      case "regelungstext" -> ldmlDeValidator.parseAndValidateRegelungstext(new String(file));
+      case "offenestruktur" -> ldmlDeValidator.parseAndValidateOffeneStruktur(new String(file));
+      case "bekanntmachungstext" -> ldmlDeValidator.parseAndValidateBekanntmachung(
+        new String(file)
+      );
+      default -> throw new InvalidDokumentTypeException(dokumentName);
+    };
+  }
+
+  private boolean isXmlFile(String dokumentName) {
```
VictorDelCampo, line 310: Also checking not 100% foolprof right? See comment about validating if ZIP is a ZIP

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/VerkuendungsImportService.java
```diff
@@ -47,4 +88,200 @@ public UUID storeNormendokumentationspaket(Query query) throws IOException {
...
+          .getReferencedDokumenteNames()
+          .stream()
+          // check if we already know about the referenced dokument
+          .filter(Predicate.not(dokumente::containsKey))
+          .filter(Predicate.not(dokumenteToProcess::contains))
+          .filter(referencedDokumentName -> {
+            if (isXmlFile(referencedDokumentName)) {
+              return true;
+            }
+            log.debug(
```
VictorDelCampo, line 274: Not 100% sure but the referenced documents are only XML? But they can also be PDF or other formats right? Like referencing the binary files with the componentRef in the Rechtsetzungsdokument?

VictorDelCampo (approved):
@malte-laukoetter this is such an amazing work. So much and good thinking into this PR. Great! I do have some questions and some remarks. Still approving this PR because it is good enough to be merged and we can follow up on many of the topics

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/Dokument.java
malte-laukoetter: In a norm with a Eingebundene Stammform the e.g. regelungstext-2 is referenced using a componentRef from the regelungstext-1. So also other Dokumente can reference Dokumente.

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/OffeneStruktur.java
malte-laukoetter: For offene-struktur no. For Regelungstext in theory it could be the case that we need different classes for verkündungsfassung und entwurfsfassung but as hopefully only verkündungsfassungen are ever in our system no. So yes I could also have used final here. Might change that later :)

VictorDelCampo: We actually only verküdungsfassungen will come in, since the entfwurfsfassung are only used on the EGG side right?

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/output/database/dto/NormManifestationDto.java
malte-laukoetter: In the[ javadoc for hibernate 6.2](https://docs.jboss.org/hibernate/orm/6.2/javadocs/org/hibernate/annotations/Immutable.html) (i think you read the one for 5.x) there is this written:

> **Immutable basic-valued attributes**
> A mutable entity may have an immutable field or property.
> An immutable attribute is ignored by the dirty-checking process, and so the persistence context does not need to keep track of its state. This may help reduce memory allocation.

I have not written a test for it but it's also the same we do in other DTOs

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/VerkuendungsImportService.java
malte-laukoetter: ldml_de allows it technically but I don't know if it is something that is also "fachlich" possible. Also as this is "just" for collecting all Dokumente I decided agains raising an exception.

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/VerkuendungsImportService.java
malte-laukoetter: this will need to be reworked after adding the rest of the import anyway as we probably "forget" the filenames before this step anyway, so I would leave it like this for now

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/VerkuendungsImportService.java
malte-laukoetter: they are all xml files, for pdf attachments they are (as far as i know) a offene-struktur-xml that then contains a reference to the pdf file

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/OffeneStruktur.java
malte-laukoetter: yes

backend/src/main/java/de/bund/digitalservice/ris/norms/utils/ZipUtils.java
malte-laukoetter: the problem with tracking the size manually is that it is harder to protect against zipbombs. If it turns out that EV is creating the zip files in a very obscure way that does not contain the sizes we can still change this implementation. For now i added a check that the size is positive.

backend/src/main/java/de/bund/digitalservice/ris/norms/utils/ZipUtils.java
malte-laukoetter: As the entryBytes thing has the exact size that the unpacked entry should have we will neither be able to "over-read" or have any bytes left unless the size data in the zip file are wrong. In that case we want to throw an exception to prevent e.g a zip bomb as in that case the entry size check against `MAX_SINGLE_FILE_UNCOMPRESSED_SIZE`  is not guaranteed to have checked against the unpacked size.

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/VerkuendungsImportService.java
malte-laukoetter: the `MediaTypeFactory` only looks at the filename. I'm not sure that we need any better validation as parsing the zip file would not work if it isn't a zip file. For reading the first few bytes I would otherwise rather integrate a library that does that like e.g. https://tika.apache.org/3.1.0/detection.html

backend/src/main/java/de/bund/digitalservice/ris/norms/utils/ZipUtils.java
malte-laukoetter: i adde a check that only flat filenames (without any path traversal) are used

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/VerkuendungsImportService.java
malte-laukoetter: A rechtsetzungsdokument is always called `rechtsetzungsdokument-1.xml` so there only ever can be one. If there would ever be a file called e.g. `rechtsetzungsdokument-2.xml` I would just say that we treat it like any other additional file

backend/src/main/java/de/bund/digitalservice/ris/norms/utils/ZipUtils.java
VictorDelCampo: Yeah but that is what I meant, it if returns 1 then we should throw the error and not in the other way around right? Throwing an error if the `available` method returns 0 (the other possible value), because that would mean that there are no bytes left

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/VerkuendungsImportService.java
VictorDelCampo: I know that parsing the ZIP would also tell us if it is truly a ZIP or not, but you are throwing here the `NotAZipFileException` and this one should also be thrown if when trying to parse the ZIP it comes out that it is not a ZIP right? Because right now if something goes wrong when trying to parse the ZIP we would get a `InvalidStructureInZipFileException`
And adding a new dependency, if needed and/or we have a big benefit, why not :)

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/VerkuendungsImportService.java
VictorDelCampo: Yeah but one AC of the JIRA ticket is:
`MUST: Validate that the archive contains a single rechtssetzungsdokument-1.xml`
Meaning we agreed with PM that we would validate this meaning if there is more than 1 Rechsetzungsdokument we should throw an error right?

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/VerkuendungsImportService.java
VictorDelCampo: But page 121 of the specification (1.7.2) shows an example like:
```
<akn:component>
<akn:documentRef href="eli/bund/bgbl-1/2002/s100/deu/externes-dokument-1.pdf" showAs="externes-dokument"/>
</akn:component>
```
With the explanation:
_Ein rechtsetzungsdokument kann beliebig viele Teildokumente umfassen. Dies können auch Doku-
mente sein, die nicht im LegalDocML.de-Format vorliegen (etwa eine als PDF-Dokument vorliegende
Stellungnahme des Normenkontrollrates). Auch Nicht-LegalDocML.de-Dokumente werden anhand der
Elemente teildokument und teildokumentVerweis eingebunden_

Only _rechtsverbindlicher Anlagen_ will be referenced within the regelungstext

malte-laukoetter: Oh good catch, but it luckily doesn't change things about the code, i'll add a test with such a documentRef

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/VerkuendungsImportService.java
malte-laukoetter: i interpreted that AC as we should validate that the archive contains a rechtssetzungsdokument (and not 0). It is technically not possible to have to files with that name so there is nothing to validate for the more than 1 case

PR #1192, opened 2025-04-09
Implement service to track import process

SebastianRossa:


backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/output/database/service/VerkuedungImportProcessDBService.java
```diff
@@ -34,4 +42,70 @@ public Optional<VerkuendungImportProcess> loadVerkuendungImportProcess(
...
+              .builder()
+              .type(incomingDetail.getType())
+              .title(incomingDetail.getTitle())
+              .detail(incomingDetail.getDetail())
+              .build();
+            verkuendungImportProcessDto.getDetail().add(safeDetail);
+          }
+        }
+      }
+      default -> log.error("Unhandled status {}", status);
```
malte-laukoetter, line 108: i dont think we need a default case here and with it missing the compiler catches it if we add a new status

backend/src/main/java/de/bund/digitalservice/ris/norms/application/exception/StatusNotFoundException.java
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  @Override
+  public URI getType() {
+    return URI.create("/errors/process-id-not-found");
+  }
+
+  @Override
+  public String getTitle() {
+    return "Could not find process with id " + processId;
```
malte-laukoetter: the title should be a static string and not depend on the paramters, the processId should instead be returned by getProperties. The message you have written here is more like the detail message and should therefore be passed to the super constructor.

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/output/database/service/VerkuedungImportProcessDBService.java
```diff
@@ -34,4 +42,70 @@ public Optional<VerkuendungImportProcess> loadVerkuendungImportProcess(
...
+    SaveVerkuendungImportProcessPort.Command command,
+    VerkuendungImportProcessDto verkuendungImportProcessDto
+  ) {
+    final VerkuendungImportProcessDto.Status status = VerkuendungImportProcessMapper.mapStatusToDto(
+      command.status()
+    );
+    verkuendungImportProcessDto.setStatus(status);
+
+    switch (status) {
+      case CREATED -> verkuendungImportProcessDto.setCreatedAt(Instant.now());
```
malte-laukoetter, line 78: its really nice to have this automated <3

backend/src/main/java/de/bund/digitalservice/ris/norms/application/port/output/SaveVerkuendungImportProcessPort.java
```diff
@@ -0,0 +1,59 @@
...
+  ) {
+    /**
+     * Parameter for saving/updating a {@link VerkuendungImportProcess} object
+     *
+     * @param id of the job
+     * @param status of the file import
+     * @param details of what went wrong
+     */
+    public Command {
+      Objects.requireNonNull(id, "id cannot be null");
```
malte-laukoetter: i'm really unsure if the message here is really helpful or just one more thing we need to update if we ever want to change this again

backend/src/main/java/de/bund/digitalservice/ris/norms/config/JobrunrExceptionFilter.java
```diff
@@ -0,0 +1,42 @@
...
+  public void onFailedAfterRetries(Job job) {
+    saveVerkuendungImportProcessPort.saveOrUpdateVerkuendungImportProcess(
+      new SaveVerkuendungImportProcessPort.Command(
+        job.getId(),
+        VerkuendungImportProcess.Status.ERROR,
+        List.of(
+          new VerkuendungImportProcessDetail(
+            "/errors/jobRunFailed",
+            "Tried to import a Normendokumentationspacket the max amount of times but failed",
+            "none"
```
malte-laukoetter, line 34: this details part is probably going to be a json string for the other cases, so just keeping it empty would probably lead to fewer errors

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/VerkuendungImportProcess.java
```diff
@@ -29,7 +26,7 @@ public class VerkuendungImportProcess {
   private Instant finishedAt;
 
   @Default
-  private List<VerkuendungImportProcessDetail> detail = new ArrayList<>();
+  private List<VerkuendungImportProcessDetail> detail = List.of();
```
malte-laukoetter: just to be sure: you want this list to be immutable?

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/output/database/service/VerkuedungImportProcessDBService.java
```diff
@@ -34,4 +42,70 @@ public Optional<VerkuendungImportProcess> loadVerkuendungImportProcess(
...
+      case PROCESSING -> verkuendungImportProcessDto.setStartedAt(Instant.now());
+      case SUCCESS -> verkuendungImportProcessDto.setFinishedAt(Instant.now());
```
VictorDelCampo, line 80: What about concatenating both cases?

` case SUCCESS, ERROR -> verkuendungImportProcessDto.setFinishedAt(Instant.now());`

backend/src/main/java/de/bund/digitalservice/ris/norms/application/port/output/SaveVerkuendungImportProcessPort.java
VictorDelCampo: I agree with malte and we don't have something like this in any other command. Also this is just a fail-fast when calling the port in case of a null value was passed. But shouldn't we "catch" such case in the implementation of the output port?

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/output/database/service/VerkuedungImportProcessDBService.java
```diff
@@ -34,4 +42,70 @@ public Optional<VerkuendungImportProcess> loadVerkuendungImportProcess(
...
+  public VerkuendungImportProcess saveOrUpdateVerkuendungImportProcess(
+    SaveVerkuendungImportProcessPort.Command command
+  ) {
+    final Optional<VerkuendungImportProcessDto> optionalExistingVerkuendungImportProcessDto =
+      verkuendungImportProcessesRepository.findById(command.id());
+
+    if (optionalExistingVerkuendungImportProcessDto.isPresent()) {
+      var existingVerkuendungImportProcessDto = optionalExistingVerkuendungImportProcessDto.get();
+      updateProcessDto(command, existingVerkuendungImportProcessDto);
+      return VerkuendungImportProcessMapper.mapToDomain(existingVerkuendungImportProcessDto);
```
VictorDelCampo, line 56: The from the DB loaded entry is not being saved again after updating the DTO right?

backend/src/main/java/de/bund/digitalservice/ris/norms/application/exception/StatusNotFoundException.java
```diff
@@ -0,0 +1,33 @@
...
+
+import de.bund.digitalservice.ris.norms.utils.exceptions.NormsAppException;
+import java.net.URI;
+import java.util.Map;
+import java.util.UUID;
+import lombok.Getter;
+
+/** Indicates that the requested status does not exist. */
+@Getter
+public class StatusNotFoundException extends RuntimeException implements NormsAppException {
```
VictorDelCampo: The naming of the exception is not that matching right? Shouldn't it be something like `ImportProcessNotFoundException`?

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/NormendokumentationspaketService.java
```diff
@@ -0,0 +1,40 @@
...
+    final Resource zipFile,
+    final Resource signature
+  ) {
+    saveVerkuendungImportProcessPort.saveOrUpdateVerkuendungImportProcess(
+      new SaveVerkuendungImportProcessPort.Command(
+        processId,
+        VerkuendungImportProcess.Status.CREATED
+      )
+    );
+
```
VictorDelCampo, line 33: As discussed we could already put a general try/catch here for then creating the details within the catch block

backend/src/main/java/de/bund/digitalservice/ris/norms/config/JobrunrExceptionFilter.java
VictorDelCampo: But shouldn't we add the actual error that was encountered? Not sure if through the Job object we can get it, because otherwise we wouldn't know what the real reason was right? And when pulling the process status the info would not be helpful for e-V

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/VerkuendungImportProcess.java
VictorDelCampo: I don't think it should be immutable..

backend/src/test/java/de/bund/digitalservice/ris/norms/application/service/VerkuendungsImportServiceTest.java
```diff
@@ -0,0 +1,80 @@
...
+import de.bund.digitalservice.ris.norms.application.port.output.SaveNormendokumentationspaketPort;
+import de.bund.digitalservice.ris.norms.application.port.output.SaveVerkuendungImportProcessPort;
+import de.bund.digitalservice.ris.norms.domain.entity.VerkuendungImportProcess;
+import java.util.List;
+import java.util.Optional;
+import java.util.UUID;
+import org.jobrunr.scheduling.JobScheduler;
+import org.junit.jupiter.api.Test;
+
+class VerkuendungsImportServiceTest {
```
VictorDelCampo, line 20: No unit test for `storeNormendokumentationspaket` ?

VictorDelCampo (changes requested):
Some comments

backend/src/main/java/de/bund/digitalservice/ris/norms/config/JobrunrExceptionFilter.java
malte-laukoetter: This case will only happen when the job is running into an unexpected error (and not if something is invalid). In that case we do not want to expose the actual error as we do not know what data it might contain.
The jobrunr job needs to finish successfully even when the uploaded file is invalid to not rerun it multiple times. So an invalid upload is still a successful jobrunr job that updates the VerkuendungsImportProcess status to error (and therefore also handles the writing of the details itself).

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/output/database/service/VerkuedungImportProcessDBService.java
SebastianRossa: It does. Added a test

PR #1198, opened 2025-04-10
RISDEV-7115-Import-incoming-announcement-archive

malte-laukoetter:


malte-laukoetter:
~This currently also includes the changes done in https://github.com/digitalservicebund/ris-norms/pull/1168 so i will rebase it once that one is merged~

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/VerkuendungsImportService.java
```diff
@@ -119,17 +144,74 @@ public VerkuendungImportProcess getStatus(
...
+      );
 
-    throw new RuntimeException("Not implemented yet");
+      saveVerkuendungImportProcessPort.saveOrUpdateVerkuendungImportProcess(
+        new SaveVerkuendungImportProcessPort.Command(
+          process.getId(),
+          VerkuendungImportProcess.Status.SUCCESS
+        )
+      );
+    } catch (Exception e) {
```
VictorDelCampo, line 183: We catch all exceptions but only create details for "our" exceptions (NormsAppException)? Meaning any other unexpected exception would only be logged but no error detail would be created. Wouldn't this be confusing for the e-V?

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/VerkuendungsImportService.java
```diff
@@ -209,107 +252,159 @@ public Norm parseAndValidate(Resource zipFile)
...
-    };
+    }
+
+    return dokument;
   }
 
   private boolean isXmlFile(String dokumentName) {
-    return MediaTypeFactory
-      .getMediaType(dokumentName)
+    return mediaTypeService
```
VictorDelCampo: So here we still rely on the file name?

VictorDelCampo (approved):
Looks good Malte. Only two minor comments

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/VerkuendungsImportService.java
malte-laukoetter: If there is an error we do not have listed here we do not want to expose the details of it as that could reveal information about our system that could help attackers (see: https://cheatsheetseries.owasp.org/cheatsheets/Error_Handling_Cheat_Sheet.html). And as it is an unexpected exception we should always investigate that anyway and it is probably nothing that the e-V could fix anyway

VictorDelCampo: Jep you are completely right. But we should probably put some general info like "internal error"? because otherwise they will get the ERROR state but no info at all

PR #1214, opened 2025-04-17
refactor: rename "announcement" to "verkuendung" throughout codebase

hamo225:
+ flyway script to rename database table from announcements to verkuendungen

RISDEV-7408


malte-laukoetter (approved):
looks good :)

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/controller/external/EverkuendungenApiController.java
```diff
@@ -17,12 +17,12 @@
 /** Controller for verkuendungen-related actions. */
 @RestController
 @RequestMapping("/api/v1/external/verkuendungen")
-public class VerkuendungController {
+public class EverkuendungenApiController {
```
malte-laukoetter: i think EverkuendungApiController would be better but thats just a feeling

hamo225: I know what you mean. I was trying to keep consistent with singular and plural. Maybe better to keep it as its spoken.

PR #1216, opened 2025-04-17
Change api spec to return specific xsd and schematron errors

SebastianRossa:
RISDEV-7562


doc/backend/norms-everkuendung-api-v1.yaml
```diff
@@ -278,3 +278,70 @@ components:
...
+      required:
+        - type
+        - file
+        - message
+        - xpath
+        - eId
+      properties:
+        type:
+          type: string
+          enum: [schematron]
```
malte-laukoetter: the value for this currently looks something like this `/errors/ldml-de-not-schematron-valid/failed-assert/SCH-00050-005`

doc/backend/norms-everkuendung-api-v1.yaml
```diff
@@ -278,3 +278,70 @@ components:
...
+      required:
+        - type
+        - file
+        - message
+        - line
+        - column
+      properties:
+        type:
+          type: string
+          enum: [xsd]
```
malte-laukoetter: the value for this currently looks somethings like this `/errors/ldml-de-not-valid/cvc-pattern-valid`

doc/backend/norms-everkuendung-api-v1.yaml
```diff
@@ -278,3 +278,70 @@ components:
...
+          example: xsd
+        file:
+          type: string
+          description: Name der Datei, in der der Fehler gefunden wurde.
+          example: regelungstext-1.xml
+        message:
+          type: string
+          description: Die Fehlermeldung des XSD-Parsers.
+          example: "cvc-complex-type.2.4.a: Invalid content was found starting with element 'author'."
+        line:
```
malte-laukoetter: lineNumber

doc/backend/norms-everkuendung-api-v1.yaml
```diff
@@ -278,3 +278,70 @@ components:
...
+          type: integer
+          description: Zeilennummer des Fehlers im XML.
+          example: 23
+        column:
```
malte-laukoetter: columnNumber

doc/backend/norms-everkuendung-api-v1.yaml
```diff
@@ -278,3 +278,70 @@ components:
...
+      properties:
+        type:
+          type: string
+          enum: [schematron]
+          example: schematron
+        file:
+          type: string
+          description: Name der Datei, in der der Fehler gefunden wurde.
+          example: regelungstext-1.xml
+        message:
```
malte-laukoetter: details

doc/backend/norms-everkuendung-api-v1.yaml
```diff
@@ -278,3 +278,70 @@ components:
...
+      properties:
+        type:
+          type: string
+          enum: [xsd]
+          example: xsd
+        file:
+          type: string
+          description: Name der Datei, in der der Fehler gefunden wurde.
+          example: regelungstext-1.xml
+        message:
```
malte-laukoetter: details

doc/backend/norms-everkuendung-api-v1.yaml
```diff
@@ -278,3 +278,70 @@ components:
...
+        - file
+        - message
+        - xpath
+        - eId
+      properties:
+        type:
+          type: string
+          enum: [schematron]
+          example: schematron
+        file:
```
malte-laukoetter: this is something we would still need to add

doc/backend/norms-everkuendung-api-v1.yaml
```diff
@@ -278,3 +278,70 @@ components:
...
+        - file
+        - message
+        - line
+        - column
+      properties:
+        type:
+          type: string
+          enum: [xsd]
+          example: xsd
+        file:
```
malte-laukoetter, line 308: this is something we would still need to add

malte-laukoetter (commented):
I added some comments where this differs from the current implementation

doc/backend/norms-everkuendung-api-v1.yaml
```diff
@@ -270,11 +273,80 @@ components:
...
           type: string
           description: Menschenlesbare Details zur Ursache des Problems. Diese Information ist zum Debugging gedacht und sollte nicht in einer Benutzeroberfläche verwendet werden.
+        errors:
+          type: array
+          description: Eine Liste strukturierter Validierungsfehler aus XSD oder Schematron.
+          items:
+            oneOf:
+              - $ref: "#/components/schemas/XsdValidationError"
+              - $ref: "#/components/schemas/SchematronValidationError"
+        extensions:
```
VictorDelCampo, line 291: Do we need these extensions? I thought we wanted to have a "fix" structure of what can be returned by us and we just wanted to have the type, title and details right? And we can just add the extra information we would put into these extensions into the detail error

VictorDelCampo (dismissed):
The mapper `VerkuendungStatusMapper` hasn't changed and I would have expected that this PR would do so. It still has the old implementation from Sebastian where both title and details are being concatenated into single strings.
The schema also doesn't have yet the possible "errors"

backend/src/main/java/de/bund/digitalservice/ris/norms/application/exception/NotLdmlDeXmlFileException.java
```diff
@@ -12,7 +12,7 @@ public class NotLdmlDeXmlFileException extends RuntimeException implements Norms
   private final String fileName;
 
   public NotLdmlDeXmlFileException(String fileName) {
```
VictorDelCampo, line 14: Shouldn't this exception be part of the `CreateVerkuendungUseCase`?

VictorDelCampo: Shouldn't this exception be part of the CreateVerkuendungUseCase?

VictorDelCampo (commented):
Looks amazing Malte, specially the serializers (both for response schema as well for persisting to db column) and also the command that gets a NormsAppException. Crazy good!

A minor comment about 1 exception and one thought, i was a bit bothered about having to extend from RuntimeException and implement the NormsAppException. Couldn't we just transform the latter from interface to abstract and extend there the RunTimeException so that our custom exceptions just beed to extend NormsAppException?

backend/src/main/java/de/bund/digitalservice/ris/norms/application/exception/NotLdmlDeXmlFileException.java
malte-laukoetter: moved it :)

malte-laukoetter:
> A minor comment about 1 exception and one thought, i was a bit bothered about having to extend from RuntimeException and implement the NormsAppException. Couldn't we just transform the latter from interface to abstract and extend there the RunTimeException so that our custom exceptions just beed to extend NormsAppException?

I was first doing that but we have some exceptions that actually shouldn't be runtime exceptions (most of the once in `ProcessNormendokumentationspaketUseCase`). So I decided against that and to instead both implement the interface and extend the RuntimeException in the other cases.

PR #1224, opened 2025-04-22
feat: improve validation error messages for time boundaries

hamo225:
- adds German translations for known backend validation messages (missing date or type, duplicate entries, maximum number of time boundaries supported)
- Previously, these were displayed in raw English from the backend

RISDEV-0000

### DIFF
--- a/frontend/src/lib/errorMessages.ts
+++ b/frontend/src/lib/errorMessages.ts
@@ -63,10 +63,24 @@ export const errorMessages = {
       "Wenn das Problem bestehen bleibt, wenden Sie sich an den Betreiber",
   }),
 
-  "/errors/input-validation-error": (e: ErrorResponse) => ({
-    title: "Eingabefehler",
-    message: `Fehler bei den eingehenden Daten: ${e.detail}`,
-  }),
+  "/errors/input-validation-error": (e: ErrorResponse) => {
+    const validationDetailTranslations: Record<string, string> = {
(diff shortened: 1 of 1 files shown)


frontend/src/lib/errorMessages.ts
```diff
@@ -63,10 +63,24 @@ export const errorMessages = {
...
+      "Not all combinations of date + art are unique.":
+        "Das Datum darf für dieselbe Art nur einmal vorkommen.",
+    }
+
+    const detail = e.detail ?? "Unbekannter Validierungsfehler"
+    const germanDetail = validationDetailTranslations[detail] ?? detail
+
+    return {
+      title: "Eingabefehler",
+      message: `Fehler bei den eingehenden Daten: ${germanDetail}`,
```
andreasphil: question: do we need the message here to have additional text ("Fehler bei den eingehenden Daten") or could it just be `germanDetail`? Since we don't have a lot of space where we show the error messages, keeping it short might make it more readable.

andreasphil (commented):
One question, otherwise LGTM. Please consider and feel free to merge afterwards

frontend/src/lib/errorMessages.ts
hamo225: Good point — the title already says “Eingabefehler”, so repeating “Fehler bei den eingehenden Daten” in the message is a bit redundant. Happy to update that!

PR #1253, opened 2025-04-29
Risdev 7557 disable updating deleting of used temporal intervals

malte-laukoetter:


malte-laukoetter:
this does not include checks in the backend, this will be done in RISDEV-7669

PR #1288, opened 2025-05-12
fix(norm): Update date logic to include inkraft and ausserkraft dates RISDEV-7673

hamo225:


backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/Norm.java
```diff
@@ -239,8 +239,10 @@ public boolean isInkraftAt(LocalDate date) {
 
     return (
       inkrafttreteDatum.isPresent() &&
-      inkrafttreteDatum.get().isBefore(date) &&
-      (ausserkrafttreteDatum.isEmpty() || ausserkrafttreteDatum.get().isAfter(date))
+      (inkrafttreteDatum.get().isBefore(date) || inkrafttreteDatum.get().isEqual(date)) &&
```
VictorDelCampo: What about:

```
    return (
      inkrafttreteDatum.isPresent() &&
      !inkrafttreteDatum.get().isAfter(date) &&
      (ausserkrafttreteDatum.isEmpty() || !ausserkrafttreteDatum.get().isBefore(date))
    );
```

Or even something like:
```
    return (
      inkrafttreteDatum.map(d -> !d.isAfter(date)).orElse(false) &&
      ausserkrafttreteDatum.map(d -> !d.isBefore(date)).orElse(true)
    );
```

I personally find "big" condition groups hard to read and therefore prefer always a more concise version of it. If you agree with me choose the one you can read better :)

backend/src/test/java/de/bund/digitalservice/ris/norms/domain/entity/NormTest.java
```diff
@@ -281,37 +283,22 @@ void itReturnsEmptyForUnknownEli() {
   @Nested
   class isInkraftAt {
 
-    @Test
-    void itReturnsTrueIfTheDateIsAfterInkrafttreteDatum() {
-      // Given
+    @ParameterizedTest
+    @CsvSource(
```
VictorDelCampo, line 287: Very nice! :)

VictorDelCampo (commented):
Looks good :) Just a possible optimization if you want

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/Norm.java
hamo225: While i do read the first option better on first glance. The second one is more concise. Thanks for the optimisation.

PR #1292, opened 2025-05-13
API integration Zielnormen preview

andreasphil:
RISDEV-7770


frontend/src/services/zielnormPreviewService.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  const mappedData = ref<ZielnormPreview[] | null>(null)
+  watch(
+    data,
+    (newData) => {
+      if (newData === null) mappedData.value = null
+      else mappedData.value = newData.map(mapResponseToDomain)
+    },
+    { immediate: true },
+  )
```
malte-laukoetter, line 76:
```ts
const mappedData = computed(() => data.value?.map(mapResponseToDomain))
```

PR #1303, opened 2025-05-16
Create affected expressions

SebastianRossa:


backend/src/main/java/de/bund/digitalservice/ris/norms/application/port/input/LoadZielnormenUseCase.java
```diff
@@ -7,23 +7,35 @@
 import javax.annotation.Nullable;
 
 /**
- *  Use case for getting information about the Zielnormen that will be created or set to gegenstandslos when applying the {@link ZielnormReference}s of the given Verkündung.
+ *  Use case for getting information about the expressions of Zielnormen that will be created
+ *  or set to gegenstandslos when applying the {@link ZielnormReference}s (containing the geltungszeit)
+ *  to the given Verkündung.
  */
-public interface LoadZielnormenPreviewUseCase {
+public interface LoadZielnormenUseCase {
```
malte-laukoetter, line 14: I find the word load here a bit confusing now that we plan to use the usecase for both loading the preview and actually **creating** the zielnormen. I think a name like `CreateZielnormenUseCase` would be more fitting. Also the doc comment could use an update :)

VictorDelCampo: I agree with the need for renaming, but not 100% sure if a "Create" word is fitting, since it may sound like they are persisted. What about `GenerateZielNormenUseCase`?

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/NormService.java
```diff
@@ -398,4 +400,33 @@ private List<NormExpressionEli> collectRelevantExistingExpressions(
       .map(Norm::getExpressionEli)
       .toList();
   }
+
+  /** check whether the expression is already stored in the database and set the isCreated flag */
+  private List<ZielnormPreview.Expression> calculateIsCreated(
```
VictorDelCampo, line 405: Not sure about this method, since we would be doing some unnecessary DB calls, as some of them would have already the `isCreated` set to `true` (we have retrieved them already from the DB when calling `collectRelevantExistingExpressions`). And if the new expressions have been already created and we cross-check with `<norms:amended-norm-expressions>` they would also have already the isCreated flag.
Meaning can we rely in the isCreated flag already from before?

backend/src/main/java/de/bund/digitalservice/ris/norms/application/port/input/LoadZielnormenUseCase.java
malte-laukoetter: >  since it may sound like they are persisted

But we do persist them, don't we? Just not in the dry-run case but that should be clear regardless as it's a dry-run

VictorDelCampo: Don't really care about the naming, just not load ;)

malte-laukoetter: same for me, just wanted to make sure we don't have a misunderstanding about the persisting ^^

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/NormService.java
SebastianRossa: @malte-laukoetter you mentioned that you are not sure that the status isCreated is currently set right. Is that obsolete with Victors comment? Or do you see a reason why we might still need a guardian-like method like calculateIsCreated.

malte-laukoetter: That should be obsolete once we are doing the matching with `<norms:amended-norm-expressions>` and also use that information for setting isCreated. So i don't think we will need that guardian :)

SebastianRossa:
Abandoned. Code is now continued under https://github.com/digitalservicebund/ris-norms/pull/1315

PR #1312, opened 2025-05-19
Edit expression view RISDEV-7861 

hamo225:


frontend/src/views/verkuendungen/verkuendungDetail/textkonsolidierung/ExpressionTextkonsolidierungEditor.view.vue
```diff
@@ -0,0 +1,270 @@
...
+          class="m-16 mt-8"
+        />
+
+        <RisEmptyState
+          v-else-if="!toc || !toc.length"
+          text-content="Keine Artikel gefunden."
+          class="m-16 mt-8"
+        />
+        <div v-else class="flex-1 overflow-auto p-10">
+          <h2
```
andreasphil: todo: can you clarify with product what "Inhaltsübersicht" means in this context? My hunch is that it's not a heading but an element inside the law, in which case we wouldn't need this. If it is supposed to be a heading, it should not be blue, because that suggests it's clickable like the other TOC items.

frontend/src/views/verkuendungen/verkuendungDetail/zielnormen/VerkuendungZielnormen.view.vue
```diff
@@ -29,7 +29,7 @@ import {
   useToast,
 } from "primevue"
 import { computed, ref, watch } from "vue"
-import RisDokumentExplorer from "./RisDokumentExplorer.vue"
+import RisDokumentExplorer from "../../../../components/RisDokumentExplorer.vue"
```
andreasphil: nitpick: use `@/...` instead of `../../...`

frontend/src/views/verkuendungen/verkuendungDetail/textkonsolidierung/ExpressionTextkonsolidierungEditor.view.vue
```diff
@@ -0,0 +1,270 @@
...
+  if (!editorView.value) return
+
+  const cursor = new SearchCursor(editorView.value.state.doc, `eId="${eid}"`)
+  if (!cursor.next()) return
+
+  const { from, to } = cursor.value
+  editorView.value.dispatch({
+    selection: { anchor: from, head: to },
+  })
+  requestAnimationFrame(() => {
```
andreasphil: suggestion: `requestAnimationFrame` is not wrong here, but in Vue the more idiomatic way would be to use [nextTick](https://vuejs.org/api/general.html#nexttick). Unless you already tried that and it didn't work for some reason :)

frontend/src/views/verkuendungen/verkuendungDetail/RisZielnormenList.vue
```diff
@@ -35,7 +35,10 @@ type MappedRisZielnormenListItem = Omit<
   }>
 }
 
-const { items } = defineProps<{ items: RisZielnormenListItem[] }>()
+const { items, verkuendungeli } = defineProps<{
+  items: RisZielnormenListItem[]
+  verkuendungeli: string
```
andreasphil: nitpick: `verkuendungEli` → camel case

frontend/src/router.ts
```diff
@@ -133,6 +141,14 @@ const routes: readonly RouteRecordRaw[] = [
       },
     ],
   },
+  // {
```
andreasphil: todo: remove

frontend/src/components/editor/RisCodeEditor.vue
```diff
@@ -93,6 +95,7 @@ watch(
       ],
       parent: editorElement.value,
     })
+    emit("ready", editorView.value)
```
andreasphil: todo: I think the logic for who manages the scrolling is backwards. Basically: instead of passing a reference to the editor outside the RisCodeEditor component and have the parent component interact with that directly, the RisCodeEditor should implement and expose a method scrollToText(text: string) (or similar) and handle the scrolling internally. The parent can then get a ref to the RisCodeEditor and call editor.scrollToText(`eli=...`)

why: because the editor is an internal implementation detail of the RisCodeEditor component that should be hidden to the outside world

andreasphil (changes requested):
Good work, just some minor things and I think the location of the logic for the scrolling should be refactored (see comment). Feel free to merge and create another PR for the changes if you want.

frontend/src/views/verkuendungen/verkuendungDetail/textkonsolidierung/ExpressionTextkonsolidierungEditor.view.vue
hamo225: Will check with product.

PR #1319, opened 2025-05-19
Add script to convert test files to ldml_de 1.8

malte-laukoetter:
- [x] Schema location
- [x] run schematron validation
- [x] remove bundesregierungs metadata from non rechtsetzungsdokumente
- [x] rename regelungstext-n.xml to regelungstext-verkuendungsfassung-n.xml


VictorDelCampo (approved):
It simply looks amazing @malte-laukoetter. The sh script is well readable and was easy for me to understand why the different XSLT and why the need for the sh script. What it would be great to have a documentation would be the list of changes made by this process like: "splitting metadata into two categories, etc... " I know we have some of this information in confluence and we will probably create JIRA tickets for making the actual changes in the code, but it would be great if we could add all steps being done by this transformation somewhere, probably in the README of the LegalDocML.de folder?

malte-laukoetter:
Some of the created files are still schematron-invalid (eg. they have duplicated eids or missing required metadata), but that requires manual fixing as they where either already violating that schematron-rule or do not have the required information for creating unique eids (eg. multiple articles with an empty akn:num tag, this would require adding a new attribute that disables the zaehlbezeichnung-based creation of eids for those elements; but this only really is a problem for a handful of files)

VictorDelCampo:
@malte-laukoetter you need another hand to do the manual fixing?

malte-laukoetter:
@VictorDelCampo i don't think so, it's not that much as I don't think it is a good investment of time to fix every schematron error and the once we should fix are not that many

PR #1327, opened 2025-05-23
RISDEV-7872 Create affected expressions

VictorDelCampo:


backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/CreateNewVersionOfNormService.java
```diff
@@ -211,6 +262,27 @@ private void setNewExpressionMetadata(Dokument dokument, DokumentExpressionEli e
...
+    Dokument dokument,
+    DokumentExpressionEli expressionEli,
+    UUID currentGuid,
+    UUID nextGuid
+  ) {
+    setNewExpressionMetadata(dokument, expressionEli);
+    var expression = dokument.getMeta().getFRBRExpression();
+    expression.setFRBRaliasCurrentVersionId(currentGuid);
+    if (nextGuid != null) {
+      expression.setFRBRaliasNextVersionId(nextGuid);
```
malte-laukoetter, line 333: should we remove an existing nextVersionId in case `nextGuid == null`?

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/NormService.java
```diff
@@ -418,9 +449,103 @@ private List<NormExpressionEli> collectRelevantExistingExpressions(
...
+            // Create new manifestation and set metadatum to gegenstandslos (date of announcement)
+            final Norm newManifestation = createNewVersionOfNormService.createNewManifestation(
+              norm
+            );
+            // Use announcement date of amending law (part of the work eli)
+            newManifestation.setGegenstandlos(
+              verkuendungNorm.getRegelungstext1().getMeta().getFRBRWork().getFBRDate()
+            );
+            updateOrSaveNormPort.updateOrSave(new UpdateOrSaveNormPort.Options(newManifestation));
+          }
```
malte-laukoetter: do we actually need this part? This should be covered by updating the previous expressions of the non-gegenstandslose expressions

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/NormService.java
```diff
@@ -438,15 +563,30 @@ private Zielnorm createZielNormen(final Zielnorm zielnorm) {
...
+      .map(eli ->
+        new AbstractMap.SimpleEntry<>(eli, loadNormPort.loadNorm(new LoadNormPort.Options(eli)))
+      )
+      .filter(entry -> entry.getValue().isPresent()) // keep only existing Norms
+      .map(entry -> new AbstractMap.SimpleEntry<>(entry.getKey(), entry.getValue().get()))
+      .filter(entry -> !entry.getValue().isGegenstandlos())
+      .max(
+        Comparator.comparing((Map.Entry<NormExpressionEli, Norm> entry) ->
+          entry.getKey().getPointInTime()
+        ).thenComparing(entry -> entry.getKey().getVersion())
```
malte-laukoetter: i think the elis are implementing Comparable and do exactly this sorting by default

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/Proprietary.java
```diff
@@ -225,6 +225,21 @@ public Optional<Gegenstandlos> getGegenstandlos() {
...
+   * Set the gegenstandlos metadata element (if it does not exist, it will create it).
+   * @param date - date to set as gegenstandslos with
+   */
+  public void setGegenstandlos(final String date) {
+    final Gegenstandlos gegenstandslos = getGegenstandlos()
+      .orElseGet(() -> {
+        final Element metadataParent = getOrCreateMetadataParent(Gegenstandlos.NAMESPACE);
+        return new Gegenstandlos(
+          NodeCreator.createElement(Gegenstandlos.NAMESPACE, Gegenstandlos.TAG_NAME, metadataParent)
+        );
```
malte-laukoetter: a lot of our domain classes have a `public static X createAndAppend(Node parentNode)` to keep the logic for creating themself contained within the class (not that much of a difference here, but if there are some additional properties it's helpful to have it in once place)

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/Proprietary.java
```diff
@@ -225,6 +225,21 @@ public Optional<Gegenstandlos> getGegenstandlos() {
     );
   }
 
+  /**
+   * Set the gegenstandlos metadata element (if it does not exist, it will create it).
+   * @param date - date to set as gegenstandslos with
+   */
+  public void setGegenstandlos(final String date) {
```
malte-laukoetter: I'm fine with having this method here but I think I would have added a `Gegenstandlos getOrCreateGegenstandlos()` method here, called that from the `Norm` class and then set the date on the `Gegenstandlos` instance directly

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/Gegenstandlos.java
```diff
@@ -19,4 +19,8 @@ public class Gegenstandlos {
   LocalDate getSinceDate() {
     return LocalDate.parse(element.getAttribute("seit"));
   }
+
+  void updateSinceDate(final String date) {
```
malte-laukoetter: `setSinceDate`?

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/Proprietary.java
VictorDelCampo: I will create a getOrCreate in the Proprietary node because I do see the creation of a node in the parent class and not in the class itself

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/NormService.java
VictorDelCampo: Without changing any code, yes this part is needed, since when the new expression is created (the one replacing this one that becomes gegenstandslos) we just create a new manifestation of the "old" expression for updating the next-version-guid, what in this case is not the correct way since we need to set first the old expression to "gegenstandslos" and actually even remove the next-version-guid right? I think found now by writing this a bug because here the one becoming gegenstandslos is set to gegenstandslos, fine (by creating new manifestatioin), nothing else changes (next-version-guid is not removed). And when the new replacing expressions is created, the previous closest one is that one we set to gegenstandslos and we would be creating a new manifestation and setting the next-version-guid to the guid of the newly created expression....

VictorDelCampo: Completely change the implementation by excluding the isGegenstandlos=true expressions and generating new manifestations set to gegenstandslos when creating the new replacing expression. Discussed with Malte, therefore resolving comment

PR #1335, opened 2025-05-26
RISDEV-6246-nullAway

SebastianRossa:
An examplary implementation to investigate and plan work for null safety using NullAway

### DIFF
--- a/backend/build.gradle.kts
+++ b/backend/build.gradle.kts
@@ -4,10 +4,14 @@ import com.github.jk1.license.filter.DependencyFilter
 import com.github.jk1.license.filter.LicenseBundleNormalizer
 import com.github.jk1.license.render.CsvReportRenderer
 import com.github.jk1.license.render.ReportRenderer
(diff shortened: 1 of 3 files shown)


backend/build.gradle.kts
```diff
@@ -4,10 +4,14 @@ import com.github.jk1.license.filter.DependencyFilter
...
 import com.github.jk1.license.render.CsvReportRenderer
 import com.github.jk1.license.render.ReportRenderer
+import net.ltgt.gradle.errorprone.CheckSeverity
+import net.ltgt.gradle.errorprone.errorprone
+
 
 buildscript { repositories { mavenCentral() } }
 
 plugins {
+    id("net.ltgt.errorprone") version "4.2.0"
```
VictorDelCampo: Can you just use the alias keyword and import from the toml file? Or is the plugin ID not declared in the Gradle plugin portal? If so is there a way to not duplicate the version number (toml and build.gradle.kts)?

PR #1339, opened 2025-05-27
LDML.de 1.8.1

malte-laukoetter:


backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/output/database/service/NormDBService.java
```diff
@@ -94,6 +94,7 @@ public Optional<Norm> updateNorm(UpdateNormPort.Options options) {
       return Optional.empty();
     }
 
+    // TODO: (Malte Laukötter, 2025-06-10) this does not work when the Dokumente or BinaryFiles of a norm change (eg. a new offenestruktur is added)
```
malte-laukoetter: - [x] create bug ticket

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/controller/VerkuendungenController.java
```diff
@@ -43,7 +44,8 @@ public VerkuendungenController(
     LoadVerkuendungUseCase loadVerkuendungUseCase,
     LoadNormExpressionsAffectedByVerkuendungUseCase loadNormExpressionsAffectedByVerkuendungUseCase,
     LoadZielnormenExpressionsUseCase loadZielnormenExpressionsUseCase,
-    CreateZielnormenExpressionsUseCase createZielnormenExpressionsUseCase
+    CreateZielnormenExpressionsUseCase createZielnormenExpressionsUseCase,
```
malte-laukoetter, line 45: - [x] Remove old usecase

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/XsdSchemaService.java
```diff
@@ -34,12 +34,13 @@ public class XsdSchemaService {
   Resource risNormsOffeneStrukturXsdSchema;
   Resource risNormsRechtsetzungsdokumentXsdSchema;
 
+  // TODO: (Malte Laukötter, 2025-05-27) include the missing once (bundesregierung & rechtestzungsdokument metadata)
```
malte-laukoetter: - [x] do this

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/VerkuendungsImportService.java
```diff
@@ -141,9 +141,20 @@ public VerkuendungImportProcess getStatus(
   }
 
   @Override
-  public void processNormendokumentationspaket(
+  public Verkuendung processNormendokumentationspaket(
```
VictorDelCampo, line 149: Why using the same `ProcessNormendokumentationspaketUseCase` but then doing a switch and calling different methods? For me it feels like two different use cases, one is the background process that takes the UUID and the other one is the uploading process which does not work with an UUID but with the ZIP directly. Also specially because the upload needs the _Verkuendung_ returned but the background process does not.

backend/build.gradle.kts
```diff
@@ -252,6 +254,56 @@ tasks {
             into("./LegalDocML.de")
         }
     }
+
+    register("processDBFixtures") {
```
VictorDelCampo, line 262: This is simply so good! Having also XMLs for the seeds and letting gradle create the SQL. Amazing. I have just some observations, written in further comments. It would be though great to document how the flyway seeds are handled somewhere? the backend README?

backend/build.gradle.kts
```diff
@@ -252,6 +254,56 @@ tasks {
...
+
+                var rechtsetzungsdokumentContent = rechtsetzungsdokumentFile.readText(Charsets.UTF_8)
+                // we just look for the first thing looking like a norm expression eli as it should always be the one of the norm, and we don't need to parse the xml this way
+                var eliNormExpression =
+                    "eli/bund/[-a-z0-9]+/\\d{4}/(s[0-9]+[a-zäöüß]*|[0-9]+(-\\d+)?)/\\d{4}-\\d{2}-\\d{2}/\\d+/[a-z]{3}"
+                        .toRegex()
+                        .find(
+                            rechtsetzungsdokumentContent,
+                        )?.value
+                var publishState = "(?<=PUBLISH_STATE:)[A-Z]+".toRegex().find(rechtsetzungsdokumentContent)?.value ?: "PUBLISHED"
```
VictorDelCampo: What is this publish state within the rechsetzungsdokument? Do we write that when publishing? Is this something written as a comment?

backend/build.gradle.kts
```diff
@@ -252,6 +254,56 @@ tasks {
             into("./LegalDocML.de")
         }
     }
+
+    register("processDBFixtures") {
+        var baseDir = fileTree("src/main/resources/db/data")
```
VictorDelCampo: Should all var just be val? They are all immutable right?

backend/build.gradle.kts
```diff
@@ -252,6 +254,56 @@ tasks {
...
+        var outdir = layout.buildDirectory.dir("resources/main/db/data")
+        outputs.dir(outdir)
+
+        doLast {
+            norms.files.forEach { rechtsetzungsdokumentFile ->
+                var relativeNormFolder = rechtsetzungsdokumentFile.parentFile.relativeTo(baseDir.dir)
+                logger.info("Creating db migration for $relativeNormFolder")
+                var dbMigrationFile = file(outdir.get().file("$relativeNormFolder.sql"))
+
+                if (dbMigrationFile.exists()) {
```
VictorDelCampo: All this

```
if (dbMigrationFile.exists()) {
    dbMigrationFile.delete()
}
dbMigrationFile.parentFile.mkdirs()
dbMigrationFile.createNewFile()
```


Could just be replaced with:
```
dbMigrationFile.parentFile.mkdirs()
dbMigrationFile.writeText("")
```

Since it wil create the file if it doesnt exist and if it does it will just be ovewritten, meaning no need to manually delete/create files and this way reducing filesystem operatins

backend/build.gradle.kts
```diff
@@ -252,6 +254,56 @@ tasks {
...
+                dbMigrationFile.createNewFile()
+
+                var rechtsetzungsdokumentContent = rechtsetzungsdokumentFile.readText(Charsets.UTF_8)
+                // we just look for the first thing looking like a norm expression eli as it should always be the one of the norm, and we don't need to parse the xml this way
+                var eliNormExpression =
+                    "eli/bund/[-a-z0-9]+/\\d{4}/(s[0-9]+[a-zäöüß]*|[0-9]+(-\\d+)?)/\\d{4}-\\d{2}-\\d{2}/\\d+/[a-z]{3}"
+                        .toRegex()
+                        .find(
+                            rechtsetzungsdokumentContent,
+                        )?.value
```
VictorDelCampo, line 290: Do we want to throw an error and not have them silently like when no eli is found because we had a completely wrong XML?

backend/build.gradle.kts
```diff
@@ -252,6 +254,56 @@ tasks {
...
+                var publishState = "(?<=PUBLISH_STATE:)[A-Z]+".toRegex().find(rechtsetzungsdokumentContent)?.value ?: "PUBLISHED"
+
+                dbMigrationFile.appendText("DELETE FROM dokumente WHERE eli_norm_expression = '$eliNormExpression';\n")
+                dbMigrationFile.appendText("DELETE FROM norm_manifestation WHERE eli_norm_expression = '$eliNormExpression';\n")
+                dbMigrationFile.appendText("DELETE FROM norm_expression WHERE eli_norm_expression = '$eliNormExpression';\n")
+
+                // This can be used for sql-injections. So please do not include any sql or the character ' in the sample files ^^
```
VictorDelCampo: Haha just a sweet comment. But should we at least escape the single quotes?

frontend/e2e/utils/dataDirectories.ts
```diff
@@ -5,7 +5,7 @@ const baseDir = path.dirname(fileURLToPath(import.meta.url))
 
 export const samplesDirectory = path.resolve(
   baseDir,
-  "../../../LegalDocML.de/1.7.2/samples",
+  "../../../LegalDocML.de/1.8/samples",
```
VictorDelCampo: This should be 1.8.1 right?

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/eid/EIdPosition.java
```diff
@@ -0,0 +1,35 @@
...
+public sealed interface EIdPosition
+  permits OrdinalEIdPosition, ZaehlbezeichnungsbasierteEIdPosition {
+  /**
+   * Find the position information that should be used for the last part of the eId of the element.
+   * @param element the element whose position information should be determined
+   * @param eIdPartType the {@link EIdPartType} of the element
+   * @return the position information for the last part of the eId for the element
+   */
+  static EIdPosition findEIdPosition(Element element, EIdPartType eIdPartType) {
+    if (eIdPartType.equals(EIdPartType.ART) || eIdPartType.equals(EIdPartType.ABS)) {
```
VictorDelCampo, line 18: Shouldn't we check also here if the ART or ABS are actually not within a quotedStructure? Otherwise they would be getting the "z" prefix and we need the "n" prefix right?

backend/build.gradle.kts
malte-laukoetter: its a comment in the xml file, i added an explanaition whenever using the comment but totally forgot to add one here as well, will do that

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/eid/EIdPosition.java
malte-laukoetter: They can use the `z` prefix in a quoted structure as well, there is no restriction on that. The difference within a quoted structure is that akn:article is using the parent elements eId as a prefix (like every other element) and not getting any special treatment for that.

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/VerkuendungsImportService.java
malte-laukoetter: That was my first idea as well, but as i couldn't find a naming that properly explains the difference between the two I got more convinced that it really is the same use case with just different arguments for how to find the data to work on.

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/eid/EIdPosition.java
VictorDelCampo: oh wait i thought the new z-prefixing was specifically only for articles within quotedStructure. It doesn't feel right that the standard is not restricting what prefixing is used when doing the ordinal, specially if the parents would have the n-prefix but then children possibly have the z-prefix?

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/VerkuendungsImportService.java
VictorDelCampo: For me personally felt quite confusing, specially because completely different Options, different return values and different processing (no signature validation). The only thing they share is the parsing, validation and saving, yeah big part of the functionality but not all. But it is just my taste, I don't expect you to separate it now into two different use cases :)

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/eid/EIdPosition.java
malte-laukoetter: the articles (outside of `quotedStructure`) do not use the parent eId stuff at all anymore so they all have eIds like `art-z1`, `art-z3b` or (in case of `@refersTo="ordinale-zaehlung-eid"` `art-n4`). The idea is to make references to articles simpler by no longer needing the whole structure of the norm. 
Even in case of the ordinale-zaehlung (n-prefix) the articles do not include the parent part so if they are eg. in different akn:section elements they still are counted over the whole document.

There are two case in which a z-prefix part can follow one with a n-prefix: either when something is inside a quoted structure or when a paragraph is within an article with `@refersTo="ordinale-zaehlung-eid"`.

backend/build.gradle.kts
malte-laukoetter: i throw an error when there is a ' in the content instead, escaping is not that simple and can quickly need to new problems (like what if the escape sequence is in there and so on)

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/eid/EIdPosition.java
VictorDelCampo: I just realized that this new _zählbezeichnung_ positioning is to be used for all articles and paragraphs, both inside and outside quotedStructure. Meaning my comment was just wrong

PR #1388, opened 2025-06-11
Handle orphaned expressions - frontend

andreasphil:
RISDEV-8137


frontend/e2e/testData/aenderungsgesetz-with-orphaned-amended-norm-expressions.xml
```diff
@@ -0,0 +1,191 @@
...
+          <meta-breg:federfuehrung>
+            <meta-breg:federfuehrend ab="2002-12-01" bis="unbestimmt">Bundesministerium des Innern und für Heimat</meta-breg:federfuehrend>
+            <meta-breg:federfuehrend ab="2002-10-01" bis="2002-12-01">Bundesministerium der Justiz</meta-breg:federfuehrend>
+          </meta-breg:federfuehrung>
+        </meta-breg:legalDocML.de_metadaten>
+        <ris:legalDocML.de_metadaten xmlns:ris="http://MetadatenRIS.LegalDocML.de/1.7.2/">
+          <norms:legalDocML.de_metadaten xmlns:norms="http://MetadatenMods.LegalDocML.de/1.7.2/">
+            <norms:geltungszeiten>
+              <norms:geltungszeit art="inkraft" id="8c5b166b-ee37-4185-9b7a-96fc16031e62">2017-03-16</norms:geltungszeit>
+              <norms:geltungszeit art="inkraft" id="e2420f07-c155-41d5-8c11-76a3e357a261">2017-05-01</norms:geltungszeit>
```
VictorDelCampo: I would remove this time boundary to make it more clear that is not being used, even if it is not being used in the zielnorm references

backend/src/main/resources/db/data/e2e/R__010_bgbl-1_1964_321.sql
```diff
@@ -364,7 +364,7 @@ VALUES ('<?xml version="1.0" encoding="UTF-8"?>
 </akn:akomaNtoso>
 ');
 
-UPDATE norm_manifestation SET publish_state = 'PUBLISHED' WHERE eli_norm_expression = 'eli/bund/bgbl-1/1964/321/2017-03-16/1/deu';
+UPDATE norm_manifestation SET publish_state = 'UNPUBLISHED' WHERE eli_norm_expression = 'eli/bund/bgbl-1/1964/321/2017-03-16/1/deu';
```
VictorDelCampo, line 367: Changing an existing and possibly already applied (other devs locally) migration will lead to a failed start-up of the app because the checksum changed. For me it is fine, just communicate with the dev team after merging that they should clean their db, or at least delete that migration of the flyway schema history

backend/src/main/resources/db/data/e2e/R__012_bgbl-1_1964_654.sql
```diff
@@ -0,0 +1,551 @@
...
+        </akn:signature>
+      </akn:blockContainer>
+    </akn:conclusions>
+  </akn:act>
+</akn:akomaNtoso>
+');
+
+UPDATE norm_manifestation SET publish_state = 'UNPUBLISHED' WHERE eli_norm_expression = 'eli/bund/bgbl-1/1964/654/2017-03-16/1/deu';
+
+-- Future expression of target law for 2017-05-01
```
VictorDelCampo, line 553: This is the orphan expression that will be deleted. Great. But I would add a second orphan expression. Why? You don't know that but the backend makes a difference between orphans that are before the first time boundary of the verkuendung for the target norm and those that are after. If you have time I would create this setup:
- Amending law with:
-- only 1 time boundary (of the one expressiont that is not orphan)
-- with only 1 zielnorm-reference (using that 1 time boundary)
-- three amended-norm-expressions (first should be with a date before the 1 time boundary, second would be the actual existing expression with that 1 time boundary and thirdly the 2nd orphan with a date after the 1 time boundary)
- E2E seed with the 3 expressions of amended-norm-expressions (or at least those that are orphan and should be deleted)

-
-

VictorDelCampo (approved):
LGMT! Just some nitpicky things

PR #1416, opened 2025-06-18
Add port for finding out if norm exists without loading it

malte-laukoetter:
This hopefully reduces the db-requests for loading the list for expressionen-erzeugen quite a bit. (Especially as we do not need to load the Dokumente and Binary Files for this step)

RISDEV-0000


malte-laukoetter:
Used a different approach: https://github.com/digitalservicebund/ris-norms/pull/1417

PR #1458, opened 2025-06-25
Add paginated endpoint for getting works

malte-laukoetter:
RISDEV-8420


doc/backend/norms-api-v1.yaml
```diff
@@ -207,6 +208,48 @@ paths:
...
+      tags:
+        - Norm work
+      summary: Get a list of all norm works
+      parameters:
+        - name: page
+          in: query
+          description: The page of the results
+          required: false
+          schema:
+            type: number
```
VictorDelCampo, line 222: Both `page` and `size` should be of type `integer`, not `number`

doc/backend/norms-api-v1.yaml
```diff
@@ -207,6 +208,48 @@ paths:
...
+                properties:
+                  content:
+                    type: array
+                    items:
+                      $ref: "#/components/schemas/NormWork"
+                  page:
+                    type: object
+                    properties:
+                      size:
+                        type: number
```
VictorDelCampo, line 245: Also all this properties in the `page` object should be of type `integer`, not `number`

doc/adr/0019-pagination-for-works.md
```diff
@@ -0,0 +1,72 @@
...
+
+Spring's Pageable also allows providing sorting information. We will not use this as that would expose our database
+column names to the rest-API.
+
+By using the native spring naming for these parameters we keep our project more idiomatic and consistent and ensure that
+the meaning of the fields is simple to understand.
+
+## Decision
+
+We use offset-based pagination. We use the nativ spring representation for a page and for the query parameters.
```
VictorDelCampo, line 59: typo "nativ spring representation" → "native Spring representation"

doc/adr/0019-pagination-for-works.md
```diff
@@ -0,0 +1,72 @@
...
+    "size": 20,
+    "number": 0,
+    "totalElements": 15,
+    "totalPages": 1
+  }
+}
+```
+
+Spring's Pageable also allows providing sorting information. We will not use this as that would expose our database
+column names to the rest-API.
```
VictorDelCampo, line 52: type and capitalize "rest-API" → "REST API"

backend/src/main/resources/application-local.yaml
```diff
@@ -4,7 +4,7 @@ spring:
     username: test
     password: test
   flyway:
-    locations: classpath:db/migration,classpath:db/data/real,classpath:db/data/synthetic
+    locations: classpath:db/migration,classpath:db/data/real,classpath:db/data/synthetic,classpath:db/data/e2e
```
VictorDelCampo, line 7: Why adding this? We said that we want to decide locally with which profile to run the backend and with this change and only the local profile, we will always be importing also the e2e test data. I guess you have it set locally this way and commit the change without realizing?

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/controller/NormController.java
```diff
@@ -0,0 +1,45 @@
...
+    this.loadNormWorksUseCase = loadNormWorksUseCase;
+  }
+
+  /**
+   * Get a list of all norms (work-level)
+   * @param pageable pagination options to only load a subset of all norms at a time
+   * @return a paged list of information about norms
+   */
+  @GetMapping(produces = { APPLICATION_JSON_VALUE })
+  public PagedModel<NormWorkResponseSchema> getNorm(Pageable pageable) {
```
VictorDelCampo, line 33: Should we set here in the backend the default of 100 size stated in the ticket?

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/output/database/repository/NormManifestationRepository.java
```diff
@@ -96,4 +99,10 @@ void updatePublishStateByManifestationEli(
     @Param("manifestationEli") String manifestationEli,
     @Param("publishState") NormPublishState publishState
   );
+
+  @NativeQuery(
+    value = "select DISTINCT ON (eli_norm_work) * from norm_manifestation ORDER BY eli_norm_work, eli_norm_manifestation DESC",
```
VictorDelCampo, line 104: Why do we need to also order DESC by eli_norm_manifestation? And I would also have done a DESC order for the eli_norm_work but the tickets says ASC? Also the method name says ASC :)

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/output/database/repository/NormManifestationRepository.java
```diff
@@ -96,4 +99,10 @@ void updatePublishStateByManifestationEli(
...
+    countQuery = "select COUNT(DISTINCT eli_norm_work) from norm_manifestation"
+  )
+  Page<NormManifestationDto> findDistinctOnWorkEliByOrderByWorkEliAsc(Pageable pageable);
```
VictorDelCampo, line 107: Didn't we think about also extracting the title on db level so that we can just do here a simple give me the work elis and the title of their latest expression/manifestation? I bet this query will take a lot if we load all dokumente and the XMLs of all of them (up to 100)

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/output/database/repository/NormManifestationRepository.java
malte-laukoetter: We need to sort the expressions of a work descendingly so we load the title from the newest expression (and then also the newest manifestation of that expression) Thats what `eli_norm_manifestation DESC` takes care of. The ascending order of the works is taken care of by `ORDER BY eli_norm_work` as that defaults to `ASC`.
I add in the `ASC` explicitly so it is clearer what is happening here.

backend/src/main/resources/application-local.yaml
malte-laukoetter: Oh yes i committed that accidentally. Needed to add that part locally so I could run the e2e tests. 

Also I'm not sure what the advantage of not having the e2e data in the local database is. We do not import any additional data there and we need it to run e2e tests.

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/output/database/repository/NormManifestationRepository.java
malte-laukoetter: Yes this implementation is quite inefficient, but that was on purpose to first get something out before doing any kind of performance work. Thats also why I immediately created https://digitalservicebund.atlassian.net/browse/RISDEV-8447 after I tested that endpoint and saw that a request takes around 6s.

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/controller/NormController.java
malte-laukoetter: I think that is something the frontend should take care of using the `size`-parameter. That way it's only defined in one place.

backend/src/main/resources/application-local.yaml
VictorDelCampo: This is a discussion we can have in the next dev weekly. If we agree to just have 1 profile for local development let's do it. For me personally i prefer having two profiles, since when I do manual testing locally I don't want to have too much test data in case we have again situations of what I called data contamination (targeting same norm, creating expressions etc)

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/controller/NormController.java
VictorDelCampo: I will make a comment in the frontend subtask, so that we don't forget.

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/output/database/repository/NormManifestationRepository.java
VictorDelCampo: I guess you'll be reworking how we retrieve the data in your new ticket but the current SQL does not seem correct to me. Since the `ORDER BY eli_norm_work` is in the inner SQL and yes it defaults to ASC but since it is in the inner SQL doing the distinct, the order by only determines which row to pick, but does not guarantee the final result is ordered by eli_norm_work. For that we would need to do the order in the outer SQL

malte-laukoetter: oh your right it's not guaranteed to work (but does accidentally), I added a second order by into the outer query on the version i created in the new ticket

PR #1463, opened 2025-06-26
feat: add XML formatting advice and integration tests

hamo225:
- implement XmlFormattingResponseBodyAdvice to format all API XML responses
- add unit and integration tests
- update norm controller test expectations to handle formatted/self-closing XML

RISDEV-8144

### DIFF
--- a/backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/formatter/XmlFormattingResponseBodyAdvice.java
+++ b/backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/formatter/XmlFormattingResponseBodyAdvice.java
@@ -1,25 +1,27 @@
 package de.bund.digitalservice.ris.norms.adapter.input.restapi.formatter;
 
-import java.io.StringReader;
+import de.bund.digitalservice.ris.norms.utils.XmlMapper;
 import java.io.StringWriter;
-import javax.xml.XMLConstants;
-import javax.xml.parsers.DocumentBuilderFactory;
+import lombok.extern.slf4j.Slf4j;
+import org.jspecify.annotations.NullMarked;
 import org.springframework.core.MethodParameter;
 import org.springframework.http.MediaType;
 import org.springframework.http.converter.HttpMessageConverter;
 import org.springframework.http.server.ServerHttpRequest;
 import org.springframework.http.server.ServerHttpResponse;
+import org.springframework.lang.Nullable;
 import org.springframework.web.bind.annotation.ControllerAdvice;
 import org.springframework.web.servlet.mvc.method.annotation.ResponseBodyAdvice;
 import org.w3c.dom.Document;
 import org.w3c.dom.ls.DOMImplementationLS;
 import org.w3c.dom.ls.LSOutput;
 import org.w3c.dom.ls.LSSerializer;
-import org.xml.sax.InputSource;
 
 /**
  * ResponseBodyAdvice that formats XML responses
  */
+@NullMarked
+@Slf4j
 @ControllerAdvice
 public class XmlFormattingResponseBodyAdvice implements ResponseBodyAdvice<Object> {
 
@@ -28,12 +30,13 @@ public class XmlFormattingResponseBodyAdvice implements ResponseBodyAdvice<Objec
     MethodParameter returnType,
     Class<? extends HttpMessageConverter<?>> converterType
   ) {
-    return true;
+    return returnType.getParameterType() == String.class;
   }
 
+  @Nullable
   @Override
   public Object beforeBodyWrite(
-    Object body,
+    @Nullable Object body,
     MethodParameter returnType,
     MediaType selectedContentType,
     Class<? extends HttpMessageConverter<?>> selectedConverterType,
@@ -41,7 +44,6 @@ public class XmlFormattingResponseBodyAdvice implements ResponseBodyAdvice<Objec
     ServerHttpResponse response
   ) {
     if (
-      selectedContentType != null &&
       selectedContentType.isCompatibleWith(MediaType.APPLICATION_XML) &&
       body instanceof String xmlBody
     ) {
@@ -52,12 +54,7 @@ public class XmlFormattingResponseBodyAdvice implements ResponseBodyAdvice<Objec
 
   private String formatXml(String xml) {
     try {
-      DocumentBuilderFactory dbf = DocumentBuilderFactory.newInstance();
-      dbf.setNamespaceAware(true);
-      dbf.setFeature("http://apache.org/xml/features/disallow-doctype-decl", true);
-      dbf.setFeature(XMLConstants.FEATURE_SECURE_PROCESSING, true);
-      dbf.setExpandEntityReferences(false);
-      Document document = dbf.newDocumentBuilder().parse(new InputSource(new StringReader(xml)));
+      Document document = XmlMapper.toDocument(xml);
 
       DOMImplementationLS impl = (DOMImplementationLS) document
         .getImplementation()
@@ -74,6 +71,7 @@ public class XmlFormattingResponseBodyAdvice implements ResponseBodyAdvice<Objec
       serializer.write(document, lsOutput);
       return writer.toString();
     } catch (Exception e) {
+      log.error("Failed to format XML response", e);
       return xml;
     }
   }
--- a/backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/formatter/XmlFormattingResponseBodyAdvice.java
+++ b/backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/formatter/XmlFormattingResponseBodyAdvice.java
@@ -62,7 +62,6 @@ public class XmlFormattingResponseBodyAdvice implements ResponseBodyAdvice<Objec
       LSSerializer serializer = impl.createLSSerializer();
 
       serializer.getDomConfig().setParameter("format-pretty-print", Boolean.TRUE);
-      serializer.getDomConfig().setParameter("xml-declaration", Boolean.TRUE);
 
       LSOutput lsOutput = impl.createLSOutput();
       StringWriter writer = new StringWriter();
--- a/backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/formatter/XmlFormattingResponseBodyAdvice.java
+++ b/backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/formatter/XmlFormattingResponseBodyAdvice.java
@@ -30,7 +30,7 @@ public class XmlFormattingResponseBodyAdvice implements ResponseBodyAdvice<Objec
     MethodParameter returnType,
     Class<? extends HttpMessageConverter<?>> converterType
   ) {
-    return returnType.getParameterType() == String.class;
+    return true;
   }
 
   @Nullable
--- a/backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/formatter/XmlFormattingResponseBodyAdvice.java
+++ b/backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/formatter/XmlFormattingResponseBodyAdvice.java
@@ -1,0 +1,80 @@
+package de.bund.digitalservice.ris.norms.adapter.input.restapi.formatter;
+
+import java.io.StringReader;
+import java.io.StringWriter;
+import javax.xml.XMLConstants;
+import javax.xml.parsers.DocumentBuilderFactory;
+import org.springframework.core.MethodParameter;
+import org.springframework.http.MediaType;
+import org.springframework.http.converter.HttpMessageConverter;
+import org.springframework.http.server.ServerHttpRequest;
+import org.springframework.http.server.ServerHttpResponse;
+import org.springframework.web.bind.annotation.ControllerAdvice;
+import org.springframework.web.servlet.mvc.method.annotation.ResponseBodyAdvice;
+import org.w3c.dom.Document;
+import org.w3c.dom.ls.DOMImplementationLS;
+import org.w3c.dom.ls.LSOutput;
+import org.w3c.dom.ls.LSSerializer;
+import org.xml.sax.InputSource;
+
+/**
+ * ResponseBodyAdvice that formats XML responses
+ */
+@ControllerAdvice
+public class XmlFormattingResponseBodyAdvice implements ResponseBodyAdvice<Object> {
+
+  @Override
+  public boolean supports(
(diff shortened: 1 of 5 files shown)


backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/formatter/XmlFormattingResponseBodyAdvice.java
```diff
@@ -0,0 +1,80 @@
...
+import org.w3c.dom.Document;
+import org.w3c.dom.ls.DOMImplementationLS;
+import org.w3c.dom.ls.LSOutput;
+import org.w3c.dom.ls.LSSerializer;
+import org.xml.sax.InputSource;
+
+/**
+ * ResponseBodyAdvice that formats XML responses
+ */
+@ControllerAdvice
```
VictorDelCampo, line 25: I would mark the whole new class with _@NullMarked_, as we wanted to do for new classes, following the nullability approach that Sebastian presented a while ago. This way the whole class will be considered not nullable in regards of the fields and input parameters of the methods.
This is a good change to use it BECAUSE the package of spring where the interface ResponseBodyAdvance resides is marked by Spring as `@NonNullApi`, meaning that effectively all parameters of both methods are implicitly non nullable by contract
You must have seen some soner warnings locally at least complaining about all non-nullable parameters with something like _Not annotated parameter overrides @NonNullApi parameter_

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/formatter/XmlFormattingResponseBodyAdvice.java
```diff
@@ -0,0 +1,80 @@
...
+  public boolean supports(
+    MethodParameter returnType,
+    Class<? extends HttpMessageConverter<?>> converterType
+  ) {
+    return true;
+  }
+
+  @Override
+  public Object beforeBodyWrite(
+    Object body,
```
VictorDelCampo: This is the only parameter marked in the interface as @Nullable, so mark it also here. From here:
`import org.springframework.lang.Nullable;`

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/formatter/XmlFormattingResponseBodyAdvice.java
```diff
@@ -0,0 +1,80 @@
...
+ */
+@ControllerAdvice
+public class XmlFormattingResponseBodyAdvice implements ResponseBodyAdvice<Object> {
+
+  @Override
+  public boolean supports(
+    MethodParameter returnType,
+    Class<? extends HttpMessageConverter<?>> converterType
+  ) {
+    return true;
```
VictorDelCampo, line 33: We could already restrict here what types should be processed by this advice, because otherwise it will apply to all responses, including non-XML (though the beforeBodyWrite later is anyway restricting already by media type). And the type we need would be only String, so something like:
`return returnType.getParameterType() == String.class;`

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/formatter/XmlFormattingResponseBodyAdvice.java
```diff
@@ -0,0 +1,80 @@
...
+      serializer.getDomConfig().setParameter("xml-declaration", Boolean.FALSE);
+
+      LSOutput lsOutput = impl.createLSOutput();
+      StringWriter writer = new StringWriter();
+      lsOutput.setCharacterStream(writer);
+
+      serializer.write(document, lsOutput);
+      return writer.toString();
+    } catch (Exception e) {
+      return xml;
```
VictorDelCampo, line 74: Please add logging in case of an exception, otherwise will always fail silently. For that add at the class level the annotation `@Slf4j` and then add a line with `log.error(`) and some descriptive message. Thanks!

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/formatter/XmlFormattingResponseBodyAdvice.java
```diff
@@ -0,0 +1,80 @@
...
+  public Object beforeBodyWrite(
+    Object body,
+    MethodParameter returnType,
+    MediaType selectedContentType,
+    Class<? extends HttpMessageConverter<?>> selectedConverterType,
+    ServerHttpRequest request,
+    ServerHttpResponse response
+  ) {
+    if (
+      selectedContentType != null &&
```
VictorDelCampo: You don't need the checking if selectedContentType is null, once you have marked the class as @NullMarked

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/formatter/XmlFormattingResponseBodyAdvice.java
```diff
@@ -0,0 +1,80 @@
...
+      body instanceof String xmlBody
+    ) {
+      return formatXml(xmlBody);
+    }
+    return body;
+  }
+
+  private String formatXml(String xml) {
+    try {
+      DocumentBuilderFactory dbf = DocumentBuilderFactory.newInstance();
```
VictorDelCampo: To get the `Document` out of the string you could just re-use our `XmlMapper` class, method `toDocument(string)`

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/formatter/XmlFormattingResponseBodyAdvice.java
```diff
@@ -0,0 +1,80 @@
...
+      dbf.setExpandEntityReferences(false);
+      Document document = dbf.newDocumentBuilder().parse(new InputSource(new StringReader(xml)));
+
+      DOMImplementationLS impl = (DOMImplementationLS) document
+        .getImplementation()
+        .getFeature("LS", "3.0");
+      LSSerializer serializer = impl.createLSSerializer();
+
+      serializer.getDomConfig().setParameter("format-pretty-print", Boolean.TRUE);
+      serializer.getDomConfig().setParameter("xml-declaration", Boolean.FALSE);
```
VictorDelCampo: Interesting line.... and I just realized that the migrated data does not include the XML declaration but our seeds do. With this line all XMLs returned through our API won't include the XML declaration and I am right now not sure what do we want...

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/formatter/XmlFormattingResponseBodyAdvice.java
```diff
@@ -0,0 +1,80 @@
...
+
+  @Override
+  public boolean supports(
+    MethodParameter returnType,
+    Class<? extends HttpMessageConverter<?>> converterType
+  ) {
+    return true;
+  }
+
+  @Override
```
VictorDelCampo, line 37: You would also need to mark this method as @Nullable

backend/src/test/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/controller/NormExpressionControllerTest.java
```diff
@@ -126,7 +126,7 @@ class getNormXml {
     void itCallsLoadNormXmlAndReturnsNormXml() throws Exception {
       // Given
       final String eli = "eli/bund/bgbl-1/1990/s2954/2022-12-19/1/deu/regelungstext-verkuendung-1";
-      final String xml = "<target></target>";
+      final String xml = "</target>";
```
VictorDelCampo, line 129: Why this changed was introduced? Is not wrong but I am wondering

backend/src/test/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/formatter/XmlFormattingResponseBodyAdviceTest.java
```diff
@@ -0,0 +1,62 @@
...
+class XmlFormattingResponseBodyAdviceTest {
+
+  final XmlFormattingResponseBodyAdvice xmlFormatter = new XmlFormattingResponseBodyAdvice();
+
+  @Test
+  void shouldReturnTransformedXml() {
+    String compactXml =
+      "<?xml version=\"1.0\" encoding=\"UTF-8\"?><akn:doc><akn:meta><akn:identification><akn:FRBRWork><akn:FRBRthis value=\"test\"/></akn:FRBRWork></akn:identification></akn:meta><akn:body><akn:article eId=\"art-1\"><akn:heading>Test</akn:heading></akn:article></akn:body></akn:doc>";
+
+    var result = xmlFormatter.beforeBodyWrite(
```
VictorDelCampo: Not sure if the unit test now will work without problems if we introduced the NullMarked at the class level.. if so we would need to create dummy mock object slike:

import static org.mockito.Mockito.mock;

```
Class<? extends HttpMessageConverter<?>> dummyConverterType = HttpMessageConverter.class;
ServerHttpRequest dummyRequest = mock(ServerHttpRequest.class);
ServerHttpResponse dummyResponse = mock(ServerHttpResponse.class);
```

frontend/e2e/application/references.spec.ts
```diff
@@ -23,18 +23,18 @@ test("should be able to add a new ref and edit it's refersTo and href and delete
     .getByRole("region", { name: "Textbasierte Metadaten" })
     .getByRole("textbox")
 
-  const textElement = container.getByText(/Zur Unterstützung der/)
+  const textElement = container.getByText(/Nach Ablauf der/)
```
VictorDelCampo, line 26: I am confused, why did you need to modify this E2E tests?

backend/src/test/java/de/bund/digitalservice/ris/norms/integration/adapter/input/restapi/XmlFormattingResponseBodyAdviceIntegrationTest.java
```diff
@@ -0,0 +1,78 @@
...
+
+  @AfterEach
+  void cleanUp() {
+    dokumentRepository.deleteAll();
+    binaryFileRepository.deleteAll();
+    normManifestationRepository.deleteAll();
+  }
+
+  @Test
+  void itReturnsPrettyPrintedXmlEvenIfStoredUgly() throws Exception {
```
VictorDelCampo, line 46: Why are you loading and saving the whole norm with eli/bund/bgbl-1/1964/s593/1964-08-05/1/deu/1964-08-05 ? This already includes the regelungstext under eli/bund/bgbl-1/1964/s593/1964-08-05/1/deu/1964-08-05/regelungstext-verkuendung-1.xml
The endpoint you use in the test /api/v1/norms/{eli}. accepts a dokumentExpressionEli, meaning a `eli/bund/bgbl-1/1964/s593/1964-08-05/1/deu/1964-08-05/regelungstext-verkuendung-1` (without the .xml at the end)
So what you would need is loading the regelungstext, transforming it into a DokumentDto and then ugly its xml before saving it into the db. Then you make the API call and test the response against the before-uglyed XML

```
    Regelungstext prettyXml = Fixtures.loadRegelungstextFromDisk(
      "eli/bund/bgbl-1/1964/s593/1964-08-05/1/deu/1964-08-05/regelungstext-verkuendung-1.xml"
    );
    DokumentDto dokumentDto = DokumentMapper.mapToDto(prettyXml);
    dokumentDto.setXml(dokumentDto.getXml().replaceAll(">\\s+<", "><").replaceAll("\\r?\\n", ""));
    dokumentRepository.save(dokumentDto);
```

Then of course at the end checking, ignoring whitespaces, both XMLs, the API response and the actual XML before you modify it

VictorDelCampo (dismissed):
Nice initial work in the backend Tarek! I made a thorough review and wrote many comments with lots of information. Read them through and let me know if you wanna a 1on1 sync on the topics :)

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/formatter/XmlFormattingResponseBodyAdvice.java
hamo225: @VictorDelCampo  Should they be the same? If there’s no strong requirement, maybe keeping it matching the migrated data (no declaration) makes things more consistent

backend/src/test/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/controller/NormExpressionControllerTest.java
hamo225: As far as I can remember, there was a test rule complaining that empty XML elements shouldn’t use separate open/close tags, so I just switched it to a self-closing form to make it pass

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/formatter/XmlFormattingResponseBodyAdvice.java
VictorDelCampo: I would just keep that xml-declaration parameter set to FALSE

frontend/e2e/application/references.spec.ts
VictorDelCampo: @hamo225 i still do not understand why you needed to adapt this E2E test

backend/src/test/java/de/bund/digitalservice/ris/norms/integration/adapter/input/restapi/XmlFormattingResponseBodyAdviceIntegrationTest.java
```diff
@@ -0,0 +1,80 @@
...
+    Regelungstext originalText = Fixtures.loadRegelungstextFromDisk(
+      "eli/bund/bgbl-1/1964/s593/1964-08-05/1/deu/1964-08-05/regelungstext-verkuendung-1.xml"
+    );
+    DokumentDto dto = DokumentMapper.mapToDto(originalText);
+    dokumentRepository.save(dto);
+
+    String formattedXml = mockMvc
```
VictorDelCampo: You don't need to call an endpoint to get the XML of the norm. the `DokumentDto` itself has the field `xml`, containing the xml in a string form. This is how I proposed it to you before:

`dokumentDto.setXml(dokumentDto.getXml().replaceAll(">\\s+<", "><").replaceAll("\\r?\\n", ""));`

PR #1481, opened 2025-06-30
Risdev 8465 add endpoint to load information title about a work

malte-laukoetter:


backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/controller/NormWorkController.java
```diff
@@ -0,0 +1,40 @@
...
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+/**
+ * Controller for getting all norms
+ */
+@RestController
+@RequestMapping("/api/v1/norms/eli/bund/{agent}/{year}/{naturalIdentifier}")
+public class NormWorkController {
```
VictorDelCampo, line 19: I think I can imagine your thinking for creating two separate controllers `NormController` and `NormWorkController` but it feels not completely right, specially because the `NormController` is using the `NormWorkResponseSchema`, thought it should be fine. In the `NormController` we have an endpoint handling works and for me it belongs to the `NormWorkController` but probably you would argument that the give me all works (paginated) is not just handling 1 work and therefore belongs to a general controller for norms. No action need IMO, just wanted to share it with you

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/controller/NormWorkController.java
```diff
@@ -0,0 +1,40 @@
...
+import de.bund.digitalservice.ris.norms.adapter.input.restapi.schema.NormWorkResponseSchema;
+import de.bund.digitalservice.ris.norms.application.port.input.LoadNormUseCase;
+import de.bund.digitalservice.ris.norms.domain.entity.eli.NormWorkEli;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+/**
+ * Controller for getting all norms
```
VictorDelCampo, line 15: But then you should change this description :) (The "but" is because I wrote the other bigger comment first)

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/controller/NormWorkController.java
malte-laukoetter: I thought about this as well and decided against combining the two. Mostly for not losing the convenience of having the `eli/bund/{agent}/{year}/{naturalIdentifier}` path part on the controller level on not needing to define it for every work-level endpoint. But also yes I would also argue that getting a list of norm works is something different than working with one work. Similarly I will place `GET /norms/{workEli}/expressions` into the NormWorkController even-though it returns information about expressions.
Maybe a better name for the `NormController` would be `NormsController` 🤔

VictorDelCampo: Yeah i think that is the main (convenience) argument, having the eli parts defined at the controller level and not at the endpoints level... and I know you are not gonna like it (don't really like it either) but what do you think of having a nested class within the NormController --> NormWorkController to which you can define the base path at the class level? Sounds nasty when I think of testing

malte-laukoetter: Hm.. yeah i think that will be inconvenient with other things like testing and also if take that pattern and apply it to all our controllers i think we would end up with way too large controller classes. If our controller package ever gets too large we can still organise it into sub-packages

PR #1483, opened 2025-06-30
Feature/bestand navbar and route

hamo225:
RISDEV-8453

### DIFF
--- a/frontend/src/components/RisNavbar.vue
+++ b/frontend/src/components/RisNavbar.vue
@@ -1,11 +1,28 @@
 <script setup lang="ts">
 import neurisLogo from "@/assets/neuRIS-logo.svg"
 import { useAuthentication } from "@/lib/auth"
-import { RouterLink } from "vue-router"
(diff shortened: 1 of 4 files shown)


frontend/src/components/RisNavbar.vue
```diff
@@ -1,11 +1,28 @@
...
+  } else if (route.path.startsWith("/bestand")) {
+    return "bestand"
+  }
+
+  return "verkuendungen"
+})
+
+const tabs = [
+  { id: "verkuendungen", label: "Verkündungen" },
+  { id: "bestand", label: "Bestand" },
```
malte-laukoetter: how about `route` property in here for every tab? that way we would only need to update one spot if we have a need to add another route (or change its path) and not touch 3 places (and might miss one).

malte-laukoetter (approved):

frontend/src/components/RisNavbar.vue
hamo225: good suggestion thanks!

PR #1491, opened 2025-07-02
Adapted proposal for ADR 0020

VictorDelCampo:


doc/adr/0020-extending-amended-expressions-for-orphan-removal-process.md
```diff
@@ -1,53 +1,81 @@
...
-- NEW: remove the orphan if the _zielnorm_ reference is gone (current state)
-- SYSTEM: remove the orphan only if there are no newly created expressions before (within `norms:amended-norm-expressions`)
-- REPLACING: remove the orphan only if the _zielnorm_ reference is gone AND there is no newly created expressions before (meaning combining conditions of NEW and SYSTEM)
+1. If `created-by-zeitgrenze=true` and `created-by-replacing-existing-expression=false` and the _zeitgrenze_ is no
+   longer present under the _zielnorm_ references, the entry is deleted
+2. If `created-by-zeitgrenze=false` and `created-by-replacing-existing-expression=true` and there are no preceding
+   expressions with `created-by-zeitgrenze=true`, the entry is deleted
+3. If `created-by-zeitgrenze=true` and `created-by-replacing-existing-expression=true` and neither the _zeitgrenze_ is
+   present under the _zielnorm_ references nor are there preceding expressions with `created-by-zeitgrenze=true`, the
+   entry is deleted
```
malte-laukoetter: 4. If `created-by-zeitgrenze=true` and `created-by-replacing-existing-expression=true` and the _zeitgrenze_ is not present under the _zielnorm_ references but there are preceding expressions with `created-by-zeitgrenze=true`, the entry is changed to `created-by-zeitgrenze=false` and `created-by-replacing-existing-expression=true`

malte-laukoetter: everything else looks good :)

malte-laukoetter: or do we also need 

> 5.  If `created-by-zeitgrenze=true` and `created-by-replacing-existing-expression=true` and the _zeitgrenze_ is present under the _zielnorm_ references but there are no preceding expressions with `created-by-zeitgrenze=true`, the entry is changed to `created-by-zeitgrenze=true` and `created-by-replacing-existing-expression=false`

I'm not 100% sure right now

VictorDelCampo: Agree with 4. but definitely not with 5. If we still have the _zeitgrenze_ we don't really care if there are preceding expressions or not. Since that expression is overriding still the already existing expression

PR #1492, opened 2025-07-02
Add endpoint to laod expressions of work

malte-laukoetter:
RISDEV-8464

### DIFF
--- a/backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/controller/NormWorkController.java
+++ b/backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/controller/NormWorkController.java
@@ -2,10 +2,14 @@ package de.bund.digitalservice.ris.norms.adapter.input.restapi.controller;
 
 import static org.springframework.http.MediaType.APPLICATION_JSON_VALUE;
 
+import de.bund.digitalservice.ris.norms.adapter.input.restapi.mapper.NormExpressionListResponseMapper;
 import de.bund.digitalservice.ris.norms.adapter.input.restapi.mapper.NormWorkResponseMapper;
+import de.bund.digitalservice.ris.norms.adapter.input.restapi.schema.NormExpressionListResponseSchema;
 import de.bund.digitalservice.ris.norms.adapter.input.restapi.schema.NormWorkResponseSchema;
+import de.bund.digitalservice.ris.norms.application.port.input.LoadExpressionsOfNormWorkUseCase;
 import de.bund.digitalservice.ris.norms.application.port.input.LoadNormUseCase;
 import de.bund.digitalservice.ris.norms.domain.entity.eli.NormWorkEli;
+import java.util.List;
 import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.RequestMapping;
@@ -19,9 +23,14 @@ import org.springframework.web.bind.annotation.RestController;
 public class NormWorkController {
 
   private final LoadNormUseCase loadNormUseCase;
+  private final LoadExpressionsOfNormWorkUseCase loadExpressionsOfNormWorkUseCase;
 
-  public NormWorkController(LoadNormUseCase loadNormUseCase) {
+  public NormWorkController(
+    LoadNormUseCase loadNormUseCase,
+    LoadExpressionsOfNormWorkUseCase loadExpressionsOfNormWorkUseCase
+  ) {
     this.loadNormUseCase = loadNormUseCase;
+    this.loadExpressionsOfNormWorkUseCase = loadExpressionsOfNormWorkUseCase;
   }
 
   /**
@@ -36,5 +45,19 @@ public class NormWorkController {
         loadNormUseCase.loadNorm(new LoadNormUseCase.EliOptions(eli))
       )
     );
+  }
+
+  /**
+   * Get a list of all expression of this work
+   * @param eli the eli of the work
+   * @return a list of information about expressions
+   */
+  @GetMapping(path = "/expressions", produces = { APPLICATION_JSON_VALUE })
+  public List<NormExpressionListResponseSchema> getExpressions(NormWorkEli eli) {
+    return loadExpressionsOfNormWorkUseCase
+      .loadExpressionsOfNormWork(new LoadExpressionsOfNormWorkUseCase.Options(eli))
+      .stream()
+      .map(NormExpressionListResponseMapper::fromUseCaseData)
+      .toList();
   }
 }
--- a/backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/schema/NormExpressionListResponseSchema.java
+++ b/backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/schema/NormExpressionListResponseSchema.java
@@ -1,0 +1,19 @@
+package de.bund.digitalservice.ris.norms.adapter.input.restapi.schema;
+
+import de.bund.digitalservice.ris.norms.domain.entity.Norm;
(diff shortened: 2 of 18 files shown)


backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/controller/NormWorkController.java
```diff
@@ -37,4 +46,18 @@ public ResponseEntity<NormWorkResponseSchema> getNorm(NormWorkEli eli) {
...
     );
   }
+
+  /**
+   * Get a list of all expression of this work
+   * @param eli the eli of the work
+   * @return a list of information about expressions
+   */
+  @GetMapping(path = "/expressions", produces = { APPLICATION_JSON_VALUE })
+  public List<NormExpressionListResponseSchema> getExpressions(NormWorkEli eli) {
```
SebastianRossa, line 56: Is this really a NormExpressionListResponseSchema or rather a NormExpressionResponseSchema right now? My assumption would be that this means it contains a list. My suggestion would be to either move the List to the response schema (preferred) or change the name.

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/input/restapi/schema/NormExpressionListResponseSchema.java
```diff
@@ -0,0 +1,19 @@
...
+import lombok.Data;
+import lombok.experimental.SuperBuilder;
+
+/**
+ * Represents the response schema for information about a {@link Norm} on expression level in the REST API in list responses.
+ */
+@SuperBuilder(toBuilder = true)
+@AllArgsConstructor
+@Data
+public class NormExpressionListResponseSchema {
```
SebastianRossa, line 14: See comment above

backend/src/main/resources/db/migration/V0.41__add_gegenstandlos_column.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE dokumente
+    ADD COLUMN gegenstandlos boolean
```
VictorDelCampo, line 2: FYI, just sharing a thought...I know the `gegenstandslos` is only present in the _rechtsetzungsdokument_ and technically it is correct that only these documents will get a `true` in the new column but it feels _fachlich_ wrong, that the _regelungstext_ referenced by the _rechtsetzungsdokument_ will have `false` in the column. But I guess we are just following the standard and we should keep in mind that for knowing if _gegenstandslos_ we should always check the _rechtsetzungsdokument_

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/output/database/repository/DokumentRepository.java
```diff
@@ -33,4 +36,25 @@ Optional<DokumentDto> findFirstByEliDokumentExpressionOrderByEliDokumentManifest
...
+    SELECT DISTINCT ON (eli_norm_expression) eli_norm_expression, gegenstandlos
+      FROM dokumente WHERE
+        eli_norm_work = :normWorkEli AND
+        subtype = '/akn/ontology/de/concept/documenttype/bund/rechtsetzungsdokument'
+      ORDER BY
+        eli_norm_expression DESC, -- Order the norms from oldest to newest
+        eli_norm_manifestation DESC -- Get the latest manifestation for each expression
+    """,
+    countQuery = """
+    SELECT count(DISTINCT eli_norm_expression)
```
VictorDelCampo, line 51: Would counting over the `norm_expression` table be faster than doing it on `dokumente`? I guess you are doing with the latter because only there we have the `subtype` but the `norm_expression` table is for sure much smaller (at least half). But didn't check if we have there also the work_eli...

backend/src/main/resources/db/migration/V0.41__add_gegenstandlos_column.sql
malte-laukoetter: yeah had the same thought and was actually thinking about placing this on the norm_expression table. But it's not possible to calculate it using a generated column there so keeping it update would be a lot harder to achieve and a lot simpler to break. So to me it seemed like the smaller disadvantage to need to know to look for it in only the _rechtsetzungsdokument_

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/output/database/repository/DokumentRepository.java
malte-laukoetter: yeah that would probably be faster. But as long as we don't see a need due to performance problems I would like to keep the queries for the count and the actual data as similar as possible as that makes understanding what is happening simpler.

backend/src/main/resources/db/migration/V0.41__add_gegenstandlos_column.sql
VictorDelCampo: I agree. Extending javadoc of `isGegenstandlos`

backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/output/database/repository/DokumentRepository.java
VictorDelCampo: I just checked on staging through the API and calling the endpoint with the norm that has the most expressions (304) and it took 193ms. So yes, just awesome fast :)

backend/src/main/resources/db/migration/V0.41__add_gegenstandlos_column.sql
SebastianRossa: Another (late) thought from my side:
Do we really need to store gegenstandslos in a table? (Would be helpful for indexing and such but i think this is not needed here?)
If not an alternative might be a **view**. Maybe something like this?

CREATE OR REPLACE VIEW norm_expression_with_gegenstandlos AS
SELECT
  ne.*,
  xpath_exists(
    '//*[local-name()="legalDocML.de_metadaten"]/*[local-name()="gegenstandlos"]',
    d.xml::xml
  ) AS gegenstandlos_exists
FROM
  norm_expression ne
JOIN
  norm_manifestation nm
    ON nm.eli_norm_expression = ne.eli_norm_expression
JOIN
  dokumente d
    ON d.eli_norm_manifestation = nm.eli_norm_manifestation;

SebastianRossa (commented):
Nice work as always! :-)

malte-laukoetter:
> Do we really need to store gegenstandslos in a table? (Would be helpful for indexing and such but i think this is not needed here?)
If not an alternative might be a view. Maybe something like this?

Calculating the results of xpath expressions in a sql-query always takes quite long. So unless we want to wait for multiple seconds to get results we need to precompute them before running the query. A view also doesn't help for that as that doesn't store the results but calculates them on the fly. A materialized-view would have been an alternative but than we need to manage the recomputation of it ourself while generated columns are automatically updated whenever the data they rely on are updated. So this is so far the best solution I could come up with.

For requesting just single norms from the DB a xpath is not a problem as calculating it for just a few documents is still fast. The problem starts appearing when we try to load 10s-100s of dokumente at once

backend/src/main/resources/db/migration/V0.41__add_gegenstandlos_column.sql
malte-laukoetter: Calculating xpath-expression takes quite a long time. So unless we want to wait many seconds to get the list we need to precomupte the gegenstandslos value. A view does not solve this problem as it still calcuates it on the fly. A materialzed-view would have been an alternative, but then we would need to manage the recomputation ourself, while a generated column manages that on its own.
So far I have therefore not found a better solution for this than using a generated column.

This is not a problem when requesting just a handfull of Dokumente but once we get to the 100s-1000s it get's quite noticable

PR #1499, opened 2025-07-02
RISDEV-8497-ensure-a-new-working-copy-manifestation-is-created-if-no-exists-already-when-saving-a-norm

malte-laukoetter:
I also used this ticket to ensure as many update as possible go through the checks and preprocessing provided by NormServer::updateNorm (fix eids, remove dead references, sort elements, validate xsd).

This also detected some places in our code where we created schema-invalid ldml files. Most noteably our references-editor added a @refersTo="zitierung" to references while this isn’t a valid value for the refersTo-attribute. I changed this to treat a missing refersTo-attribute as if it was containing zitierung. We currently do not support any other types anyway.


backend/src/main/java/de/bund/digitalservice/ris/norms/adapter/output/database/service/NormDBService.java
```diff
@@ -97,8 +95,7 @@ public Optional<Norm> loadNormByGuid(LoadNormByGuidPort.Options options) {
       .map(NormManifestationMapper::mapToDomain);
   }
 
-  @Override
-  public Optional<Norm> updateNorm(UpdateNormPort.Options options) {
+  private Optional<Norm> updateNorm(UpdateOrSaveNormPort.Options options) {
```
VictorDelCampo: So you removed UpdateNormPort and converted the overridden method to a private function. Fine. But Why passing the options of the other port? Why not just take the Norm object as the input parameter? It doesn't feel right to use the input parameter of the port, specially if the options only have a norm object

backend/src/main/java/de/bund/digitalservice/ris/norms/application/port/output/UpdateOrSaveNormPort.java
```diff
@@ -1,9 +1,14 @@
 package de.bund.digitalservice.ris.norms.application.port.output;
 
+import de.bund.digitalservice.ris.norms.application.service.NormService;
 import de.bund.digitalservice.ris.norms.domain.entity.Norm;
 
 /**
  * Interface representing a port for saving or updating a {@link Norm}.
+ * <p>
+ *   It is better to use {@link NormService#updateNorm(Norm)} as that method also ensures all preprocessing is done
```
VictorDelCampo, line 9: Good. I would maybe also add the info about "better call that one, which is at the end calling this output port". So that it is clear that the pre-processing needs to be done in the application layer before calling the output port for the actual saving to the db. I was also thinking about adding kind of a check in the implementation of this output port so that we disallow calling it if the pre-processing has not been done? But I guess we would need to hold the info about the pre-processing been done somewhere in the norm object and feels like an overkill. So just an FYI

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/NormService.java
```diff
@@ -136,24 +141,36 @@ public String updateRegelungstextXml(UpdateRegelungstextXmlUseCase.Options optio
     return XmlMapper.toString(updatedRegelungstext.getDocument());
   }
 
+  private void prepareNormForSaving(Norm norm) {
+    norm
+      .getDokumente()
+      .forEach(dokument -> {
+        EidConsistencyGuardian.eliminateDeadReferences(dokument.getDocument());
```
VictorDelCampo: I know it is not part of this PR but feels odd that the `EidConsistencyGuardian` is a utility class with static methods but both the LdmlDeElementSorter and the LdmlDeValidator are proper spring services. I guess the reason was the utility class came first and didn't need till know any further services and the other too rely in other services. Shall we also convert the guardian into a service? Maybe tech debt ticket?

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/NormService.java
```diff
@@ -136,24 +141,36 @@ public String updateRegelungstextXml(UpdateRegelungstextXmlUseCase.Options optio
...
+        EidConsistencyGuardian.correctEids(dokument.getDocument());
+        ldmlDeElementSorter.sortElements(dokument.getDocument().getDocumentElement());
+      });
+
+    ldmlDeValidator.validateXSDSchema(norm);
+  }
+
   /**
-   * It not only saves a {@link Norm} but makes sure that all Eids are consistent.
+   * It not only saves a {@link Norm} but makes sure that all Eids are consistent. It always saves to the working copy
```
VictorDelCampo: Would adapt the docs so that it does not only talk about correcting the Eids but rather doing all the pre-processing needed (maybe just linking to the private method or better add all the pre-processing steps here?)

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/NormService.java
```diff
@@ -136,24 +141,36 @@ public String updateRegelungstextXml(UpdateRegelungstextXmlUseCase.Options optio
...
    * @throws NormNotFoundException if the norm cannot be found
    */
   public Norm updateNorm(Norm normToBeUpdated) {
-    normToBeUpdated
-      .getDokumente()
-      .forEach(dokument -> {
-        EidConsistencyGuardian.eliminateDeadReferences(dokument.getDocument());
-        EidConsistencyGuardian.correctEids(dokument.getDocument());
-      });
+    var norm = createNewVersionOfNormService.createNewManifestation(
```
VictorDelCampo, line 155: But shouldn't we actually check first if the latest manifestation is already a "working copy"? meaning has the unpublished state? Otherwise we would be creating all the time new manifestations? Or wait... this would lead to an override on the db because of the same point-in-time-manifestation because of the usage of the WORKING_COPY_DATE right? Meaning till it is actually published we would always be overriding the working copy. Does it make sense to having this "hidden" behaviour? or would it be better to make it explicit in the code like: search first for an already existing working copy, if not existent, create a new one?

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/NormService.java
```diff
@@ -745,17 +754,13 @@ private void fixTimeline(final NormExpressionEli removedNormExpressionEli) {
     // 2. The removed orphan was at the end of the timeline  (previous-norm is present)
     // The case that only the next-norm is present is not possible because at least the original expression must be present
     if (previousNorm.isPresent() && nextNorm.isPresent()) {
-      final Norm newPreviousManifestation = createNewVersionOfNormService.createNewManifestation(
-        previousNorm.get()
-      );
-      final FRBRExpression previousFrbrExpression = newPreviousManifestation
+      final Norm previousExpression = previousNorm.get();
```
VictorDelCampo, line 754: Why this change? We do need a new manifestation of that expression because of changing the next version id

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/NormService.java
```diff
@@ -745,17 +754,13 @@ private void fixTimeline(final NormExpressionEli removedNormExpressionEli) {
...
+      final FRBRExpression previousFrbrExpression = previousExpression
         .getRegelungstext1()
         .getMeta()
         .getFRBRExpression();
-      final Norm newNextManifestation = createNewVersionOfNormService.createNewManifestation(
-        nextNorm.get()
-      );
-      final FRBRExpression nextFrbrExpression = newNextManifestation
+      final Norm nextExpression = nextNorm.get();
```
VictorDelCampo, line 759: Same here right? We also need a new manifestation but for changing this time the previous version id

VictorDelCampo (commented):
Looks good @malte-laukoetter. Thanks for cleaning up the updateNorms methods at the db level. Just some comments

backend/src/main/java/de/bund/digitalservice/ris/norms/application/port/output/UpdateOrSaveNormPort.java
malte-laukoetter: We currently still have two other methods calling this port (as they need to save the publish state at the same time). But I would want to change them as well that they also use `NormService::updateNorm`. Afterwards we might be able to add an architecture check that the port is only used by that method, which I would prefer over passing along information about which pre-processing steps have been done.

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/NormService.java
malte-laukoetter: The reason they have static methods is that so far there is no need for them to have any state and that they do not rely on other services. So there was no need for them to be a Service. But yes for consistency we could change that, but I don't see that change as something important

backend/src/main/java/de/bund/digitalservice/ris/norms/application/port/output/UpdateOrSaveNormPort.java
VictorDelCampo: Should we create a tech debt ticket for those two methods?

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/NormService.java
malte-laukoetter: > Or wait... this would lead to an override on the db because of the same point-in-time-manifestation because of the usage of the WORKING_COPY_DATE right?

yes, there is only one possible eli for the working copy so calling `createNewVersionOfNormService.createNewManifestation` when already provided with a working-copy does nothing.

> Does it make sense to having this "hidden" behaviour? 

With this change no other part of the code really needs to know about manifestations and working-copies as that is fully handled here.

> or would it be better to make it explicit in the code like: search first for an already existing working copy, if not existent, create a new one?

That would just create an additional loading step of the existing working-copy that then is not really used for anything. I actually implemented it first in that way and then realised that I was not using that loaded norm for anything other than deciding if `createNewVersionOfNormService.createNewManifestation` should be called. But as that method is idempotent it is not important to ensure it is only called on non-working-copies.


There is 1 case when there is a unpublished manifestation in the database that is not using the working-copy date: Newly imported Verkündungen.
As they have there own point-in-time-manifestation date set by the e-verkündung we keep that date. But as we do not publish those they are also in the publishing state UNPUBLISHED. But also in that case this logic is sound as it then creates a working-copy when we change something in the verkündung. Also that way we still have the original imported manifestation in our database which sees like a good idea to me.

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/NormService.java
malte-laukoetter: the updateNorm method now takes care of that, so we can just update the previousExpression directly and when it gets saved the new manifestation for it is created.

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/NormService.java
malte-laukoetter: same here

backend/src/main/java/de/bund/digitalservice/ris/norms/application/port/output/UpdateOrSaveNormPort.java
malte-laukoetter: https://digitalservicebund.atlassian.net/browse/RISDEV-8577

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/NormService.java
VictorDelCampo: Mmmm i am sorry but I am not fully convinced. When you say

> yes, there is only one possible eli for the working copy so calling createNewVersionOfNormService.createNewManifestation when already provided with a working-copy does nothing.

Is not really true right? The method will create a new norm object and yes have the same eli. But the method does something and we just let our db layer "override" the existing one. I think making the code more explicit would be better.

> That would just create an additional loading step of the existing working-copy that then is not really used for anything. I actually implemented it first in that way and then realised that I was not using that loaded norm for anything other than deciding if createNewVersionOfNormService.createNewManifestation should be called. 

Wouldn't you be passing the loaded norm to the pre-processing method and then again to the updateNorm port?

And about the method being sound also for the use case of UNPUBLISHED manifestations because of newly imported Verkündungen, yes it would be "safe" but again a bit hidden for the naked eye ;)

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/NormService.java
VictorDelCampo: Oh ok now I get it... so when updating one expression we don't need to take care anymore of creating the new manifestations... that will be done by the updateNorm now

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/NormService.java
malte-laukoetter: >Is not really true right? The method will create a new norm object and yes have the same eli. But the method does something and we just let our db layer "override" the existing one. I think making the code more explicit would be better.

2 norms with the same content are equal. So even after creating a new norm object it is still the same norm (as all the content is copied over). The only other thing that method is doing is updating the FRBRManifestation fields related to the eli (FRBRthis, FRBRuri and FRBRdate) but those stay the same when the eli stays the same. So this is not doing anything when getting passed a working-copy and the newly created norm object is equal to the norm put into it.

> Wouldn't you be passing the loaded norm to the pre-processing method and then again to the updateNorm port?

If we would want to do that we would need to first update the loaded norm with the information from the norm object passed to the method. Which would mean overwriting both the Dokumente and BinaryFiles (as we do not know what has changed). But that would mean we are using none of the information we have just loaded but are just copying the norm once more.

VictorDelCampo: Alright. I am getting there :) Go for the merge

PR #1500, opened 2025-07-02
feat(frontend): add Bestand paginated list view

hamo225:
RISDEV-8454


frontend/src/views/datenbank/RisDatenbankTable.vue
```diff
@@ -0,0 +1,80 @@
...
+
+defineProps<{
+  items: DatenbankTableItem[]
+  total: number
+  currentPage: number
+  pageSize: number
+}>()
+
+const emit = defineEmits<{
+  page: [page: number]
```
malte-laukoetter, line 22: i would have included the action that happens with the page in the name of the event, so something like `pageChange`. Or maybe even if we only have next and previous events `onPreviousPageClick` `onNextPageClick`. With just the name `page` it's atleast to me not clear when this event happens.

malte-laukoetter: nevermind it's the same name primevue uses for that event so I actually prefer it this way as we can then find it in the primevue docu ^^

frontend/e2e/application/datenbank.spec.ts
```diff
@@ -0,0 +1,16 @@
...
+  "Datenbank view with list of works",
+  { tag: ["@RISDEV-8331"] },
+  () => {
+    test("shows expressions in Abgabe view", async ({ page }) => {
+      await page.goto("./datenbank")
+
+      await expect(
+        page.getByRole("heading", { name: "Datenbank" }),
+      ).toBeVisible()
+    })
```
malte-laukoetter, line 14: Do you still plan to add an e2e test that checks that some norms are shown in the table?

frontend/src/views/datenbank/Datenbank.view.vue
```diff
@@ -0,0 +1,47 @@
...
+const currentPage = ref(0)
+const pageSize = ref(100)
+
+const {
+  data,
+  isFetching: normsIsFetching,
+  error: normsError,
+} = useGetNorms(currentPage, pageSize)
+
+const norms = computed(() => data.value?.content || [])
```
malte-laukoetter: note: i prefer to use `??` instead of `||` for default values as that handles cases in which the value is falsy but not undefined or null (like the number 0 or an empty string) correctly

malte-laukoetter (approved):

frontend/e2e/application/datenbank.spec.ts
hamo225: My bad. Will add, commit and merge

PR #1508, opened 2025-07-04
feat: add work expression view with TOC navigation and preview

hamo225:
RISDEV-8517

### DIFF
--- a/frontend/src/router.ts
+++ b/frontend/src/router.ts
@@ -114,6 +114,11 @@ const routes: readonly RouteRecordRaw[] = [
         name: "Datenbank",
         component: () => import("@/views/datenbank/Datenbank.view.vue"),
       },
+      {
+        path: `${createNormWorkEliPathParameter()}`,
+        name: "DatenbankWorkDetail",
+        component: () => import("@/views/datenbank/WorkDetail.view.vue"),
+      },
     ],
   },
 
--- a/frontend/src/router.ts
+++ b/frontend/src/router.ts
@@ -118,6 +118,14 @@ const routes: readonly RouteRecordRaw[] = [
         path: `${createNormWorkEliPathParameter()}`,
         name: "DatenbankWorkDetail",
         component: () => import("@/views/datenbank/WorkDetail.view.vue"),
+        children: [
+          {
+            path: `expression/${createDokumentExpressionEliPathParameter("expression")}`,
+            name: "DatenbankWorkExpressionDetail",
+            component: () =>
+              import("@/views/datenbank/WorkExpressionDetail.view.vue"),
+          },
+        ],
       },
     ],
   },
--- a/frontend/src/views/datenbank/WorkDetail.view.vue
+++ b/frontend/src/views/datenbank/WorkDetail.view.vue
@@ -27,12 +27,9 @@ const selectedExpression = ref<string | null>(null)
 const { expressionsHeadingId, tocHeadingId, expressionHtmlHeadingId } =
   useElementId()
 
-console.log(selectedExpression)
 const selectedExpressionEli = computed(() => {
   if (!selectedExpression.value) return undefined
-  return DokumentExpressionEli.fromString(
-    selectedExpression.value + "/regelungstext-verkuendung-1",
-  )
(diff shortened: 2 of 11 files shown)


frontend/src/router.ts
```diff
@@ -114,6 +114,19 @@ const routes: readonly RouteRecordRaw[] = [
         name: "Datenbank",
         component: () => import("@/views/datenbank/Datenbank.view.vue"),
       },
+      {
+        path: `${createNormWorkEliPathParameter()}`,
+        name: "DatenbankWorkDetail",
+        component: () => import("@/views/datenbank/WorkDetail.view.vue"),
+        children: [
+          {
+            path: `expression/${createDokumentExpressionEliPathParameter("expression")}`,
```
malte-laukoetter, line 123: one thought i just had: In theory it should be enough to just include the additional parts of the expression eli here (pointInTime, version & language) and use the existing parts of the work eli. That way the path for a work would look like `/datenbank/eli/bund/bgbl-1/2025/1` and for an expression of that work like `/datenbank/eli/bund/bgbl-1/2025/1/2025-03-12/1/deu`. That way we wouldn't need to include the work parts twice

frontend/src/views/datenbank/WorkDetail.view.vue
```diff
@@ -0,0 +1,161 @@
...
+  const base = [
+    {
+      key: "normWorkTitle",
+      title: () => normWork.value?.title ?? "...",
+    },
+  ]
+
+  if (route.name === "DatenbankWorkExpressionDetail") {
+    const expressionEli = useDokumentExpressionEliPathParameter("expression")
+    console.log(expressionEli)
```
malte-laukoetter:
```suggestion
```

frontend/src/views/datenbank/WorkExpressionDetail.view.vue
```diff
@@ -0,0 +1,153 @@
...
+          :selected="selectedEids"
+          class="-mx-24"
+        />
+      </section>
+    </SplitterPanel>
+  </Splitter>
+</template>
+
+<style scoped>
+:deep(.selected) {
```
malte-laukoetter: could you scope this to just affect the stuff in the preview. Especially with `:deep` it is otherwise quite likely to affect the styling of something unrelated

frontend/src/views/datenbank/WorkDetail.view.vue
```diff
@@ -0,0 +1,161 @@
...
+    </Splitter>
+  </RisViewLayout>
+</template>
+
+<style scoped>
+:deep(.selected) {
+  background-color: var(--color-element-select-selected-background);
+  outline: 2px solid var(--color-element-select-selected-border);
+  outline-offset: 2px;
+}
```
malte-laukoetter: i dont think this is used in this view

frontend/src/views/datenbank/WorkDetail.view.vue
```diff
@@ -0,0 +1,161 @@
...
+      },
+    ]
+  }
+  return base
+})
+
+function getInForceDateFromEli(eli: string) {
+  try {
+    return formatDate(NormExpressionEli.fromString(eli).pointInTime)
+  } catch {
```
malte-laukoetter: when does this error case happen? And should we log something when it happens?

malte-laukoetter (commented):
Overall nice work! Just left some smaller comments and an idea i just had :)

frontend/src/views/datenbank/WorkDetail.view.vue
hamo225: when the ELI string that doesn't match the expected format, malformed, missing parts but i have now inlined the function within the computed treeNodesExpressions.

frontend/src/router.ts
hamo225: I do understand you. Lets pair on this on Wednesday. Or you can go ahead an make the change tomorrow if you have time. I will leave it as it is for now as i will not be here tomorrow.

malte-laukoetter: I merged this pr and sure we can have a look into that together tomorrow :-)

PR #1520, opened 2025-07-08
Add editor for Bestandskorrektur

SebastianRossa:
This is basically a copy of the Textkonsolidierungseditor with some minor tweaks regarding isGegenstandslos.


andreasphil (approved):

andreasphil:
@SebastianRossa Since we expect this view to change in the future, I am ok with accepting the code duplication

PR #1521, opened 2025-07-08
Replace dokument by norm elis

malte-laukoetter:
RISDEV-7987


VictorDelCampo (commented):
Looks good malte BUT what about the textko and metadata pages? They still have the subtype in the URLs

https://norms.ris.dev.ds4g.net/app/verkuendungen/eli/bund/bgbl-1/2024/10/2024-01-18/1/deu/textkonsolidierung/eli/bund/bgbl-1/2021/s818/2022-02-02/1/deu/regelungstext-verkuendung-1

https://norms.ris.dev.ds4g.net/app/eli/bund/bgbl-1/2021/s818/2022-02-02/1/deu/regelungstext-verkuendung-1/metadata

But both should also work on the norm expression eli right? Specially for the metadata because there we load first the frame metadata (which some are also coming from the rechtsetzunfsdokument right?) Even if we then load the regelungstext for the einzelelement metadata

And for the texto page sure we need the regelungstext BUT the url should be the norm expression eli?

PR #1522, opened 2025-07-08
Implement ADR 0020

VictorDelCampo:
So important stuff:

- New domain entity `NormExpression` to represent now the node that has not only text content but got the two new mandatory attributes
- Domain entity `AmendedNormExpressions` does not implement `AbstractSet` anymore (nor it implements `AbstractCollection`) because all add/remove/etc operations are based solely on the expression eli and do not care about the attributes
- The important stuff is happening in the `NormService`:
   - No more separation between retrieving the expressions before the first geltungszeit and the ones after and including it. We just take all existing expressions and check if the point-in-time is before or after/including. Needed for identifying the possible orphans in the before part
   - Method strongly reduced in complexity by extracting into private helper methods, separating two main branches (`handleExistingExpression` and `handleGeltungszeit`)
   - When handling existing expressions we now also calculate if it was created by the system (so a replacing one) or not using the helper function `isSystemExpression`
   - When handling geltungszeiten, we now differentiate between retrieving an already previewed `SYSTEM` and a `THIS_VERKUENDUNG` expression. For the first we replace it with `THIS_VERKUENDUNG` and the second helps us to know if the expression had already been created (override.)

### DIFF
--- a/backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/NormService.java
+++ b/backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/NormService.java
@@ -328,18 +328,16 @@ public class NormService
 
     if (earliestGeltungszeit.isEmpty()) return List.of();
 
-    var existingExpressionElis = loadNormExpressionElisPort.loadNormExpressionElis(
-      new LoadNormExpressionElisPort.Options(zielnormWorkEli)
-    );
-
-    var previousPossibleOrphanExpressions = collectPreviousExpressionsThatMayBeOrphans(
-      existingExpressionElis,
-      earliestGeltungszeit.get()
-    );
-    var relevantExistingExpressions = collectRelevantExistingExpressions(
-      existingExpressionElis,
-      earliestGeltungszeit.get()
-    );
+    var existingSortedExpressionElis = loadNormExpressionElisPort
+      .loadNormExpressionElis(new LoadNormExpressionElisPort.Options(zielnormWorkEli))
+      .stream()
+      .flatMap(normExpressionEli ->
+        loadNormPort.loadNorm(new LoadNormPort.Options(normExpressionEli)).stream()
+      )
+      .filter(Predicate.not(Norm::isGegenstandlos))
+      .map(Norm::getExpressionEli)
+      .sorted()
+      .toList();
 
     var affectedExpressionElis = verkuendungNorm
       .getRegelungstext1()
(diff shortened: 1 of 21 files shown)


backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/NormService.java
```diff
@@ -324,131 +324,198 @@ private List<Zielnorm.Expression> generateZielnormPreviewExpressions(
...
+    final boolean zeitgrenzeMissing = !geltungszeiten.contains(pointInTime);
+    final boolean hasAnyPrecedingWithZeitgrenze = amendedNormExpressions
+      .getNormExpressions()
       .stream()
-      .sorted(Comparator.comparing(Zielnorm.Expression::normExpressionEli))
-      .toList();
+      .takeWhile(expr -> !expr.equals(normExpression))
+      .anyMatch(NormExpression::getCreatedByZeitgrenze);
+
+    final boolean isCase1 = createdByZeitgrenze && !createdByReplacing && zeitgrenzeMissing;
```
malte-laukoetter, line 508: could you add a comment that these cases to refer to the cases in the ADR :)

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/NormService.java
```diff
@@ -324,131 +324,198 @@ private List<Zielnorm.Expression> generateZielnormPreviewExpressions(
...
+      .findFirst();
+    if (existingExpressionByThisVerkuendung.isEmpty()) {
+      var eli = eliService.findNextExpressionEli(zielnormWorkEli, geltungszeit, "deu");
+      expressions.add(
+        new Zielnorm.Expression(eli, false, false, false, Zielnorm.CreatedBy.THIS_VERKUENDUNG)
+      );
+    }
+  }
+
+  private Zielnorm.CreatedBy isSystemExpression(
```
malte-laukoetter, line 462: with the is in the name i would have expected a boolean as a response

malte-laukoetter (commented):
overall i found it quite understandable :-)

malte-laukoetter:
> No more separation between retrieving the expressions before the first geltungszeit and the ones after and including it. We just take all existing expressions and check if the point-in-time is before or after/including. Needed for identifying the possible orphans in the before part

I would expect this to take a long time for norms with many many expressions, but for now it should be fine and if it creates a problem we also have options to limit the list again (eg. taking all expressions after the first one that is not published)

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/NormService.java
VictorDelCampo: Opps I wanted it to be a boolean return but became something else. Converting to boolean method

VictorDelCampo:
> > No more separation between retrieving the expressions before the first geltungszeit and the ones after and including it. We just take all existing expressions and check if the point-in-time is before or after/including. Needed for identifying the possible orphans in the before part
> 
> I would expect this to take a long time for norms with many many expressions, but for now it should be fine and if it creates a problem we also have options to limit the list again (eg. taking all expressions after the first one that is not published)

But we were already doing that before, we were just doing it separately (see lines 331-342 of the previous version in NormService). At the beginning we only had the getting of the expressions after the first geltungszeit but then we to also load the previous ones in case the first time boundary moved (the previous one deleted). I also realized the API call takes 1-2s depending if there are more than 1 work and not really containing many expressions... So yes I think we will need to revisit this (and then you can explain what you mean with  _taking all expressions after the first one that is not published_)

PR #1542, opened 2025-07-11
feat(datenbank): persistent UI state for expression and TOC nav

hamo225:
- adds URL-based pagination with query parameters

RISDEV-8332


frontend/src/views/datenbank/Datenbank.view.vue
```diff
@@ -1,13 +1,24 @@
...
+const route = useRoute()
+const router = useRouter()
+
+const currentPage = ref(Math.max(0, (Number(route.query.page) || 1) - 1))
 const pageSize = ref(100)
 
+watch(
+  () => route.query.page,
+  (newPage) => {
+    currentPage.value = Math.max(0, (Number(newPage) || 1) - 1)
```
malte-laukoetter, line 18: in combination with `{ immediate: true }` you wouldn't need the same calculation in line 12

frontend/src/views/datenbank/WorkDetail.view.vue
```diff
@@ -60,21 +60,54 @@ const selectionKeysExpressions = ref<Record<string, boolean>>({})
...
+      })
     : [],
 )
 
-const handleExpressionNodeSelect = (node: TreeNode) => {
+const handleExpressionNodeUnselect = (node: TreeNode) => {
+  nextTick(() => {
+    selectionKeysExpressions.value = { [node.key]: true }
+  })
+}
```
malte-laukoetter, line 84: it's possible to select multiple elements right? Thats why we can't just load the selected eIds from the params?

frontend/src/views/datenbank/Datenbank.view.vue
```diff
@@ -19,7 +30,34 @@ const total = computed(() => data.value?.page.totalElements ?? 0)
...
+          query: { ...route.query, page: maxPage > 1 ? maxPage : undefined },
+        })
+      } else if (requestedPage < 1) {
+        router.replace({
+          query: { ...route.query, page: undefined },
+        })
+      }
+    }
+  },
+)
```
malte-laukoetter, line 60: i was wondering for a moment if it would make sense to move all the logic related to selecting pages into a composable? That way we could also write unit tests for it and it would be simpler to reuse it if we need it on a different page as well

malte-laukoetter (approved):
Looks good :) Also no need to change anything just left some thoughts I had while reading the code ^^

PR #1543, opened 2025-07-11
Eingebundene Stammform link in frontend

andreasphil:


backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/TableOfContentsService.java
```diff
@@ -93,7 +96,45 @@ private TableOfContentsItem getTableOfContentsItem(final Element element) {
...
+    }
+
+    return new TableOfContentsItem(eId, marker, heading, type, hasEingebundeneStammform, children);
+  }
+
+  private Optional<DokumentEli> getEingebundeneStammformEli(Element element) {
+    return NodeParser.getElementFromExpression(
+      "componentRef[@showAs=\"regelungstext-eingebundene-stammform\"]",
+      element
+    ).map(contentRef -> {
```
malte-laukoetter:
```suggestion
    ).map(componentRef -> {
```

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/TableOfContentsService.java
```diff
@@ -93,7 +96,45 @@ private TableOfContentsItem getTableOfContentsItem(final Element element) {
...
+        var stammformEliRaw = contentRef.getAttribute("src");
+        // The src of an eingebundene Stammform has an extra file extension, which is not part of
+        // any type of ELI we support. Removing it to get a regular Dokument Expression ELI
```
malte-laukoetter: i think the DokumentManifestationEli supports it (even without the pointInTimeManifestation)

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/TableOfContentsService.java
```diff
@@ -93,7 +96,45 @@ private TableOfContentsItem getTableOfContentsItem(final Element element) {
...
+  private Optional<String> getHeadingForEingebundeneStammform(DokumentEli stammformEli) {
+    return loadRegelungstextPort
+      .loadRegelungstext(new LoadRegelungstextPort.Options(stammformEli))
+      .map(stammformRegelungstext -> {
+        return NodeParser.getElementFromExpression(
+          "//longTitle",
+          stammformRegelungstext.getDocument()
+        )
+          .orElse(null)
+          .getTextContent();
```
malte-laukoetter: i think the regelungstext has a method like `getTitle()`, not 100% sure if it returns the correct title though

malte-laukoetter (approved):
not sure why the tests failing, checked them locally as well and there everything is working on my machine as well... Maybe try it again tomorrow? Otherwise maybe running them on a clean installation triggers them?

That the eingebundene-stammform has schematron-errors is (almost?) unavoidable so for me that's fine, eventhough i would preffer our test data to be as valid as possible

For loading stuff from the xml, if it's a one off thing i'm okay with having the xpath in the service otherwise i prefer to use the domain classes (or create new once) and add methods to them that take care of loading the needed information from the xml-element. (If i would rewrite the toc-service i would probably use the Article class in a lot of the places that use xpaths right now, but for now i don't think investing the time to change that is worth it)

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/TableOfContentsService.java
andreasphil: It's not, that only returns the `docTitle`. I don't think we use the `longTitle` as a whole anywhere in the backend right now but I could add it to the Regelungstext?

malte-laukoetter: alright, then i don't care, we already have 2 different title methods in the Regelungstext so adding one more will probably just make it more confusing

PR #1562, opened 2025-07-16
Add service for removing empty ldml.de elements

malte-laukoetter:
RISDEV-7906

### DIFF
--- a/backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/LdmlDeEmptyElementRemover.java
+++ b/backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/LdmlDeEmptyElementRemover.java
@@ -1,5 +1,7 @@
 package de.bund.digitalservice.ris.norms.application.service;
 
+import de.bund.digitalservice.ris.norms.domain.entity.AmendedNormExpressions;
+import de.bund.digitalservice.ris.norms.domain.entity.CustomModsMetadata;
 import de.bund.digitalservice.ris.norms.domain.entity.Geltungszeiten;
 import de.bund.digitalservice.ris.norms.domain.entity.Namespace;
 import de.bund.digitalservice.ris.norms.domain.entity.ZielnormReferences;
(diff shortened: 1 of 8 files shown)


VictorDelCampo:
@malte-laukoetter I was thinking if it wouldn't be just better to remove ALL XML nodes without any children but I just realized that we may have empty elements carrying semantic meaning (though not sure). Was your approach to just add elements that we "touch" to this new component and only check these elements? Would require of course to extend it if we manipulate another node (where we are also removing children)

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/LdmlDeEmptyElementRemover.java
```diff
@@ -0,0 +1,93 @@
...
+import org.w3c.dom.Node;
+
+/**
+ * Service for removing elements within a LegalDocML.de element that are empty and should only exist if they have child
+ * elements.
+ */
+@Service
+public class LdmlDeEmptyElementRemover {
+
+  private static final Map<Namespace, Set<String>> ELEMENTS_TO_REMOVE_IF_EMPTY = Set.of(
```
VictorDelCampo, line 24: Why this complicated thing and not just directly have the map like?

```
private static final Map<Namespace, Set<String>> ELEMENTS_TO_REMOVE_IF_EMPTY = Map.of(
    Namespace.INHALTSDATEN, Set.of(
        "proprietary",
        "analysis",
        "activeModifications",
        "passiveModifications",
        "notes",
        "lifecycle",
        "components"
    ),
    Namespace.METADATEN_BUNDESREGIERUNG, Set.of(
        "legalDocML.de_metadaten",
        "federfuehrung"
    ),
    Namespace.METADATEN_RIS, Set.of(
        "legalDocML.de_metadaten",
        "einzelelement"
    ),
    CustomModsMetadata.NAMESPACE, Set.of(CustomModsMetadata.TAG_NAME),
    ZielnormReferences.NAMESPACE, Set.of(ZielnormReferences.TAG_NAME),
    Geltungszeiten.NAMESPACE, Set.of(Geltungszeiten.TAG_NAME),
    AmendedNormExpressions.NAMESPACE, Set.of(AmendedNormExpressions.TAG_NAME)
);
```

malte-laukoetter:
We have lots of "empty" elements that have meaning due to there attributes - but also elements that have attributes but should still removed if they have not children. So manually managing a list of the elements is probably the only good option. (also the list should already have all the elements that we should remove when they are empty from the standard)

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/LdmlDeEmptyElementRemover.java
malte-laukoetter: multiple of the namespace are the same (eg. `ZielnormReferences.NAMESPACE` and `Geltungszeiten.NAMESPACE`) so that code wouldn't work

VictorDelCampo: Oh sure, missed that, but then we could do:
```
 private static final Map<Namespace, Set<String>> ELEMENTS_TO_REMOVE_IF_EMPTY = Map.of(
            Namespace.INHALTSDATEN, Set.of(
                    "proprietary",
                    "analysis",
                    "activeModifications",
                    "passiveModifications",
                    "notes",
                    "lifecycle",
                    "components"
            ),
            Namespace.METADATEN_BUNDESREGIERUNG, Set.of(
                    "legalDocML.de_metadaten",
                    "federfuehrung"
            ),
            Namespace.METADATEN_RIS, Set.of(
                    "legalDocML.de_metadaten",
                    "einzelelement"
            ),
            Namespace.METADATEN_NORMS_APPLICATION_MODS, Set.of(
                    CustomModsMetadata.TAG_NAME,
                    ZielnormReferences.TAG_NAME,
                    Geltungszeiten.TAG_NAME,
                    AmendedNormExpressions.TAG_NAME
            )
    );
```
I find that logically more readable.

PR #1565, opened 2025-07-16
Replace ZIP entry size check

VictorDelCampo:
Before we were using getSize() for checking the size of the entry because it is an optional metadata and not guaranteed to be accurate or present. Malicious ZIP files can spoof this value to evade size checks (e.g. in a zip bomb). By manually counting the uncompressed bytes read from the stream, we enforce accurate size validation and improve protection against oversized or malformed inputs.

RISDEV-8326


backend/src/test/java/de/bund/digitalservice/ris/norms/utils/TestZipHelper.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.zip.ZipOutputStream;
+
+/**
+ * Helper class to create test ZIP files with various structures
+ * for use in unit tests verifying ZIP extraction and validation.
+ * <p>
+ * Supports creating ZIPs with path traversal entries, folder entries,
+ * and archives with too many files.
+ */
+public class TestZipHelper {
```
malte-laukoetter, line 17: is there a reason for this class to exist and to create the files on the fly instead of adding the 3 files to the resources folder? This seems a lot more fragile than having the files laying around, especially as i do not expect anything about them to ever need to change.

VictorDelCampo: I actually started writing this helper for the test of a big zip, but I removed that test because of memory problems in CI (and locally when running tests with gradle). Then I left the other because I found it nice to not need more test files in our repo. But not really hard feelings....

malte-laukoetter: I rather have more tests files than code. But I'm also very fine with keeping it this way :)

PR #1573, opened 2025-07-17
rework-s3client-configuration

malte-laukoetter:


backend/src/test/resources/application.yaml
```diff
@@ -18,23 +18,14 @@ frontend:
```
VictorDelCampo, line 18: Doing here a comment because the commit where I saw it does not belong to a PR. You removed
```
local:
  file-storage: .local-storage
```
From the application.yml and moved it to the application-local.yml. But what about the *-local-docker.yml?

backend/src/main/java/de/bund/digitalservice/ris/norms/config/PublishingS3ClientConfig.java
```diff
@@ -0,0 +1,65 @@
...
+/**
+ * Configuration class for setting up Amazon S3 clients for private and public buckets.
+ *
+ * <p>The configuration properties for the S3 clients are loaded from the application properties file.</p>
+ *
+ * For local development or testing the {@link MockPublishingS3ClientConfig} can be used instead.
+ */
+@Slf4j
+@Configuration
+@ConditionalOnProperty(value = "otc.obs.public.access-key-id")
```
VictorDelCampo, line 20: It does not really belong to this PR but I just realized that the publication job is deactivated on production, should we also add to this config and also the `BucketServiceConfiguration`, which btw. could be named more suitably, the conditional on the enabled property?
`@ConditionalOnProperty("publish.enabled")`

VictorDelCampo:
@malte-laukoetter I would also extend a bit the part you added to the README of the infra repo (writing here so all comments of my review in the same place). When I saw what you added

> All kubernetes config maps with the label spring-config: "true" are automatically loaded by the Spring application.
> See also:
> https://docs.spring.io/spring-cloud-kubernetes/reference/property-source-config/configmap-propertysource.html

I didn't understand it but then I saw in our app repo what you added to the bootstrap file for the spring cloud configuration and realized it was just a marker for indicate what sources to load. Could you add this info to the README?

backend/src/test/resources/application.yaml
malte-laukoetter: Whenever we use the local-docker profile we also use the local profile so it's always loaded (if we wanted to have them independently there are a lot of other configuration things missing from local-docker as well)

malte-laukoetter:
@VictorDelCampo 
> Could you add this info to the README?

Hm... i don't know what additional information you are missing exactly. I linked to the spring docs that explain how this is implemented and explicitly didn't explain this again in here. This way our Readme only contains the information specific to our application and is not full of details not really needed (that we would also need to keep up to date)

VictorDelCampo:
> @VictorDelCampo
> 
> > Could you add this info to the README?
> 
> Hm... i don't know what additional information you are missing exactly. I linked to the spring docs that explain how this is implemented and explicitly didn't explain this again in here. This way our Readme only contains the information specific to our application and is not full of details not really needed (that we would also need to keep up to date)

I just saw the part with the labels in the link. thanks!

PR #1580, opened 2025-07-18
Frontend editing of eingebundene Stammformen (continued)

andreasphil:
Follow-up to #1576

RISDEV-8575


frontend/src/lib/eli/DokumentManifestationEli.ts
```diff
@@ -0,0 +1,136 @@
...
+   *
+   * @returns Norm Work ELI
+   */
+  asNormWorkEli(): NormWorkEli {
+    let match
+    let naturalIdentifier = this.naturalIdentifier
+    if ((match = this.subtype.match(/-(\d+)$/))) {
+      const subtypeCounter = Number.parseInt(match[1]) - 1
+      if (subtypeCounter > 0) naturalIdentifier += `-${subtypeCounter}`
+    }
```
malte-laukoetter, line 122: i was quite confused that this logic is hiding in here. Everywhere else asWorkEli is just extracting parts from an eli and not creating a totally new one. Also technically the work eli for `eli/bund/bgbl-1/2025/12/.../regelungstext-2` is still `eli/bund/bgbl-1/2024/12` and not `eli/bund/bgbl-1/2024/12-1`. The `12-1` eli is just the eli for the *new* work that is created for this eingebundene Stammform during the Fortschreibung.

Maybe another method name for this would be better suited? Also I'm not sure that this method even should be part of the DokumentManifestationEli class or would be better suited as a standalone function.

malte-laukoetter (commented):
@andreasphil everything else looked quite good and understandable to me :)

PR #1594, opened 2025-07-21
Stop create working copies during volldokumentation publishing

malte-laukoetter:
RISDEV-8696

### DIFF
--- a/backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/ReleaseService.java
+++ b/backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/ReleaseService.java
@@ -8,6 +8,8 @@ import de.bund.digitalservice.ris.norms.domain.entity.eli.NormManifestationEli;
 import java.time.LocalDate;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.stream.Collectors;
+import lombok.extern.slf4j.Slf4j;
 import org.jspecify.annotations.NonNull;
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Transactional;
@@ -17,12 +19,12 @@ import org.springframework.transaction.annotation.Transactional;
  * all the input ports. It is annotated with {@link Service} to indicate that it's a service
  * component in the Spring context.
  */
+@Slf4j
 @Service
 public class ReleaseService implements ReleaseAllNormExpressionsUseCase {
 
   private final UpdateOrSaveNormPort updateOrSaveNormPort;
   private final NormService normService;
-  private final CreateNewVersionOfNormService createNewVersionOfNormService;
   private final DeleteNormPort deleteNormPort;
   private final LdmlDeValidator ldmlDeValidator;
   private final LoadNormExpressionElisPort loadNormExpressionElisPort;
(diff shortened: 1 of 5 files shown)


backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/ReleaseService.java
```diff
@@ -75,64 +74,80 @@ public ReleaseService(
...
-    return workingCopies;
+    return releasedNorms;
   }
 
-  private static void setNewStandDerBearbeitung(ReleaseType targetReleaseType, Norm norm) {
+  private static ReleaseType effectiveReleaseType(ReleaseType targetReleaseType, Norm norm) {
     ReleaseType currentReleaseType = norm.getReleaseType();
 
     // PRAETEXT_RELEASED and VOLLDOKUMENTATION_RELEASED are not relevant as they remain unchanged.
     if (
```
VictorDelCampo, line 144: So what happens here if the target release os PRAETEXT_RELEASED but the current release type is VOLLDOKUMENTATION_RELEASED? Shouldn't we return VOLLDOKUMENTATION_RELEASED?

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/ReleaseService.java
```diff
@@ -75,64 +74,80 @@ public ReleaseService(
...
-        !updatedNorm.getManifestationEli().equals(oldManifestationEli) &&
-        // Do not delete the working copy
-        !Norm.WORKING_COPY_DATE.isEqual(oldManifestationEli.getPointInTimeManifestation());
-      if (newManifestationWasCreatedOnSave) {
+
+      if (effectiveReleaseType == ReleaseType.PRAETEXT_RELEASED) {
+        // save a new working copy that still includes the stuff removed by the pretext cleanup (so the information
+        // isn't lost)
+        var newWorkingCopy = normService.updateNorm(uncleanedCopy);
+        // add the working copy to the released norm list as it includes the release type information which is removed
```
VictorDelCampo, line 121: This is such a good catch Malte. Noice!

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/ReleaseService.java
```diff
@@ -75,64 +74,80 @@ public ReleaseService(
...
+        // in the pretext cleanup
+        releasedNorms.add(newWorkingCopy);
+      } else {
+        // Remove the working copy as it has 1:1 the same information as the just published one. A new working copy will
```
VictorDelCampo, line 125: Why needing to delete the working copy? Or better asked, why creating the working copy at all if the effective release type is not PRÄTEXT? As far as I see you always create the working copy, even if it is VOLLTEXT. Couldn't we just create the working copy only if PRÄTEXT?

VictorDelCampo (commented):
Hey Malte nice. Just two comments

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/ReleaseService.java
malte-laukoetter: we only create it in the prätext case. But we still need to remove the one that was already in the database before starting the release process. In the prätext case we do this by creating the new working copy (as it is overwriting the existing one). In the volldokumentationscase we remove it here.

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/ReleaseService.java
malte-laukoetter: oh right this logic is totally broken, i fixed it

PR #1595, opened 2025-07-21
Load expression preview of eingebundene Stammform

SebastianRossa:


backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/NormService.java
```diff
@@ -301,8 +311,69 @@ private List<Zielnorm> loadZielnormenPreview(final NormExpressionEli eli) {
...
+        var latestZielnormExpression = extractNorm(verkuendungNorm, zielnormWorkEli);
+        return new Zielnorm(
+          zielnormWorkEli,
+          latestZielnormExpression.getLongTitle().map(String::trim).orElse(""),
+          latestZielnormExpression.getShortTitle().map(String::trim).orElse(""),
+          List.of(
+            new Zielnorm.Expression(
+              latestZielnormExpression.getExpressionEli(),
+              false,
+              true,
```
VictorDelCampo: Why setting the isCreated always to `true`? It should only be the case if we know that the expression exists already on our db (if somebody click on "Expressionen erzeugen" already)

Maybe you could try to load first the latest expression for the given work using the standard loadNorm (but catching the NormNotFoundException if not found) and only then you do the extractNorm you have. This way you can set the isCreated accordingly

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/NormService.java
```diff
@@ -301,8 +311,69 @@ private List<Zielnorm> loadZielnormenPreview(final NormExpressionEli eli) {
...
+  private Norm extractNorm(Norm verkuendung, final NormWorkEli zielNormWorkEli) {
+    Set<Dokument> dokumente = verkuendung
+      .getRegelungstext1()
+      .getMeta()
+      .getProprietary()
+      .flatMap(Proprietary::getCustomModsMetadata)
+      .flatMap(CustomModsMetadata::getZielnormenReferences)
+      .stream()
+      .flatMap(ZielnormReferences::stream)
+      .filter(ZielnormReference::isNewWork)
```
VictorDelCampo: You don't really need this isNewWork filter right? Since you want to filter by the work eli

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/NormService.java
```diff
@@ -301,8 +311,69 @@ private List<Zielnorm> loadZielnormenPreview(final NormExpressionEli eli) {
...
+      .filter(z -> z.getZielnorm().equals(zielNormWorkEli))
```
VictorDelCampo, line 371: Actually we can only have 1 zielnorm-reference with the same work-eli (for the cases of isNewWork, at least in this first iteration without gestuftes inkraftreten), so why not just do a findFirst? I was confused why we were getting a Set<Dokument> for constructing the norm and it is because of this step right?

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/NormService.java
```diff
@@ -301,8 +311,69 @@ private List<Zielnorm> loadZielnormenPreview(final NormExpressionEli eli) {
...
+          .orElseThrow(() ->
+            new IllegalStateException(
+              "Reference was wrong: No article found for EId " + zielnormReference.getEId()
+            )
+          );
+
+        return referencingArticle
+          .getEingebundeneStammform()
+          .flatMap(eli ->
+            verkuendung
```
VictorDelCampo, line 388: Filtering from the dokumente of the verkuendung is maybe tricky? Since you will only get that 1 dokument and not possible attachments of that xml? But actually you don't care about the attachments here, you just need the long and short titles... it felt weird to create a "fake" norm object our of the dokumente just for extracting the titles. You could also just load from db the regelungstext belonging to that manifestation eli and extract titles from there, like done in the `TableOfContentService`.

VictorDelCampo (changes requested):
Hey Sebastian, I left some comments, some of them are for me a must to change.

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/NormService.java
SebastianRossa: Right. That is already filtered before

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/NormService.java
SebastianRossa: We get a Set because of the .collect(Collectors.toSet()); at the end. I changed that for clarity now.
Deleted the .filter(ZielnormReference::isNewWork) too since we have filtered for that before.

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/NormService.java
SebastianRossa: I think in the future we might need to extract additional information from the norm. Than it might come handy to always use a norm to extract information for the Zielnorm object as we always have the same interface than. I think the overhead is also minimal

VictorDelCampo (approved):
LGTM

PR #1622, opened 2025-07-28
Create new works (eS)

VictorDelCampo:
RISDEV-8675


LegalDocML.de/1.8.1/fixtures/eli/bund/bgbl-1/2024/17/2024-01-24/1/deu/2024-01-24/regelungstext-verkuendung-1.xml
```diff
@@ -570,12 +585,12 @@
       </akn:preamble>
       <akn:body GUID="9addbde1-187b-4173-8ad8-0ab8f88636fa" eId="hauptteil-n1">
          <akn:article GUID="64a65b7e-7a03-47c7-b91e-2bd600d858d8"
-                      eId="art-z"
+                      eId="art-z1"
```
malte-laukoetter, line 588: I think this should stay `art-z`. The akn:num element has no content so there is nothing to extract and therefore the eId should just be `art-z` and not `art-z1`.

backend/src/main/java/de/bund/digitalservice/ris/norms/application/port/input/CreateZielnormenExpressionsUseCase.java
```diff
@@ -28,4 +31,34 @@ public interface CreateZielnormenExpressionsUseCase {
...
+    }
+
+    @Override
+    public URI getType() {
+      return URI.create("/errors/new-work-already-exists");
+    }
+
+    @Override
+    public String getTitle() {
+      return "Expression %s of new work already exists".formatted(eli);
```
malte-laukoetter, line 56: The title should not include any variables as it should stay the same for all occurrences of the error (https://www.rfc-editor.org/rfc/rfc9457#name-title)

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/CreateNewWorkService.java
```diff
@@ -0,0 +1,171 @@
...
+/**
+ * Service for creating new expressions and manifestations of a norm.
+ */
+@Service
+public class CreateNewWorkService {
+
+  /**
+   * Creates a completely new work out of an embedded norm, the Rechtsetzungsdokument of the Verkündung and the generated expression eli for the new norm computed by the zielnorm-reference
+   * The Rechtsetzungsdokument of the new work is created out of the newly created Regelungstext. The redok metadata are copied from the Rechtsetzungsdokument of the Verkündung.
+   * @param embeddedNorm taking out of the Verkündung (not loaded from the DB)
```
malte-laukoetter, line 27: ~taking~ -> taken

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/CreateNewWorkService.java
```diff
@@ -0,0 +1,171 @@
...
+      .forEach(dokument -> {
+        final DokumentManifestationEli dokumentManifestationEli = dokument.getManifestationEli();
+        final String newNaturalIdentifier = expressionEli.getNaturalIdentifier();
+        final LocalDate newPointIntTime = expressionEli.getPointInTime();
+        dokumentManifestationEli.setNaturalIdentifier(newNaturalIdentifier);
+        dokumentManifestationEli.setPointInTime(newPointIntTime);
+        dokumentManifestationEli.setPointInTimeManifestation(Norm.WORKING_COPY_DATE);
+        final String newSubtype = dokumentManifestationEli
+          .getSubtype()
+          .replaceAll("-(\\d+)$", "-1");
```
malte-laukoetter, line 49: i think this doesn't work for other dokumente than regelungstexte? Eg. a `anlage-regelungstext-3` would now be called `anlage-regelungstext-1`

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/CreateNewWorkService.java
```diff
@@ -0,0 +1,171 @@
...
+    Rechtsetzungsdokument rechtsetzungsdokumentVerkuendung,
+    NormExpressionEli expressionEli
+  ) {
+    var newNorm = new Norm(embeddedNorm);
+    newNorm
+      .getDokumente()
+      .forEach(dokument -> {
+        final DokumentManifestationEli dokumentManifestationEli = dokument.getManifestationEli();
+        final String newNaturalIdentifier = expressionEli.getNaturalIdentifier();
+        final LocalDate newPointIntTime = expressionEli.getPointInTime();
```
malte-laukoetter, line 43: newPointIn**t**Time -> `newPointInTime`

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/CreateNewWorkService.java
```diff
@@ -0,0 +1,171 @@
...
+        dokumentManifestationEli.setNaturalIdentifier(newNaturalIdentifier);
+        dokumentManifestationEli.setPointInTime(newPointIntTime);
+        dokumentManifestationEli.setPointInTimeManifestation(Norm.WORKING_COPY_DATE);
+        final String newSubtype = dokumentManifestationEli
+          .getSubtype()
+          .replaceAll("-(\\d+)$", "-1");
+        dokumentManifestationEli.setSubtype(newSubtype);
+        dokument.getMeta().getFRBRManifestation().setEli(dokumentManifestationEli);
+        dokument.getMeta().getFRBRManifestation().setURI(dokumentManifestationEli.toUri());
+        dokument.getMeta().getFRBRExpression().setEli(dokumentManifestationEli.asExpressionEli());
```
malte-laukoetter, line 53: I'm not sure but does this also update the `akn:FRBRversionNumber` element? (and the other once that contain the parts of the eli) As they could change as well (atleast in theory, i think in practice they should stay the same, but I'm not 100% sure how that works if the amending norm was changed by another amending norm)

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/CreateNewWorkService.java
```diff
@@ -0,0 +1,171 @@
...
+
+    // Copy meta but only identification and proprietary
+    Element newMetaElement = NodeCreator.createElementWithEidAndGuid(
+      Namespace.INHALTSDATEN,
+      "meta",
+      documentCollection
+    );
+    final NodeList nodes = oldMeta.getElement().getChildNodes();
+    for (int i = 0; i < nodes.getLength(); i++) {
+      final Node node = nodes.item(i);
```
malte-laukoetter, line 114: We have `NodeParser#nodeListToList` that could simplify this kind of code a bit. But also no need to use it :)

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/CreateNewWorkService.java
VictorDelCampo: The `FRBRversionNumber` (present in the `FRBRExpression`) is not being updated, no. And the other ones holding parts of the eli in the `FRBRExpression` would be the `FRBRdate` which is already being updated and `FRBRlanguage` which is also not being updated. And I think for this iteration of the flow and with the current situation I wouldn't invest more thinking into it, you agree?

malte-laukoetter: Fine with me, I wanted to update the whole logic for updating the FRBR elements at some point anyway (to keep them updated between all the different dokumente and also ensure that the individual elements and the eli are always in sync)

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/CreateNewWorkService.java
VictorDelCampo: Jep you are right. And by doing the fix and testing I realized that the anlage were not being picked up and put into the new norm. Fixing

VictorDelCampo: Handled in [this](https://github.com/digitalservicebund/ris-norms/pull/1634) other PR.

PR #1634, opened 2025-07-30
Include XML attachments and binary files when creating eS

VictorDelCampo:
In a previous PR the [topic](https://github.com/digitalservicebund/ris-norms/pull/1622#discussion_r2238971187) was raised about how to correct the numbering of the subtype.

By analyzing it I realized that both XML and binary attachments were not being included into the new work when creating the new stammform.

This PR fixes both:
1. When extracting out of the Verkündung the embedded norm (Regelungstext), also references documents (ONLY IN THIS DOCUMENT, meaning no recursively) are also extracted
2. When correcting the subtype counter, it differentiates between regelugnstext and offene-struktur (checking the total of offene-struktur present and setting by ordinal after having sorting them by dok-manifestation-eli)

EDIT: all referenced documents and binary files (independent on which level) are extracted. Also when creating working copy all references are updated (not only the elis of the embedded norms a.k.a componentRef/@src)


backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/NormService.java
```diff
@@ -408,9 +408,40 @@ private Norm extractNorm(Norm verkuendung, final NormWorkEli zielNormWorkEli) {
...
+            .stream()
+            .filter(b -> b.getDokumentManifestationEli().equals(eli))
+            .findFirst()
+            .orElseThrow(() ->
+              new IllegalStateException(
+                "Referenced Dokument or BinaryFile not found for eli " + eli
+              )
+            );
+          binaryFiles.add(binaryFile);
+        }
```
malte-laukoetter: we should do this loop also for each dokument we find as also the attachments could reference binary files

backend/src/main/java/de/bund/digitalservice/ris/norms/domain/entity/Dokument.java
```diff
@@ -183,6 +183,21 @@ public List<String> getReferencedDokumentAndBinaryFileFileNames() {
...
+  public List<DokumentManifestationEli> getReferencedDokumentAndBinaryFileElis() {
+    return NodeParser.getNodesFromExpression(
+      "//componentRef/@src|//documentRef/@href|//img/@src",
+      document
+    )
+      .stream()
+      .map(Node::getNodeValue)
+      .map(DokumentManifestationEli::fromString)
+      .toList();
+  }
```
malte-laukoetter: i have the feeling we have this logic somewhere else already, but i'm not 100% sure

malte-laukoetter: `Dokument#getReferencedDokumentAndBinaryFileFileNames`, could you change that method to use the one you have written here so we only have one place with the xpath to get those

backend/src/main/java/de/bund/digitalservice/ris/norms/application/service/NormService.java
malte-laukoetter: It could probably work similar to `VerkuendungsImportService#findParseAndValidateFilesAsNorm`

VictorDelCampo: Added with [this](https://github.com/digitalservicebund/ris-norms/pull/1634/commits/4d78f5c8d226cf4c74887867fe7a170c6d1da36b) commit. The TODO in the Create...Service will finish that later. @malte-laukoetter wanna have a look?

EDIT: TODO is now done in [this](https://github.com/digitalservicebund/ris-norms/pull/1634/commits/5fc7d598f7d7da6f3c9dd0821c74acf81a2d196b) commit

malte-laukoetter (approved):

